PR #1, opened 2021-02-18
Add support for testing / lint / typecheck / format

LMS007:
### Commits

Add lint and typecheck support …

- Add devDependencies for missing packages
- Add support to lint exactly how the client does
- Add support to typecheck exactly how the client does
- Add build-types script to output .ts files into the /lib folder
- Modify build script to also output the .ts file
- Add typecheck and lint script shortcuts

------------

Remove npx command from scripts …

$ yarn <name> works without npx command

-----------

Add support for formatting …

Matches the scripts from the client
- $ yarn format
- $ yarn checkformatting

-------------

Add preact patch number

------------

Add support for unit tests …

- Add gulp and gulpfile.js with a single "test" task
- Add karam config similar to the client / lms repos but with more limited functionality for now.
  - no grep
  - no polyfills
  - no mockable-imports
  - no codecov
  - no css bundles
  - no "empty.html" file
  - no custom asserts
- Add required karma and karma plugin packages
- Add bootstrap.js test start-up file
- Add enzyme package for testing preact components
- Add browserify to bundle js files for testing

-----------------

Partially fixes https://github.com/hypothesis/client/issues/3017

### TODO in follow up PRs

- CI for this repository that run the basics here, test, lint, codecov, etc.
- More functionality for testing (grep, mockable-imports, css, etc...)
- How to docs for a release + update outdated docs relating to client repo
- Publish action to npmjs on-release 
- Probably a few other things I'm totally forgetting.


test/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
LMS007, line 1: This is not a folder for unit tests, but a folder for things needed for testing such as the bootstrap.js file. I'm not sure this is the best name for the folder, but I could not think of a better one.

.eslintrc
```diff
@@ -1,5 +1,14 @@
 {
-  "parserOptions": {
-    "sourceType": "module"
+  "extends": ["hypothesis", "plugin:jsx-a11y/recommended"],
+  "rules": {
+    // Suppressed to make ESLint v6 migration easier.
+    "no-prototype-builtins": "off",
+
+    // Handled by Prettier.
+    "comma-dangle": "off",
```
robertknight, line 8: It doesn't look like these two rule customizations are needed in this repository using the current version of ESLint and the current eslint-config-hypothesis package.

.prettierignore
```diff
@@ -0,0 +1,3 @@
+.*/**
+docs/
```
robertknight, line 2: Prettier can format markdown. Might be useful to set it up to do that in future for files in `docs/`.

src/karma.config.js
```diff
@@ -0,0 +1,126 @@
+/* global process */
+
+let chromeFlags = [];
+
+process.env.CHROME_BIN = require('puppeteer').executablePath();
+
+// In Docker, the tests run as root, so the sandbox must be disabled.
+if (process.env.RUNNING_IN_DOCKER) {
```
robertknight, line 8: The `RUNNING_IN_DOCKER` section is only needed in Jenkins IIRC. Since we're using GitHub Actions I think we can drop this.

src/karma.config.js
```diff
@@ -0,0 +1,126 @@
...
+      // Output only summary and errors in development to make output easier to parse.
+      output: 'minimal',
+    },
+
+    // Use https://www.npmjs.com/package/karma-mocha-reporter
+    // for more helpful rendering of test failures
+    reporters: ['progress', 'mocha'],
+
+    // web server port
+    port: 9876,
```
robertknight, line 89: Might be useful in future to pick a different port that doesn't conflict with the client's tests. Let's leave as-is for the moment.

test/.eslintrc
robertknight: Yes, this is potentially a little confusing. We don't have an alternate convention to apply, so I think we can stick with this for the moment.

robertknight (approved):
I noted a couple of ESLint rules that can be removed as well as a simplification of the Karma config. Otherwise this is good to go.

src/karma.config.js
LMS007: @robertknight , the client and lms both use this port too, we probably want 3 different ports.

PR #2, opened 2021-02-24
Standup CI & publish action

LMS007:
Mostly just cut from the client with the minor difference of not installing the /frontend-shared subfolder or doing any of the version bumping. I performed various testing to ensure all pieces fail individually when appropriate. 

The releases (for now) will be done manually via github IU. At some point we might want to perform automated releases similar to how the client does it with `create-github-release.js`, but I don't know how to do that without a manual step intervention when we take away Jenkins. So for now, the UI is sufficient. 

We also may want to soon port over the `generate-change-list.js` script so when we do a release, we can accurately document what has changed.

### DIFF
--- a/.github/wip.yml
+++ b/.github/wip.yml
@@ -1,0 +1,8 @@
+locations:
+- title
+- label_name
+- commit_subject
+terms:
+- do not merge
+- fixup
+- wip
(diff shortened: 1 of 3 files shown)


robertknight (approved):
Looks fine to me. As discussed on a video call, a manually triggered publish workflow might be fine. The main requirement is that the workflow is run on GitHub Actions so that it can be executed consistently.

PR #4, opened 2021-02-25
Add build-css gulp task for testing

LMS007:
Add styles.css to karam config. Unit tests rely on css for a11y requirements such as contrast.

The `styles.css` and `styles.css.map` files are output to `build/` dir which has no other purpose than to support these files for testing **only**. I'm not sure that's the best choice for the folder name and I'm open to other suggestions. 

TODO:
add `accessibility.js` module from the client.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 lib/
+build/
 
 # Use Yarn rather than npm to manage the lockfile.
 package-lock.json
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,7 +2,27 @@
 
 'use strict';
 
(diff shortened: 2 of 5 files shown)


robertknight (approved):

robertknight:
> I'm not sure that's the best choice for the folder name and I'm open to other suggestions.

For library projects I think `build/` is a reasonable choice for a dir in which to generate files that are needed only during development. There is some possible confusion with `lib/` for newcomers. I think this is a case where choosing a convention and applying it widely is more important than exactly what the convention is.

LMS007:
> ely is more important than exact

Probably good to call this distinction out in docs following this PR.

PR #5, opened 2021-02-25
Add accessibility file to run axe tests

LMS007:
This file can be used to run axe tests via checkAccessibility locally or it can be exported to other projects.

nb. the global `assert` must be exposed prior to running checkAccessibility.

-----------

There are no files that currently use `checkAccessibility` in this package, but we expect some to land in the near future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "@babel/core": "^7.1.6",
     "@babel/preset-env": "^7.1.6",
     "@babel/preset-react": "^7.0.0",
+    "axe-core": "^4.0.0",
     "babel-plugin-mockable-imports": "^1.5.1",
     "babelify": "^10.0.0",
     "browserify": "^17.0.0",
--- a/test/util/accessibility.js
+++ b/test/util/accessibility.js
@@ -1,0 +1,97 @@
+import { run } from 'axe-core';
+import { ReactWrapper, mount } from 'enzyme';
+import { isValidElement } from 'preact';
+
+/**
+ * @typedef Scenario
+ * @prop {string} [name] -
+ *   A descriptive name for the scenario. Defaults to "default" if not specified.
+ * @prop {() => import('preact').VNode|ReactWrapper} content -
+ *   A function that returns the rendered output to test or an Enzyme wrapper
+ *   created using Enzyme's `mount` function.
+ * @prop {string} [backgroundColor] -
+ *   Background color onto which to render the element. This can affect the
+ *   result of color contrast tests. Defaults to white.
+ */
+
+async function testScenario(
+  elementOrWrapper,
+  { backgroundColor = 'white' } = {}
+) {
+  const container = document.createElement('div');
+  container.style.backgroundColor = backgroundColor;
+  document.body.appendChild(container);
+
+  let wrapper;
+  if (elementOrWrapper instanceof ReactWrapper) {
+    wrapper = elementOrWrapper;
+    container.appendChild(elementOrWrapper.getDOMNode());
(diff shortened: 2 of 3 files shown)


src/test-util/accessibility.js
```diff
@@ -0,0 +1,100 @@
...
+ *   A descriptive name for the scenario. Defaults to "default" if not specified.
+ * @prop {() => import('preact').VNode|ReactWrapper} content -
+ *   A function that returns the rendered output to test or an Enzyme wrapper
+ *   created using Enzyme's `mount` function.
+ * @prop {string} [backgroundColor] -
+ *   Background color onto which to render the element. This can affect the
+ *   result of color contrast tests. Defaults to white.
+ */
+
+/** @typedef {ScenarioProps & Object} Scenario */
```
LMS007, line 17: _This is a change from the client_ 
Because this file is now exported in `index.js`, we can't skip the typecheck. It was complaining about a missing iterator due to this line
`for (let { name = 'default', content, ...config } of scenarios) {`

Extending the `Object` type seemed to do the trick, but I'm not sure that is best practice here.

src/test-util/accessibility.js
```diff
@@ -0,0 +1,100 @@
...
+        !(elementOrWrapper instanceof ReactWrapper) &&
+        !isValidElement(elementOrWrapper)
+      ) {
+        throw new Error(
+          `Expected "content" function for scenario "${name}" to return a Preact element or an Enzyme wrapper`
+        );
+      }
+
+      const violations = await testScenario(elementOrWrapper, config);
+      // @ts-ignore `assert` must be exposed in a testing bootstrap file
```
LMS007, line 92: Also, this line is new

src/index.js
```diff
@@ -6,3 +6,6 @@ export { useElementShouldClose } from './hooks/use-element-should-close';
 
 // Utilities
 export { normalizeKeyName } from './browser-compatibility-utils';
+
+// Test utilities
+export { checkAccessibility } from './test-util/accessibility';
```
robertknight, line 11: We'll need to put test utilities in a separate module than runtime utilities, otherwise Browserify is going to try and include all the transitive dependencies of the test utilities, such as enzyme and axe-core into the app bundle!

If we had a smarter bundler which supports tree shaking (eg. modern versions of Webpack, Rollup or Parcel) then they might be able to avoid doing this when the test utilities are not directly used. However even if that worked I'd still advise having a separate entry point for test dependencies.

To specify dependencies properly, the test utilities also ought to declare their Enzyme and axe-core peer dependencies. We don't really want the frontend-shared package to depend on those, so it is likely that test utilities should really go in their own package, which this repository then imports. The `frontend-toolkit/` repo might be an appropriate place for this, since it already has other development-only packages.

robertknight (changes requested):
The approach of exporting test utilities from the same module as the pattern library isn't going to work because it will add unused code, including all of the transitive dependencies of the test utilities, to applications which use them.

To handle this properly I think we'll need to put test utilities in a separate package, possibly alongside our other testing packages in https://github.com/hypothesis/frontend-toolkit/tree/master/packages.

robertknight (approved):
To clarify: In this revised PR, `test/util/accessibility.js` is purely for use within the package's own test suite? Other frontend projects will need to keep their own versions until a separate shared package is published that other projects can consume?

LGTM on that basis.

LMS007:
> s purely for use within the package's own test suite? Other frontend projects will need to keep their own versions until a separate shared package is published that other projects can consume?

correct

PR #6, opened 2021-02-25
Move build scrip to a gulp task

LMS007:
Remove `build-types` and `build-js` package scripts and instead export only a single gulp task called "build-js" and bind that to a package script called "build". Moving this work into gulp tasks will be required when we add any watchers in the future needed for a live pattern lib or express server.

Copy run.js from the client and move it into a new exported scripts/ folder that can be used for common scripts between frontend packages.

---------

### Part 1
This is a **low** priority PR and I'm not entirely convinced its the right approach for shared scripts. In short, I copied the gulp tasks from the client to here that more or less gets rid of the shell commands scripts in `package.json` and replaces them with js versions. Down the road this will be needed (I think) if we add watchers which we will want to have for a pattern lib/express server. 

### Part 2
The client uses a new helper function called `run()` and that "feels" share worthy enough to prep it for shared use. This gets a little confusing because its a build-time tool that also gets exported after build time (wut??). But because it's not a "module", we don't need to transpile it so I can see that part being confusing because it is in fact being transpiled somewhat benignly. I don't think this "run.js" file is critical either way, but I do think there may be scripts in the future that have value being shared and its something we may want to think on (or just say no, and copy-paste)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,6 +3,42 @@
 'use strict';
 
 const gulp = require('gulp');
+const babel = require('gulp-babel');
+const sourcemaps = require('gulp-sourcemaps');
+
+const { run } = require('./src/scripts/gulp/run');
+
+const buildJs = () => {
+  // There does not appear to be a simple way of forcing gulp-babel to use a config
+  // file. Load it up and pass it in manually.
+  const babelConfig = require('./.babelrc.cjs');
+
+  return (
+    gulp
+      .src(['./src/**/*.js', '!**/test/*.js'])
+      // Transpile the js source files and write the output in the frontend-shared/lib dir.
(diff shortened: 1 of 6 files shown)


robertknight (changes requested):
I'm fine with moving logic from scripts run by `package.json` to a gulpfile.js if that gives us more flexibility that we find useful. As you noted though, it seems like this isn't something we need right away - so perhaps defer merging this until we have a more concrete need?

I agree that the `run` utility may be useful to share, but it will need to go in a separate package than the pattern library because of the assumption of a Node runtime environment and dependencies on Node's `child_process` module.

LMS007:
Going to close this PR since its not critical at this time to add these gulp tasks. We can revisit later.

PR #7, opened 2021-02-26
Add changelog

LMS007:
These seems to be a very straightforward automated script to maintain a change log based on commits between tags. It works in concert with `npm version` so there are no extra steps.

To create a new version: 

`npm version 1.9.0`

You will see two changes committed to that branch. One to package.json (version number) and a second to CHANGELOG.md. 

_note: In order to capture the first set of changes, I added a v1.0.0 tag to the first commit._ 

The most important bit is probably to adopt a stand for breaking changes. With this tool, we can spec a regex and as long as we stick to that standard they should be called out with emphasis in the changelog.

For example.
`auto-changelog --breaking-pattern "BREAKING CHANGE:"`

fixes https://github.com/hypothesis/client/issues/3016


robertknight:
I think it is fine to use a tool to help generate a draft changelog as part of creating a new release, but the draft is generally going to need some editing to make it more useful to consumers of the package. Even though initially we are frequently going to be making a small change or addition to the package and then immediately using it in a project, I think we should get into the habit of documenting changes between versions properly. FWIW I personally like the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) standards for changelogs because of the emphasis on human readability.

To proceed here, I think I would adjust the workflow so that there is an easy way to generate a draft changelog for a new release, but it is made clear that the developer is supposed to edit the changelog to call out breaking changes, remove unimportant internal refactorings etc. before publishing the release.

LMS007:
I might come back to this, I think it's going to require a bit more thinking to get it right. 

My main questions are do we want to keep unreleased changes ongoing as we make commits, or should 1 person create / edit the changelog file before each release.  How do we categorize things as added/changed/removed, etc.  If basing a changelog off the git log is bad practice, then perhaps the only reasonable way to manage it is for each dev to modify it ongoing.

PR #8, opened 2021-02-27
Add yalc

LMS007:
depends on https://github.com/hypothesis/frontend-shared/pull/7

-------------

Add command to build, then publish a local yalc package. This can then be consumed by another application using `yalc add @hypothesis/frontend-shared`

Yalc allows the installed package to behave exactly like a published built package rather than a symlinked folder to the source files.

Add new script `push` which builds the src files into the /lib dir and builds the yalc package and then pushes that out to any consumers that have already added that package.

nb. A future change could do this with a file watcher.

------------

TODO:

- [ ]  update docs/

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     "sinon": "^9.0.0",
     "stringify": "^5.1.0",
     "typescript": "^4.1.5",
-    "watchify": "^3.7.0"
+    "watchify": "^3.7.0",
+    "yalc": "^1.0.0-pre.50"
   },
   "peerDependencies": {
     "preact": "^10.4.0"
@@ -53,7 +54,8 @@
     "lint": "eslint .",
     "checkformatting": "prettier --check '**/*.{js,scss}'",
     "format": "prettier --list-different --write '**/*.{js,scss}'",
-    "test": "gulp test"
+    "test": "gulp test",
+    "push": "yarn build && yalc push"
   },
   "prettier": {
     "arrowParens": "avoid",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,7 +56,8 @@
     "checkformatting": "prettier --check '**/*.{js,scss}'",
     "format": "prettier --list-different --write '**/*.{js,scss}'",
     "test": "gulp test",
-    "version": "auto-changelog -p && git add CHANGELOG.md"
+    "version": "auto-changelog -p && git add CHANGELOG.md",
+    "push": "yarn build && yalc push"
```
robertknight, line 60: I didn't know about the `yalc push` functionality. That's useful! Docs for other developers reading this: https://github.com/wclr/yalc#pushing-updates-automatically-to-all-installations

robertknight (commented):
The yalc-related changes on this branch look fine. Can you rework this PR to avoid the dependency on #7 (the changelog PR) though? This one is going to be easier to merge.

package.json
```diff
@@ -40,7 +40,8 @@
     "sinon": "^9.0.0",
     "stringify": "^5.1.0",
     "typescript": "^4.1.5",
-    "watchify": "^3.7.0"
+    "watchify": "^3.7.0",
+    "yalc": "^1.0.0-pre.50"
```
robertknight, line 44: This version number looks like we accidentally added a beta, but I confirmed that it is the current stable version.

robertknight (approved):

PR #14, opened 2021-03-09
Add codecov & mockable-imports

LMS007:
The codecov.yml config is still somewhat experimental and I don't believe it will produce a result until we merge this into main and generate a baseline.  Additionally, it appears we don't need to do any webhooks according to the docs:
https://app.codecov.io/gh/hypothesis/frontend-shared/settings _"A webhook has been enabled through GitHub Integrations. No setup required."_

### Codecov.yml docs (for reference)
https://docs.codecov.io/docs/commit-status#branches

This will close out epic: https://github.com/hypothesis/client/issues/3017. We can leave the pattern lib as a separate issue.


test/util/mock-imported-components.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Return true if `value` "looks like" a React/Preact component.
+ */
+function isComponent(value) {
+  return (
+    typeof value === 'function' &&
+    Object.prototype.hasOwnProperty.call(value, 'propTypes') &&
```
LMS007, line 7: Note: Small change on this line from the client to pass lint.

robertknight: We'll need to modify this if/when we remove `propTypes` from the components in this package. See the current version of this module in the client repo.

robertknight (approved):
LGTM. I ran it locally and it generated the expected coverage files. I'll merge this and see what happens.

PR #17, opened 2021-03-10
Add missing ".md" file types for `checkformatting` script

LMS007:
Format CHANGELOG.md so checkformatting passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v1.10.0] - 2021-03-09
 
 ### Added
+
 - Add `yarn push` to be able to link to the client for development [#8](https://github.com/hypothesis/frontend-shared/pull/8)
 - Add commands for testing / lint / typecheck / format [#1](https://github.com/hypothesis/frontend-shared/pull/1)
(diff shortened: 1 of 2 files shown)


lyzadanger (commented):
Hi! I was curious about the motivation for this change. I notice that neither the client nor LMS enforce formatting on markdown files.

I am given slight pause because it's not uncommon that markdown files, especially READMEs, are edited/PR-ed via GitHub's web interface. A missing line break in this case could cause CI to fail. Maybe that's OK, but I'm not sure that the benefits outweigh this and the inconsistency with other repositories. Perhaps we can chat as a team about this? I'm not opposed to it, necessarily, but would like to discuss.

robertknight:
I can see pros and cons to this. On the one hand, it ensures some consistency of formatting of all our documentation files. On the other hand, as Lyza notes, there is no way to run it when editing files directly in the GitHub interface. I have to say that's not something I find myself doing a lot of though - if it did happen, it would be easy enough to push a commit to fix it.

Perhaps we could give this a try as a practice and see if it is useful or annoying?

LMS007:
The primary motivation for this was to keep the CHANGELOG tidy as it goes through more frequent updates. I could see extra spaces being added over time. Thoughts?

lyzadanger (approved):
How about this: Can you add a reminder/TODO for, say, three months from now and we check in to see if we feel like this change has been uneventful and perhaps something we should reproduce in other repos? Obviously if it gets in our way between now and then we can remedy, but I'd like to loop back after some time to see if we can make formatting consistent across repositories.

PR #18, opened 2021-03-10
Add comment about dependent packages

LMS007:
Remove browserify as a dependent package as its not actually required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,19 @@
 
 A package of resources for Hypothesis front-end applications.
 
-#### Requirements
+## Requirements
 
-- preact
-- browserify
+Your project must have Preact installed as a dependency.
 
-### Usage
+`npm install --save preact`
+
+## Usage
 
 ```
 $ npm install @hypothesis/frontend-shared --save
 ```
 
-#### In SASS modules
+### In SASS modules
 
 To import default styling of frontend-shared components, include this line in the main project's SASS.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,8 +4,9 @@ A package of resources for Hypothesis front-end applications.
 
 #### Requirements
 
+Dependencies **required** for importing modules from `@hypothesis/frontend-shared`:
+
 - preact
```
LMS007, line 9: browserify is not really a requirement at all so I think it's best not to confuse this with a actual  dependency such as preact.

README.md
```diff
@@ -4,8 +4,9 @@ A package of resources for Hypothesis front-end applications.
 
 #### Requirements
```
robertknight, line 5: The heading levels are a bit wonky here. We've got a level 1 heading and underneath that a level 4 sub-heading ("Requirements") followed by a level 3 heading ("Usage"). I would correct this so that there is a level 1 heading at the top with level 2 for the next logical level down and so on.

README.md
robertknight: I would probably suggest rewording this section, including the list, as:

"Your project must have Preact installed as a dependency."

The main change here is just to make it clear that this is an unconditional dependency. npm/yarn will expect Preact to be installed whether or not you actually use any UI components, as long as preact is listed as a mandatory peer dependency. 

In writing the above comment I tried to find a link in the Preact docs that would tell you how to install Preact with yarn and npm but they miss that step, so perhaps we could add it here (`yarn install --save preact`).

robertknight (commented):
I suggested a revised wording. Please let me know what you think.

README.md
```diff
@@ -2,18 +2,19 @@
...
 
-#### Requirements
+## Requirements
 
-- preact
-- browserify
+Your project must have Preact installed as a dependency.
 
-### Usage
+`yarn install --save preact`
```
robertknight, line 9: Can I suggest changing this to `npm install` for consistency with the command below.

robertknight (approved):
LGTM with a change to use `npm` in both install commands for consistency, even though we use `yarn` for several of our projects.

PR #21, opened 2021-04-13
Update project configuration

lyzadanger:
This PR makes a couple of smallish changes to bring this project even further into consistency with other projects.

* Remove `propTypes` enforcement (shared button components, which come next, don't use `propTypes`)
* Add a `Makefile`. This is especially useful for the `make sure` command, and there is a stub of a target for `make dev`, which I anticipate using once there is a pattern library/local devserver in this project

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,57 @@
+.PHONY: default
+default: help
+
+.PHONY: help
+help:
+	@echo "make help              Show this help message"
+	@echo "make dev               Run the pattern library application (TODO)"
+	@echo "make build             Build the package"
+	@echo "make lint              Run the code linter(s) and print any warnings"
+	@echo "make checkformatting   Check code formatting"
+	@echo "make format            Automatically format code"
+	@echo "make test              Run the unit tests once"
+	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
+	@echo "make clean             Delete development artefacts (cached files, "
+	@echo "                       dependencies, etc)"
+
+.PHONY: dev
+dev: node_modules/.uptodate
+	@echo 'TODO: run local dev server with pattern library'
+
+.PHONY: test
+test: node_modules/.uptodate
+ifdef ARGS
+	yarn test $(ARGS)
+else
+	yarn test
+endif
+
+.PHONY: lint
+lint: node_modules/.uptodate
+	yarn run lint
+	yarn run typecheck
+
+.PHONY: clean
+clean:
+	rm -f node_modules/.uptodate
+	rm -rf lib
+	rm -rf build
+
+.PHONY: format
+format: node_modules/.uptodate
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,64 @@
...
+	yarn test
+endif
+
+.PHONY: lint
+lint: node_modules/.uptodate
+	yarn run lint
+	yarn run typecheck
+
+.PHONY: docs
+docs: python
```
LMS007, line 35: I think this can be removed

Makefile
```diff
@@ -0,0 +1,64 @@
...
+	@echo "make help              Show this help message"
+	@echo "make dev               Run the pattern library application (TODO)"
+	@echo "make build             Build the package"
+	@echo "make lint              Run the code linter(s) and print any warnings"
+	@echo "make checkformatting   Check code formatting"
+	@echo "make format            Automatically format code"
+	@echo "make test              Run the unit tests once"
+	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
+	@echo "make clean             Delete development artefacts (cached files, "
+	@echo "                       dependencies, etc)"
```
LMS007, line 15: Almost might be worth added a `make push` to map to the `yarn push` which will update the lib/ folder in linked references such as the client.

LMS007 (changes requested):
Lyza, can you make those 3 changes, then this is gtg

Makefile
```diff
@@ -0,0 +1,64 @@
...
+.PHONY: docs
+docs: python
+	tox -e py36-docs
+
+.PHONY: checkdocs
+checkdocs: python
+	tox -e py36-checkdocs
+
+.PHONY: clean
+clean:
```
LMS007, line 43: Can you also make this remove the /lib folder?

Makefile
robertknight: Indeed. We need to figure out the optimal tools for documentation for this project but I suspect it will involve a combination of markdown files and generating a static site with the pattern library in it.

robertknight (approved):
Kyle noted some obsolete Makefile commands related to docs. Otherwise LGTM.

Makefile
lyzadanger: Let's perhaps discuss this at the next frontend sync or other group chat? I can see arguments both ways.

Makefile
lyzadanger: Whoops. Oversight, thanks.

Makefile
lyzadanger: I think this generally sounds like a sound idea, but since I'm unfamiliar with the build process that generates `lib` in the first place (and thus unable to meaningfully test that I'm not breaking things), would you mind taking this as a follow-up?

lyzadanger: Yep!

PR #23, opened 2021-04-15
Simplification and type improvement in button component

esanzgar:
IconButton should no have children props.

Depends on #22

### DIFF
--- a/src/components/buttons.js
+++ b/src/components/buttons.js
@@ -26,6 +26,7 @@ import { SvgIcon } from './SvgIcon';
  * @typedef IconButtonBaseProps
  * @prop {string} icon - Icon is required for icon buttons
  * @prop {string} title - Title is required for icon buttons
+ * @prop {never} [children] - children are not allowed (use LabeledButton instead)
  */
 
 /**
@@ -36,34 +37,18 @@ import { SvgIcon } from './SvgIcon';
  * @param {ButtonProps} props
  */
 function ButtonBase({
-  children,
   className,
   icon,
   iconPosition = 'left',
-  disabled,
+  size = 'medium',
+  variant = 'normal',
   expanded,
   pressed,
-  onClick = () => {},
-  size = 'medium',
-  style = {},
-  title,
-  variant = 'normal',
+  ...restProps
 }) {
-  const otherAttributes = {};
-  if (typeof disabled === 'boolean') {
-    otherAttributes.disabled = disabled;
-  }
-  if (typeof title !== 'undefined') {
-    otherAttributes.title = title;
-    otherAttributes['aria-label'] = title;
-  }
-
-  if (typeof expanded === 'boolean') {
-    otherAttributes['aria-expanded'] = expanded;
-  }
-  if (typeof pressed === 'boolean') {
-    otherAttributes['aria-pressed'] = pressed;
-  }
+  restProps['aria-expanded'] = expanded;
+  restProps['aria-pressed'] = pressed;
+  restProps['aria-label'] = restProps.title;
 
   return (
     <button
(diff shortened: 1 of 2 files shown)


lyzadanger:
@esanzgar Do you want to rebase to fix conflicts and then I'll review?

esanzgar:
Fixed the conflict.

src/components/buttons.js
```diff
@@ -36,34 +37,18 @@ import { SvgIcon } from './SvgIcon';
...
-    otherAttributes['aria-label'] = title;
-  }
-
-  if (typeof expanded === 'boolean') {
-    otherAttributes['aria-expanded'] = expanded;
-  }
-  if (typeof pressed === 'boolean') {
-    otherAttributes['aria-pressed'] = pressed;
-  }
+  restProps['aria-expanded'] = expanded;
```
esanzgar, line 49: If `expanded` is undefined the prop is ignored

src/components/test/buttons-test.js
```diff
@@ -124,6 +124,24 @@ function addCommonTests({ componentName, createComponentFn, withIcon = true }) {
         attributeName: 'disabled',
         attributeValue: '',
       },
+      {
+        propName: 'expanded',
```
esanzgar, line 128: This test shows that for undefined prop the attribute is not set.

src/components/buttons.js
```diff
@@ -26,6 +26,7 @@ import { SvgIcon } from './SvgIcon';
  * @typedef IconButtonBaseProps
  * @prop {string} icon - Icon is required for icon buttons
  * @prop {string} title - Title is required for icon buttons
+ * @prop {never} [children] - children are not allowed (use LabeledButton instead)
```
lyzadanger, line 29: This is so handy! I had been wondering how to "remove" a property from a typedef and hadn't been able to google this up.

src/components/buttons.js
lyzadanger: Thank you; much simpler.

src/components/buttons.js
```diff
@@ -36,34 +37,18 @@ import { SvgIcon } from './SvgIcon';
...
-  expanded,
-  pressed,
-  onClick = () => {},
   size = 'medium',
-  style = {},
-  title,
   variant = 'normal',
+  expanded,
+  pressed,
+  ...restProps
```
lyzadanger, line 47: I wonder if hiding, e.g., `onClick` in these `restProps` makes it a little bit opaque...

src/components/buttons.js
```diff
@@ -89,9 +70,8 @@ function ButtonBase({
  *
  * @param {IconButtonProps} props
  */
-export function IconButton(props) {
-  const { className = 'IconButton', ...restProps } = props;
-  const { icon } = props;
+export function IconButton({ className = 'IconButton', ...restProps }) {
```
lyzadanger, line 73: I definitely like the `restProps` usage here 👍🏻

lyzadanger (approved):
With the caveat that I cannot preview these buttons visually in this package until we've got the pattern library in place (it's on the list!), these changes look good.

The only thing I wasn't 100% sure about was hiding meaningful props in `ButtonBase` in `...restProps`, e.g. the `onClick` property, which is rather fundamental. Using props spread for properties that are just "passing through" components seems like a nice pattern, I'm just wondering if we should stay more explicit in the components where they're actually _used_. Not a strong opinion.

src/components/buttons.js
esanzgar: I was recently reminded about `never` watching this video: https://www.youtube.com/watch?v=vXh4PFwZFGI

src/components/buttons.js
esanzgar: Personally, I think it makes a better distinction between the props that are allowed but not directly used or manipulated by `ButtonBase` (`children`, `disabled`, `onClick`, `style`, `title`), which go to `restProps`, from the props that are exclusive of `ButtonBase` (`className`, `icon`, `iconPosition`, `expanded`, `pressed`, `size`)

PR #25, opened 2021-04-16
Update package release documentation

lyzadanger:
This PR updates the release documentation based on some recent discoveries in the process. It adds some documentation about getting the tag to point at a single commit that contains both the version bump and the changelog edits, and a bit more detail about publishing and checking that the publishing worked as expected.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,48 +1,73 @@
 # Release process
 
-The release process in broken up into 3 steps.
+## 1. Version and tag
 
-1. Increment the version and add a new tag
-2. Modify the changelog
-3. Merge the release branch and create a "github release"
+We use [Semantic Versioning](https://semver.org/#semantic-versioning-200).
 
-## 1. Increment the version and add a new tag
+1. Create a new working branch
+2. Run
 
-To update the version in package.json and create a tag run:
+   ```shell
+   $ yarn version
+   ```
 
-```shell
-$ yarn version
-```
+   This will update the `package.json` version, create a commit and add
+   an annotated version tag.
 
-The version number should be updated following [Semantic Versioning](https://semver.org/#semantic-versioning-200). The only change should be to package.json, and there should be a new tag on the commit prefixed with a "v".
+## 2. Update the changelog
 
-## 2. Modify the changelog
+We use conventions from [keepachangelog](https://keepachangelog.com/en/1.0.0/).
 
-`CHANGELOG.md` highlights changes that are relevant for consumers of the package. Our changelog follows [keepachangelog](https://keepachangelog.com/en/1.0.0/).
+1. Run
 
-### Steps to append to the changelog
+   ```shell
+   $  gulp changelog
+   ```
 
-Immediately after a version commit & tag, run the following command to create a draft list of changes relating to the new version.
+   This outputs some suggested changelog content based on `git` history
 
-```shell
-$ gulp changelog
-```
+2. Edit `CHANGELOG.md` and commit changes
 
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -1,48 +1,68 @@
...
 
-## 3. Create a new GitHub release
+   ```shell
+   $ git tag -af <tag> <commit>
+   ```
 
-Create a [new github release](https://github.com/hypothesis/frontend-shared/releases/new/). The title of the release should be identical to the tag name and the message can copied from the result of step #2. When you are done, click "Publish".
+3. Push
+
+   Push both the branch and the tag
```
robertknight, line 52: I suggest spelling out the CLI command for doing this.

docs/releases.md
```diff
@@ -1,48 +1,68 @@
...
+   - Title: use the tag name
+   - Message/Description: Paste in changelog changes
+   - Binary: nope
+
+   Publishing this will kick off a Google Action that will publish the package.
+
+## 5. Check your work
+
+- Check to see if the version has been updated on [npm](https://www.npmjs.com/package/@hypothesis/frontend-shared)
+- Check the contents of the package on UNPKG. The URL pattern is `https://unpkg.com/browse/@hypothesis/frontend-shared@<version>/`, e.g. [https://unpkg.com/browse/@hypothesis/frontend-shared@1.11.0/]
```
robertknight, line 68: The link at the end here doesn't need to be surrounded by braces. GitHub will auto-link HTTPS URLs outside of code fences.

docs/releases.md
```diff
@@ -1,48 +1,68 @@
...
+## 4. Merge and release
+
+1. Go through the PR/merge cycle for the version-update branch
+2. Create a [new GitHub release](https://github.com/hypothesis/frontend-shared/releases/new/)
+
+   - Title: use the tag name
+   - Message/Description: Paste in changelog changes
+   - Binary: nope
+
+   Publishing this will kick off a Google Action that will publish the package.
```
robertknight, line 63:
```suggestion
   Publishing this will kick off a GitHub Action that will publish the package.
```

😉

docs/releases.md
```diff
@@ -1,48 +1,68 @@
...
+   ```
 
-```shell
-$ yarn version
-```
+   This will update the `package.json` version, create a commit and add
+   an annotated version tag.
 
-The version number should be updated following [Semantic Versioning](https://semver.org/#semantic-versioning-200). The only change should be to package.json, and there should be a new tag on the commit prefixed with a "v".
+## 2. Update the changelog
```
robertknight, line 17: Since you know that what the new version of the package will be and the CHANGELOG entry, it would be possible to re-order the process to avoid the need for step 3. This would require some changes to the way `gulp changelog` works to output changes since the last release. The process would then be:

1. Run `gulp changelog` to get changes since last release and copy the output
2. Edit `CHANGELOG.md`, paste in the change list and edit as needed
3. Run `git add CHANGELOG.md` to stage the changelog for the next commit
4. Run `yarn version` to update the version number in `package.json`, commit the `package.json` and `CHANGELOG.md` updates and tag the result
5. Push the resulting commit and tag to GitHub

robertknight (approved):
I noted a couple of minor mistakes, otherwise this is good to merge.

I did note though that the overall process could be optimized to avoid the need for the third stage where you squash the commits and re-create the tag. This would also avoid a hazard due to the use of the `--force` argument when creating the tag, which could override the tag if you using the wrong value. Perhaps you or @LMS007 could take a look at that when convenient?

docs/releases.md
lyzadanger: Blush.

docs/releases.md
lyzadanger: > This would require some changes to the way gulp changelog works to output changes since the last release. 

That felt a little out of scope for the immediate term.

PR #27, opened 2021-04-23
Add "cancel" icon SVG 

LMS007:
This is copied from the client and would need to be removed there in a subsequent PR. This is needed for dialogs in both LMS and the client. This is not used in the frontend-shared package, so no need to add to registered icons or any code updates. This just gets bundled up so it can be imported in other apps.

Commit
------------
Previously, this icon only existing in the client, but also has uses inside LMS. Moving the icon to this repository allows it to be shared in both places without duplication.


robertknight (approved):
I assume that both the client and LMS apps had exactly the same `cancel.svg` icon in their source code?

lyzadanger:
@LMS007 I'm having trouble understanding the PR description and title. Could you please clarify a little what the intent of this PR is?

LMS007:
> @LMS007 I'm having trouble understanding the PR description and title. Could you please clarify a little what the intent of this PR is?

np, fixed

PR #28, opened 2021-04-23
Add `buttonRef` to buttons

LMS007:
In LMS, we need to pass along a button ref in order to set focus() on the button when a dialog opens. In order to pass along the ref, we need to forward the ref down to the HTML component. Preact consumes the `ref` prop so we can't override that directly and `forwardRef()` is only included in `preact/compact` lib which has some adverse side effects--so we should avoid that import. So the simple solution is to use a custom named prop (`buttonRef`) to pass along the ref.

### DIFF
--- a/src/components/buttons.js
+++ b/src/components/buttons.js
@@ -4,6 +4,7 @@ import { SvgIcon } from './SvgIcon';
 
 /**
  * @typedef ButtonProps
+ * @prop {import('preact').Ref<HTMLButtonElement>} [buttonRef]
  * @prop {import("preact").ComponentChildren} [children]
  * @prop {string} [className]
  * @prop {string} [icon] - Name of `SvgIcon` to render in the button
@@ -37,6 +38,7 @@ import { SvgIcon } from './SvgIcon';
  * @param {ButtonProps} props
  */
 function ButtonBase({
+  buttonRef,
   className,
   icon,
   iconPosition = 'left',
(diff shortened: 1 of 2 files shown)


src/components/test/buttons-test.js
```diff
@@ -20,6 +21,13 @@ function addCommonTests({ componentName, createComponentFn, withIcon = true }) {
       });
     }
 
+    it('passes along a `ref` to the button element', () => {
+      const buttonRef = createRef();
+      createComponentFn({ ref: buttonRef });
+
+      assert.isTrue(buttonRef.current instanceof Node);
```
LMS007, line 28: This would fail if we don't forward the ref because the object would be a Preact component, not a Node.

lyzadanger:
@LMS007 It appears that this PR branch contains the changes from #27 . Can you rebase or otherwise clarify?

LMS007:
> @LMS007 It appears that this PR branch contains the changes from #27 . Can you rebase or otherwise clarify?

done

src/components/buttons.js
```diff
@@ -52,6 +54,7 @@ function ButtonBase({
 
   return (
     <button
+      ref={buttonRef}
```
lyzadanger, line 57: What is the behavior of this component if `buttonRef` is `undefined`?

lyzadanger (changes requested):
This seems mostly innocuous, but I'd like to see a test for the `undefined` path of `buttonRef` just to sew up some loose ends.

lyzadanger (approved):
👍🏻

PR #29, opened 2021-04-23
More accurately define the onClick event type

LMS007:
I noticed some typing errors in LMS when using buttons, this fixes the issue.


lyzadanger (approved):
LGTM

PR #30, opened 2021-04-23
Add hover cursor styling to buttons

LMS007:
LMS does not have the same defaults as the client. The hover styling in the client is done in reset.scss and LMS has does not have that.  The styling should probably be in the frontend-shared for buttons in the first place. I filed a ticket for the client to clean up styling there.

relates to https://github.com/hypothesis/client/issues/3327


lyzadanger:
> LMS does not have the same defaults as the client. The hover styling in the client is done in reset.scss and LMS has does not have that. 

The browser should provide an appropriate cursor for `button` elements unless it has been interfered with. Can you clarify exactly what is going on in LMS? Is something else overriding the browser's default cursor styling for `button` elements?

Issue #20, opened 2021-02-15
Extract Pattern Library to `frontend-shared`


lyzadanger:
`frontend-shared` should own the pattern library source and provide it (optionally) to users of the package.

Updated from a spike card, as implementation is in progress.

PR #31, opened 2021-04-27
Add reset styles, mixins

lyzadanger:
This PR adds reset styles and mixins copied from the client. These reset styles and mixins are a dependency for the pattern library.

The new SASS modules follow our recent conventions in `styles/util/reset`:

* `_index` defines the public API
* `_mixins` is what it sounds like
* `_styles` outputs associated styles

The `styles/reset.scss` module is provided because I don't feel like I want to make the jump to packaging any `reset` styles into the main `index` module. It is at the top module level because it is intended to be used by users of the package.

I would like to keep moving quickly right now, because there are many moving parts in play with extracting the pattern library and other related things in flight, but also acknowledge some inconsistencies that need to be sorted out in subsequent work:

* `mixins/focus.scss` and `util/reset` organization are a mismatch
* `styles/reset.scss` at the top level does not have a leading underscore in the module name, while `_index.scss` does. I believe there should be no underscore in this case, and, in fact, I had to _remove_ the underscore to get styles to output as I'd expect when consuming this module from the `client` (linked locally). But I acknowledge the inconsistency and lack of clarity about SASS underscore conventions.
* package users should be updated to use these `reset` styles and eliminate local SASS source. I can open a ticket to track this.

Part of #20

### DIFF
--- a/styles/base/_reset.scss
+++ b/styles/base/_reset.scss
@@ -1,0 +1,114 @@
+@use '../mixins/reset';
+
+/*
+  Consistency fixes
+  adopted from http://necolas.github.com/normalize.css/
+  */
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  height: 100%;
+  text-size-adjust: 100%;
+}
+
+body {
+  min-height: 100%;
+  font-size: 100%;
+  margin: 0;
+}
+
+sub,
+sup {
+  font-size: 75%;
+  line-height: 0;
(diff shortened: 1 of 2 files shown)


lyzadanger:
Putting in draft because I'm having further thoughts.

lyzadanger:
The re-work of this is different enough to warrant a fresh PR for less confusion for the reviewer.

PR #33, opened 2021-04-27
Add shared base element styling

lyzadanger:
Again a dependency for pattern library styling and again copied directly from the client.

Depends on #32 

Part of #20


LMS007 (approved):
👍

PR #36, opened 2021-04-27
Establish Pattern Library in this repo

lyzadanger:
Depends on #38

This PR contains the needed bits and bobs to get the pattern library going (source, build, bundle, serve) in this repository, and leaves some "edges" that will allow the pattern library to be extended by users of the package.

### Known rough edges

* The pattern library currently doesn't serve at the root level of the port it's served on, but instead is nested under `ui-playground`. **This is annoying**. I was having some woes with the `router` when I tried to serve at web root:

    Screen Shot 2021-04-27 at 1 09 44 PM
    
    And didn't feel like tracking it down just now. For now, the landing page provides a simple link to the pattern library.
* ~~The location of the `serve` module for the pattern library is temporary and almost certain to move. It doesn't belong with packaged source, but instead, likely, with the build/tooling of the package.~~ Addressed.
* How and if the bundled JS and CSS for the pattern library should be published with the package. 
* What default port to use here. I chose `4001` out of the air.

### Try it out

Check out/pull this branch and run `make dev` and open the logged URI in your browser.

Part of #20


.eslintignore
```diff
@@ -1 +1,2 @@
 lib/**
+build/**
```
lyzadanger, line 2: Don't lint or prettify stuff output to `build`...

Makefile
```diff
@@ -4,7 +4,7 @@ default: help
 .PHONY: help
 help:
 	@echo "make help              Show this help message"
-	@echo "make dev               Run the pattern library application (TODO)"
+	@echo "make dev               Run the pattern library app in a development server"
```
lyzadanger, line 7: Adding `make dev` target to act more like other projects.

gulpfile.js
```diff
@@ -1,12 +1,34 @@
 /* eslint-env node */
+('use strict');
```
lyzadanger, line 2: REVIEW ME (changes based on other projects).

scripts/gulp/create-bundle.js
```diff
@@ -0,0 +1,236 @@
+/* eslint-env node */
```
lyzadanger, line 1: COPIED. No review needed.

scripts/gulp/create-style-bundle.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
```
lyzadanger, line 1: COPIED. No review needed.

scripts/gulp/minify-stream.js
```diff
@@ -0,0 +1,56 @@
+/* eslint-env node */
```
lyzadanger, line 1: COPIED. No review needed.

src/pattern-library/components/PatternPage.js
```diff
@@ -0,0 +1,137 @@
+import { toChildArray } from 'preact';
```
lyzadanger, line 1: COPIED. No review needed.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -0,0 +1,116 @@
+import { SvgIcon } from '../../components/SvgIcon';
```
lyzadanger, line 1: COPIED. No review needed.

src/pattern-library/components/patterns/SharedButtonPatterns.js
```diff
@@ -0,0 +1,383 @@
+import { IconButton, LabeledButton, LinkButton } from '../../../';
```
lyzadanger, line 1: COPIED. No review needed.

src/pattern-library/icons.js
```diff
@@ -0,0 +1,13 @@
+// @ts-nocheck
```
lyzadanger, line 1: NEW, but based on other "icons" modules in other projects.

src/pattern-library/index.js
```diff
@@ -0,0 +1,4 @@
+// Entry point for default pattern-library bundle
```
lyzadanger, line 1: NEW. Separate bundle entry point.

src/pattern-library/router.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useMemo, useState } from 'preact/hooks';
```
lyzadanger, line 1: COPIED. No review needed.

src/pattern-library/templates/index.mustache
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
lyzadanger, line 1: TEMPORARY: because the pattern library is not served at the web root.

src/pattern-library/templates/pattern-library.mustache
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
lyzadanger, line 1: COPIED.

src/pattern-library/util/jsx-to-string.js
```diff
@@ -0,0 +1,93 @@
+/**
```
lyzadanger, line 1: COPIED. No review needed.

styles/pattern-library.scss
```diff
@@ -0,0 +1,167 @@
+@use 'base/reset';
```
lyzadanger, line 1: COPIED. No review needed; paths at top updated.

gulpfile.js
```diff
@@ -40,6 +62,17 @@ function parseCommandLine() {
 const karmaOptions = parseCommandLine();
 const { run } = require('./scripts/gulp/run');
 
+function runKarma(done) {
```
lyzadanger, line 65: This already existed but I moved it within the module.

gulpfile.js
```diff
@@ -67,7 +100,7 @@ gulp.task('changelog', async () => {
  *
  * nb. This is only used for unit tests that need CSS to verify accessibility requirements.
  */
-gulp.task('build-css', done => {
+gulp.task('build-test-css', done => {
```
lyzadanger, line 103: I renamed this task because we have other CSS-building needs now.

src/pattern-library/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
```
lyzadanger, line 1: COPIED.

src/pattern-library/app.js
```diff
@@ -0,0 +1,21 @@
+// Entry point for extending the pattern library
```
lyzadanger, line 1: This is...sorta new but mostly just adapted. Wraps the app startup for use by other projects, eventually.

lyzadanger:
See also: https://github.com/hypothesis/client/pull/3348

LMS007 (changes requested):
Thanks Lyza, lots of moving parts here, and this mostly worked for me right away, but I did have some questions. Looks like I got a mine type error and the styles did not load. I will look into that later--it may be local to me.

Screen Shot 2021-04-27 at 11 31 52 AM

It looks like these new tasks are completely orthogonal to the tasks that build the /lib folder. Is that correct? I believe we could have gone either way and used lib/ as a base for the bundling OR, just use the src/ files directly and transpile everything specifically for the pattern lib. I feel like the latter is more clear so I think that's a good choice. Is that correct? I did had an older PR that attempted to prep some of this but it looks like its irreverent now.

https://github.com/hypothesis/frontend-shared/pull/6

Secondly, I don't see a use for this to exist outside "development" so I'm not even sure we need a  the concept of "production". I believe this is always going to be dev right? Unless we host this pattern lib somewhere live, but even so, it would likely be helpful to not have minified code even in that case imo. If thats true, I wonder if we can trim down some of the checks for IS_PRODUCTION_BUILD and just never minify the code.  The concept of minify implies that could be used in production so that's a little confusing to me.

I'd also like to see very clear distinction between the tasks that are bundling the pattern lib and the ones that create the lib/. At present we have a script called `build-js` in package.json and a task called `build-js` in gulpfile and they do different things. So that just means `yarn build-js` and `gulp build-js` don't do the same thing here.  I'd like to see a different name so we can clearly understand the discussion between release bundling into the lib/ and dev bundling for the express server inside of /build

Finally, I think the pattern lib code may not be well suited for the /src folder. The intention of the src folder was for the things that this repo exports. The pattern lib is more an internal tool to visualize the things. At present, when you run `yarn build-js` it also transpiles the pattern lib files and you can see them inside /lib which we don't want. I think instead of making complex rules for ignoring files, we just need a clear distinction  between files for /lib and the other stuff. Perhaps we can just have a root level folder called `pattern-library`?

lyzadanger:
> Finally, I think the pattern lib code may not be well suited for the /src folder. The intention of the src folder was for the things that this repo exports.

The `pattern-library` source _is_ exported. We'll need to use it, e.g. from the client.

lyzadanger:
> At present, when you run yarn build-js it also transpiles the pattern lib files and you can see them inside /lib which we don't want.

I'm pretty sure we do?

lyzadanger:
> It looks like these new tasks are completely orthogonal to the tasks that build the /lib folder. Is that correct?

I'm not sure what this question means, exactly.

lyzadanger:
@LMS007 I want to correct what I think is a bit of a misunderstanding: the pattern library is not private/internal to this project. It will be imported and extended by other projects—immediately in the client, for example. The intention is that the pattern library source should be part of the published package, but should have a different entry point than the "main" part of the package, as it won't be used by all package consumers. I delineated in the PR description that there are a few rough edges yet in terms of what is bundled and packaged, and will continue to tighten it up. The `serve` module needs to move out of `src`, for example.

lyzadanger:
> Looks like I got a mine type error and the styles did not load. I will look into that later--it may be local to me.

We should debug this. What browser are you using and what URL are you attempting to load in your browser? There are some odd "spaces" in the URLs in the console that seem a little bit fishy, but it's also possible that the `express` static route handling needs to be tweaked here; I can't tell right off.

lyzadanger:
@LMS007 I've reproduced the CSS issue and it looks like a race condition/ordering problem with some gulp tasks. Working on a fix for yah/us.

lyzadanger:
@LMS007 I pushed a commit that renames some `gulp` tasks for clarity and fixes the CSS-bundling ordering issue you ran into.

LMS007:
I believe I had a misunderstanding about the pattern lib expectations. I thought It was for internal usage, but if we are going to export it, then ignore my suggestions about separating it from /lib.

lyzadanger:
Update: I have rebased this PR on #37, which breaks out the rote copy-n-paste of pattern-library source so that it isn't cluttering up this branch.

lyzadanger:
This PR is superseded by #39 (and also #37 and #38).

PR #37, opened 2021-04-28
Add pattern-library source JS and SASS (Pattern Library 1/n)

lyzadanger:
To make #36 more human-manageable, this PR breaks off a chunk of it that is simply (yeah, I said _simply_) copying over components, JS source, and SASS source from the `client` repo. The only changes from what exists now in the `client` are some paths.

The JS and SASS will be included in the published package.

Part of #20


LMS007 (approved):
👍

PR #38, opened 2021-04-28
Add dependencies and config needed for building pattern library assets (Pattern Library 2/n)

lyzadanger:
Depends on #37 

This PR is another boilerplate chunk of #36 : it adds gulp scripts (copied from other repos) and dependencies that will be needed to build pattern-library assets. These will be used in subsequent PRs/changes. These should not require much human oversight...

Part of #20


scripts/gulp/minify-stream.js
```diff
@@ -0,0 +1,56 @@
...
+ * transform, and also to be applied to the output of the whole bundle to
+ * compress the module infrastructure that Browserify adds.
+ *
+ * @example
+ *   browserify(['src/app.js'])
+ *     .transform({ global: true }, minifyStream) // Minify individual modules
+ *     .pipe(minifyStream()) // Minify the code added by Browserify
+ *     .pipe(output);
+ */
+function minifyStream() {
```
LMS007, line 21: Lots of copied code, but I'm still unclear if we will ever minify, but I don't quit have the full context in my head. It almost might make sense if we start moving universal gulp tasks and subsequent functions upstream to the toolkit (for example). Maybe that could come down the road.

LMS007 (approved):
👍

PR #39, opened 2021-04-28
Add entry points, build and bundle scripts for Pattern Library (Pattern Library 3/n)

lyzadanger:
This PR contains a packaged entry point, scripts, templates and build for both packaging the pattern library and running it in a local dev server. It supersedes #36  and depends on #37 and #38

To run the pattern library:

`make dev`

See https://github.com/hypothesis/client/pull/3348 for an example of how the `client` can import, use and extend the pattern library.

Part of #20 

## Concepts

* Users of the `package` should be able to `import` the pattern-library and use its `startApp` function, optionally passing in additional routes, icons and `baseURL`. See https://github.com/hypothesis/client/pull/3348 for example.
* The pattern-library should serve locally via `make dev` and watch for changes. To do this, it bundles its needed JS and CSS into `build` and serves this via `express.static` within the `serve-pattern-library` express webserver.
* At present, it is the responsibility of the consuming project to bundle and serve their own JS/CSS if they extend the pattern library (again, see the example in https://github.com/hypothesis/client/pull/3348), and to provide their own templating. This slotted in easily for the `client`. It could be a few steps (and needs to be documented) for other projects, but the pain/complexity of trying to automate this and include, e.g., templates and webserver scripts in the package seemed to outweigh the benefits for now. That is: premature extrapolation that we're going to be wanting to serve the pattern library in the local dev servers of a lot of various projects...we don't really know that.


gulpfile.js
```diff
@@ -80,18 +113,48 @@ gulp.task('build-css', done => {
...
+      input: entry,
+      output: `${STYLE_DIR}/${path.basename(entry, path.extname(entry))}.css`,
+      minify: IS_PRODUCTION_BUILD,
+    })
+  );
+  return Promise.all(bundles);
+});
+
+gulp.task(
+  'watch-css',
```
LMS007, line 147: I can see some confusion here because the watchers only rebuild into /build, but not for /lib as near as I can tell That means the pattern lib will live-load but not any consumers of the shared-package such as lms or the client.

We may want to ass a second set of watchers later on for this and migrate the scripts from package.json into gulp.

LMS007 (approved):
👍

gulpfile.js
lyzadanger: > I can see some confusion here because the watchers only rebuild sass and js into /build, but not for /lib as near as I can tell That means the pattern lib will live-load but not any consumers of the shared-package such as lms or the client.

I'm sorry, I'm having trouble parsing this—could you possibly rephrase?

LMS007: Ignore me. Thinking more on this, because it's inside the context of `make dev`, it does only make sense that build/ gets updated with changes.

Issue #26, opened 2021-04-19
Define a checkbox design pattern


lyzadanger:
The consensus-approved mockups for the LMS Small Groups assignment-configuration workflow (https://github.com/hypothesis/lms/issues/2523) include a checkbox. I don't believe we have a checkbox pattern: we should define one.

PR #43, opened 2021-04-28
Add Checkbox component and patterns

lyzadanger:
This PR adds a very basic `Checkbox` component, tests and some simple patterns in the pattern library.

~~Putting this in draft because of an issue with the `axe` a11y tests—failure on "no-explicit-label", which I believe to be possibly...wrong.~~ (Update: After discussion with team, we've decided to look at the test failure as a separate issue—I believe this component meets general a11y guidelines).

Aside:

I can make the a11y tests pass if:

* I wrap the `<input>` field in the `<label>` element, OR
* I put the `<label>` element before the `<input>` (source order)

This bugs me, because both break what is otherwise completely appropriate default styling in the browser. And I’m fairly sure there’s nothing wrong with the original markup (it follows MDN examples, for one thing, exactly). But maybe I have a thing or two to learn?

This task was good for dogfooding the new pattern-library setup and `frontend-shared` workflow. Seemed to go pretty well.

Oh, you'll note there's no actual _styling_. That's by, pun, design. I'm not sure yet if we want to impose any visual styling on our checkboxes just yet—we certainly can if we desire. In the case where this checkbox is to be used in the immediate future—the LMS assignment-configure interface—the interface we're modeling after doesn't style their checkboxes. So I've opted to start simple. But: world's our oyster.

## To see it

Pull this branch and run `make dev`; go here http://localhost:4001/ui-playground and look at the "Forms" patterns.

Part of https://github.com/hypothesis/frontend-shared/issues/26

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -1,16 +1,25 @@
 /**
  * @typedef CheckboxBaseProps
  * @prop {string} name - The `name` of the checkbox.
- * @prop {import('preact').Ref<HTMLInputElement>} [inputRef]
+ * @prop {import('preact').Ref<HTMLInputElement>} [inputRef] - Access to the input
+ *    element in case a parent element wants for example to focus on it.
  * @prop {(checked: boolean) => void} [onToggle] - Callback when checkbox is
  *   checked/unchecked
  * @prop {never} [type] - Type is always 'checkbox'
  * @prop {never} [children] - Children are not allowed
  *
- * @typedef {Omit<import('Preact').JSX.HTMLAttributes<HTMLInputElement>, 'onToggle'> & CheckboxBaseProps} CheckboxProps
+ * The props for Checkbox component extends and narrows the attributes of the native input element.
+ * `onToggle` event should only be associated to HTMLDetailsElement, but Preact is not very strict with types.
+ * We omit the `onToggle` because it clashes with our definition.
+ * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLInputElement>, 'onToggle'> & CheckboxBaseProps} CheckboxProps
+ */
+
+/**
+ * @typedef LabeledCheckboxBaseProps
+ * @prop {import('preact').ComponentChildren} children - Label text or elements
+ * @prop {'before'|'after'} [position] - Position the label before or after the checkbox. Defaults to 'before'.
  *
- * @typedef LabeledCheckboxProps
- * @prop {string} label - Label text
+ * @typedef {Omit<CheckboxProps, 'children'> & LabeledCheckboxBaseProps} LabeledCheckboxProps
  */
 
 /**
@@ -20,7 +29,7 @@
  */
 export function Checkbox({ inputRef, onToggle, onClick, ...restProps }) {
   /**
-   * @param {import('Preact').JSX.TargetedMouseEvent<HTMLInputElement>} event
+   * @param {import('preact').JSX.TargetedMouseEvent<HTMLInputElement>} event
    * @this HTMLInputElement
    */
   function onPressed(event) {
@@ -39,14 +48,20 @@ export function Checkbox({ inputRef, onToggle, onClick, ...restProps }) {
 /**
  * A labeled checkbox input
  *
- * @param {CheckboxProps & LabeledCheckboxProps} props
+ * @param {LabeledCheckboxProps} props
  */
-export function LabeledCheckbox({ label, id, ...restProps }) {
+export function LabeledCheckbox({
+  children,
+  position = 'before',
+  id,
+  ...restProps
+}) {
   id ??= restProps.name;
   return (
     <>
-      <label htmlFor={id}>{label}</label>
+      {position === 'before' && <label htmlFor={id}>{children}</label>}
       <Checkbox id={id} {...restProps} />
+      {position === 'after' && <label htmlFor={id}>{children}</label>}
     </>
   );
 }
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -1,0 +1,58 @@
+import { useEffect, useRef, useState } from 'preact/hooks';
+
+const noop = () => {};
+
+/**
+ * @typedef CheckboxProps
+ * @prop {boolean} [defaultChecked] - Should the checkbox be selected/checked
+ *  by default?
+ * @prop {string} name - The `name` of the checkbox. Will be used as `id` if
+ *   `id` not provided
+ * @prop {string} [id] - HTML ID for the checkbox element
+ * @prop {string} label - Checkbox label
+ * @prop {(checked: boolean) => any} [onChanged] - Callback when checkbox is
+ *   checked/unchecked
+ */
+
+/**
+ * A checkbox input
+ *
+ * @param {CheckboxProps} props
+ */
+export function Checkbox({
+  defaultChecked = false,
+  id,
+  name,
+  label,
+  onChanged,
+}) {
+  let [isChecked, setChecked] = useState(defaultChecked);
+  onChanged = onChanged ?? noop;
+  id = id ?? name;
+
+  const wasChecked = useRef(isChecked);
+  useEffect(() => {
+    if (wasChecked.current !== isChecked && typeof onChanged === 'function') {
+      wasChecked.current = isChecked;
+      onChanged(isChecked);
+    }
+  }, [isChecked, onChanged]);
+
+  function onChange(event) {
+    const isChecked = event.target.checked;
+    setChecked(isChecked);
+  }
+
+  return (
+    <>
+      <input
+        type="checkbox"
+        id={id}
+        name={name}
+        checked={isChecked}
+        onChange={e => onChange(e)}
+      />
+      <label htmlFor={id}>{label}</label>
+    </>
+  );
+}
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -1,17 +1,16 @@
-import { useEffect, useRef, useState } from 'preact/hooks';
-
-const noop = () => {};
-
 /**
- * @typedef CheckboxProps
- * @prop {boolean} [defaultChecked] - Should the checkbox be selected/checked
- *  by default?
- * @prop {string} name - The `name` of the checkbox. Will be used as `id` if
- *   `id` not provided
- * @prop {string} [id] - HTML ID for the checkbox element
- * @prop {string} label - Checkbox label
- * @prop {(checked: boolean) => any} [onChanged] - Callback when checkbox is
+ * @typedef CheckboxBaseProps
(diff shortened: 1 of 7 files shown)


src/components/Checkbox.js
```diff
@@ -0,0 +1,59 @@
+//import classnames from 'classnames';
```
lyzadanger, line 1: Derp.

lyzadanger: (I fixed this)

src/components/Checkbox.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {CheckboxProps} props
+ */
+export function Checkbox({
+  defaultChecked = false,
+  id,
+  name,
+  label,
+  onChanged,
+}) {
+  let [isChecked, setChecked] = useState(defaultChecked);
```
lyzadanger, line 29: Note: the state modeling in this component is based on the client's `Menu` component (roughly)...just trying to reuse an established pattern.

src/components/Checkbox.js
```diff
@@ -0,0 +1,58 @@
+import { useEffect, useRef, useState } from 'preact/hooks';
+
+const noop = () => {};
+
+/**
+ * @typedef CheckboxProps
```
esanzgar, line 6: I wonder if we should restrict the type JSX.Intrinsic

src/pattern-library/components/patterns/SharedFormPatterns.js
```diff
@@ -0,0 +1,52 @@
+import { Checkbox } from '../../../';
+
+import {
+  PatternPage,
+  Pattern,
+  PatternExamples,
+  PatternExample,
+} from '../PatternPage';
+
+function gimmeSandwich(isChecked) {
```
esanzgar, line 10: That's a bit unusual. Is there a reason why you don use `useState` inside the `SharedFormComponent` component?

esanzgar (changes requested):
The accessibility test is failing because the `<label>` should be located first.

I suggest a generalisation and simplification of the component in this way:

```
/**
 * @typedef CheckboxProps
 * @prop {import('preact').Ref<HTMLInputElement>} [inputRef]
 * @prop {string} name - The `name` of the checkbox. Will be used as `id` if
 *   `id` not provided
 * @prop {string} label - Checkbox label
 * @prop {(checked: boolean) => void} [onToggle] - Callback when checkbox is
 *   checked/unchecked
 *
 */

/**
 * A checkbox input
 *
 * @param {import('Preact').JSX.HTMLAttributes<HTMLInputElement> & CheckboxProps} props
 */
export function Checkbox({
  inputRef,
  id,
  name,
  label,
  onToggle,
  onChange,
  ...restProps
}) {
  id ??= name;

  /**
   * @callback
   * @params {import('Preact').JSX.TargetedEvent<HTMLInputElement, Event>} event
   */
  function onPressed(event) {
    /** @type {boolean} */
    const checked = event?.target?.['checked'] || false;
    onToggle?.(checked);
    // @ts-ignore
    onChange?.(event);
  }

  return (
    <>
      <label htmlFor={id}>{label}</label>
      <input
        ref={inputRef}
        type="checkbox"
        id={id}
        name={name}
        onChange={onPressed}
        {...restProps}
      />
    </>
  );
}
```

I would use as much as possible the native preact API for the `<input type="checkbox">`. For example, I would use `checked` property to set the initial state rather than `defaultChecked`:
https://github.com/preactjs/preact/issues/2762#issuecomment-695361838

esanzgar:
I also recommend to add a ref to the input element.

src/components/Checkbox.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * @typedef CheckboxBaseProps
+ * @prop {string} name - The `name` of the checkbox.
+ * @prop {import('preact').Ref<HTMLInputElement>} [inputRef]
```
lyzadanger, line 4: Possible to document this a little more? What's the intent of this? I think I get it but want to feel more certain.

src/components/Checkbox.js
```diff
@@ -0,0 +1,52 @@
...
+ * @prop {(checked: boolean) => void} [onToggle] - Callback when checkbox is
+ *   checked/unchecked
+ * @prop {never} [type] - Type is always 'checkbox'
+ * @prop {never} [children] - Children are not allowed
+ *
+ * @typedef {Omit<import('Preact').JSX.HTMLAttributes<HTMLInputElement>, 'onToggle'> & CheckboxBaseProps} CheckboxProps
```
lyzadanger, line 10: For future-dev comprehension, it might be nice to document what's going on here. I grok it, but a little bit of hand-holding would seem friendlier.

src/components/Checkbox.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @typedef LabeledCheckboxProps
```
lyzadanger, line 12: The name of this typedef is a little bit misleading because the eventual props type ends up being `{CheckboxProps & LabeledCheckboxProps} props`

Can this be simplified/clarified?

src/components/Checkbox.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {CheckboxProps} props
+ */
+export function Checkbox({ inputRef, onToggle, onClick, ...restProps }) {
+  /**
+   * @param {import('Preact').JSX.TargetedMouseEvent<HTMLInputElement>} event
+   * @this HTMLInputElement
+   */
+  function onPressed(event) {
+    onToggle?.(event.currentTarget.checked);
+    // onChange expects `this` context to be of type `HTMLInputElement`
```
lyzadanger, line 28: Boy, that must have been "fun" to figure out!

src/components/Checkbox.js
```diff
@@ -0,0 +1,52 @@
...
+ * @prop {import('preact').Ref<HTMLInputElement>} [inputRef]
+ * @prop {(checked: boolean) => void} [onToggle] - Callback when checkbox is
+ *   checked/unchecked
+ * @prop {never} [type] - Type is always 'checkbox'
+ * @prop {never} [children] - Children are not allowed
+ *
+ * @typedef {Omit<import('Preact').JSX.HTMLAttributes<HTMLInputElement>, 'onToggle'> & CheckboxBaseProps} CheckboxProps
+ *
+ * @typedef LabeledCheckboxProps
+ * @prop {string} label - Label text
```
lyzadanger, line 13: I have a sense we might run into a need to have a label that is richer than a string in the future (e.g. contains a link, or an image, or something). We can probably defer dealing with this for now.

src/pattern-library/components/patterns/SharedFormPatterns.js
```diff
@@ -0,0 +1,40 @@
...
+          aria-label="seriously, just an emoji of sandwich"
+          style={{
+            fontSize: '2em',
+            visibility: show ? 'visible' : 'hidden',
+          }}
+        >
+          🥪
+        </span>
+        <PatternExamples>
+          <PatternExample details="A checkbox, defaulting to checked">
```
lyzadanger, line 28: Can you add an example that doesn't default to checked?

lyzadanger (commented):
Note: ARGH. GitHub will not let me approve or request-changes on this PR. I can only add a comment. 

Thanks for working through some of the challenges this ended up presenting! In general, this is looking good. On a high level:

We need to "meet the brief!" The design pattern we're intending to encapsulate comes from wireframes for LMS course-group selection (linked to from the GH issue driving this work). That design pattern has the checkbox input to the left of (before) the label:
image

That should be the default layout of a `LabeledCheckbox`. Can you make that happen?

The type definitions in `Checkbox`, while accurate and structured, are feeling on the brink of cumbersome to hold in one's head all at once. I added a few possible suggestions for how to aid future developers in getting their heads around it faster...

Thanks for working through all of this...

src/components/test/Checkbox-test.js
```diff
@@ -0,0 +1,106 @@
...
+import { checkAccessibility } from '../../../test/util/accessibility';
+
+import { LabeledCheckbox, Checkbox } from '../Checkbox';
+
+describe('Checkbox', () => {
+  function createComponent(props = {}) {
+    return mount(<Checkbox name="my-checkbox" {...props} />);
+  }
+
+  it('inputRef points to the input element', () => {
```
lyzadanger, line 12: The wording of this test description is a tad awkward.

src/components/test/Checkbox-test.js
```diff
@@ -0,0 +1,106 @@
...
+    inputRef.current.click();
+    assert.calledOnce(onClick);
+  });
+});
+
+describe('LabeledCheckbox', () => {
+  function createComponent(props = {}) {
+    return mount(
+      <div>
+        {/** wrapped on a div to pass the a11y test */}
```
lyzadanger, line 32: Thank you for this work and this comment!

src/components/test/Checkbox-test.js
```diff
@@ -0,0 +1,106 @@
...
+
+  it('contains a label after the input element (default position)', () => {
+    const inputRef = {};
+    const wrapper = createComponent({ inputRef });
+    const labelElement = wrapper.find('label');
+    assert.equal(labelElement.text(), 'My Action');
+    assert.equal(inputRef.current.nextElementSibling.tagName, 'LABEL');
+  });
+
+  it('contains a label before the input element ', () => {
```
lyzadanger, line 48: I think this test description could use an additional clause, to explain in which cases the label would come before the input (i.e. the `position` prop is set to `before`)...

src/components/test/Checkbox-test.js
```diff
@@ -0,0 +1,106 @@
...
+    const inputRef = {};
+    const wrapper = createComponent({ inputRef, position: 'before' });
+    const labelElement = wrapper.find('label');
+    assert.equal(labelElement.text(), 'My Action');
+    assert.equal(inputRef.current.previousElementSibling.tagName, 'LABEL');
+  });
+
+  it('contains children element in the label', () => {
```
lyzadanger, line 56: Perhaps: `it('renders children in a label element')` ?

lyzadanger (commented):
Once again, I cannot approve, only comment! Drat! But, consider this approved.

I noted a couple of awkward test descriptions, but nothing big!

src/components/test/Checkbox-test.js
esanzgar: Changed to "passes along a `ref` to the input element through `inputRef`" (copied from the button component)

src/components/test/Checkbox-test.js
esanzgar: Done.

src/components/test/Checkbox-test.js
esanzgar: Addressed, thanks.

esanzgar (approved):
Lyza gave the green light.

PR #44, opened 2021-04-29
Adjust SASS module file naming vis-a-vis underscores

lyzadanger:
This PR adjusts a couple of SASS module filenames for consistency. A README in the `styles` directory is forthcoming with more information, but in general:

* We want to use underscores for any SASS _partial_, that is, a chunk of SASS that will not output a standalone stylesheet
* We want to not use underscores for any SASS module that might be used as a standalone stylesheet; typically the "entry points" for SASS compilation.

We want to avoid using underscores in `@use` ("import") statements (and we have to this point, good). There is a subtle division here between module-file naming and module naming:

* `foo/_index.scss` and `foo/index.scss` -> Both of these module _file_ names are interpreted by SASS such that the module name becomes `foo/index` (or just `foo`).
* `mixins/focus.scss` and `mixins/_focus.scss` -> Both of these module _file_ names are interpreted by SASS such that the module name becomes `mixins/focus`.

Because of the division between module-file naming and module naming described above, the changes in this PR should not affect users of the package. I have tested the client against these changes.


LMS007:
Understood, and  I think the direction makes sense with the file names. I tested LMS too just out of caution, looks good!

LMS007 (approved):

lyzadanger:
@LMS007 Thanks for testing on LMS!

PR #47, opened 2021-05-05
Add underlying `panel` pattern and supporting structure

lyzadanger:
This PR adds a `panel` pattern (organism) and some supporting structure for it. This pattern will be used by upcoming shared `Panel` and `Dialog` components. This work extracts the visual styling implementation from the UI component implementation.

Most of the diff here is in the final commit, which updates the pattern-library pages for `panel` other molecule/organism patterns and shouldn't require close scrutiny.

## Additions

* `panel` organism mixin and utility styles. See the `Organisms` page in the local pattern library to review its purpose and constituent parts
* `actions` molecule mixin — we commonly lay out a series of buttons in a single row, generally aligned right. And we often call the container element `*__actions`. This extracts the pattern. See the `Molecules` page in the pattern library to review
* Copied over a handful of layout-related mixins and util classes from the `client` projects that were dependencies for the above

## Changes

* The `card` pattern is now a molecule instead of an organism. Expect to see things move around as we build this design system. It felt more like a molecule at second glance.
* The named spacing variables have been removed in favor of a single base-unit value: `var.$layout-space` that can be scaled (using standard multiplier math) in rules as needed. Expect that this may likely change in the future as we see how things feel. The named variants were problematic as the scale wasn't immediately evident based on the naming.
* Increased specificity on a few pattern-library CSS rules to avoid interference with patterns themselves. There is more to do here.

## To Review

Check out this branch, run `make dev` and visit the [Pattern Library](http://localhost:4001/ui-playground) _Molecules_ and _Organisms_ pages.

## What's next?

Implementation of `Panel` or `Dialog` component, with a possible interstitial step if an underlying additional styling pattern is needed for `Dialog`.

### DIFF
--- a/src/pattern-library/components/patterns/SharedOrganismPatterns.js
+++ b/src/pattern-library/components/patterns/SharedOrganismPatterns.js
@@ -5,19 +5,79 @@ import {
   Pattern,
 } from '../PatternPage';
 
+import { IconButton, LabeledButton } from '../../../';
+
 export default function SharedOrganismPatterns() {
   return (
     <PatternPage title="Organisms">
-      <Pattern title="Card">
+      <Pattern title="Panel">
+        <p>
+          A panel is a card with a header and affordances for a close button.
+          Use the <code>--closeable</code> modifier when using{' '}
+          <code>panel</code> with an icon-only close button.
+        </p>
         <PatternExamples>
-          <PatternExample details="A card is a framed element with internal margins and padding, and a hover shadow effect">
-            <div className="card">This is in a card.</div>
+          <PatternExample
+            details="Panel with no header"
+            style={{ maxWidth: '400px' }}
+          >
+            <div className="panel">
+              This is in a panel that has no header. A header is not required,
+              but you are encouraged to use <code>card</code> in that case. Note
+              that a <code>panel</code> will currently fill all available space.
+              The containing element has been set to{' '}
+              <code>max-width: 400px</code> here.
+            </div>
           </PatternExample>
-          <PatternExample details="A card applies vertical rhythm to its immediate-child elements: borders added here to highlight this">
-            <div className="card">
-              <div className="u-border">Child content in a card.</div>
-              <div className="u-border">Child content in a card.</div>
-              <div className="u-border">Child content in a card.</div>
+          <PatternExample details="Panel with title but no close button">
+            <div className="panel">
+              <header>
+                <h2 className="panel__title">
+                  This is a panel title in a panel header
+                </h2>
+              </header>
+              <div>
+                This is panel content. There is a header but no close
+                affordance.
+              </div>
+            </div>
+          </PatternExample>
+
+          <PatternExample details="Closeable panel (using IconButton): preferred">
+            <div className="panel--closeable">
+              <header>
+                <h2 className="panel__title">
+                  Panel title on a closeable panel
+                </h2>
+                <div className="panel__close">
+                  <IconButton icon="cancel" title="Close" />
+                </div>
+              </header>
+              <div>
+                This is panel content in a panel, using <code>--closeable</code>
+                .
+              </div>
+            </div>
+          </PatternExample>
+
+          <PatternExample details="Panel with actions">
+            <div className="panel--closeable">
+              <header>
+                <h2 className="panel__title">Panel title</h2>
+                <div className="panel__close">
+                  <IconButton icon="cancel" title="Close" />
+                </div>
+              </header>
+              <div>
+                This is panel content in a panel that also has some available
+                actions.
+              </div>
+              <div className="actions">
+                <LabeledButton title="Cancel">Cancel</LabeledButton>
+                <LabeledButton title="Try again" variant="primary">
+                  Try again
+                </LabeledButton>
+              </div>
             </div>
           </PatternExample>
         </PatternExamples>
--- a/styles/mixins/_layout.scss
+++ b/styles/mixins/_layout.scss
@@ -1,6 +1,50 @@
 @use '../variables' as var;
 
-$-layout-space: var.$layout-space;
+$-space: var.$layout-space;
+
+/**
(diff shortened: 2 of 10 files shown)


styles/util/_organisms.scss
```diff
@@ -4,6 +4,24 @@
...
+  * <div class="panel">
+  *  <header><h2 class="panel__title">Panel title</h2></header>
+  *  <div>This panel has no close button</div>
+  * </div>
+  *
+  * 2. Panel with title and close button
+  * <div className="panel--closeable">
+  *   <header>
+  *     <h2 class="panel__title">Panel title</h2>
+  *     <IconButton icon="cancel" title="Close" />
```
lyzadanger, line 19: This is slightly inaccurate. The markup should be:

```html
<header>
  <h2 class...></h2>
  <div class="panel__close">
    <IconButton... />
  </div>
</header>
```

I'll push a fixup commit

styles/mixins/_layout.scss
```diff
@@ -1,6 +1,50 @@
 @use '../variables' as var;
 
-$-layout-space: var.$layout-space;
+$-space: var.$layout-space;
+
+/**
```
lyzadanger, line 5: These mixins were copied from the `client`

styles/mixins/_molecules.scss
```diff
@@ -30,3 +32,25 @@ $-color-background: var.$color-background;
     }
   }
 }
+
+/**
```
lyzadanger, line 36: This was moved from organisms to here.

styles/mixins/_molecules.scss
```diff
@@ -30,3 +32,25 @@ $-color-background: var.$color-background;
...
+  @include frame($with-hover: true);
+  @include layout.vertical-rhythm;
+
+  width: 100%;
+  padding: $-space;
+
+  // TODO: Add clean-theme affordances
+}
+
+/**
```
lyzadanger, line 50: This one is new.

styles/util/_layout.scss
```diff
@@ -1,7 +1,23 @@
 @use '../mixins/layout';
 
-// Apply vertical rhythm (margins between immediate-child elements) at the
-// default rhythm-unit size
+// Apply vertical rhythm (top/bottom margins between immediate-child elements)
```
lyzadanger, line 3: Copied over from `client`

styles/mixins/_organisms.scss
```diff
@@ -1,19 +1,53 @@
...
-  @include layout.vertical-rhythm;
+@mixin panel {
+  @include molecules.card;
 
-  padding: $-layout-space;
+  & > header,
+  &__header {
+    @include layout.row($align: center);
+    @include atoms.border(bottom);
+    @include layout.horizontal-rhythm($-space);
```
esanzgar, line 22: Would an horizontal rhythm be always be suitable if there are several elements in the header?

styles/util/_layout.scss
```diff
@@ -1,7 +1,23 @@
...
+// at the default rhythm-unit size
 .u-vertical-rhythm {
   @include layout.vertical-rhythm;
 }
+
+// Apply horizontal rhythm (left/right margins between immediate-child elements)
+// at the default horizontal-rhythm size
+.u-horizontal-rhythm {
```
esanzgar, line 11: Question, does this utility class allow the parameter in the `horizontal-rhythm`?  Like this `u-horizontal-rythm(1em)`, or it takes the default parameter.

styles/variables/_layout.scss
```diff
@@ -1,8 +1,2 @@
 // Base rhythm units for layout and spacing (margins, padding)
 $space: 1rem;
-$space--xxs: $space * (1/4);
-$space--xs: $space * (1/2);
-$space--s: $space * (3/4);
-$space--m: $space;
-$space--l: $space * 2;
-$space--xl: $space * 4;
```
esanzgar, line 8: You didn't think they were useful? Or did you think it was going to be a lot of importing?
Expressions like this ones `$-space * 0.75` feel a little bit arbitrary. On the other hand `$-space--s` feels more codified.

esanzgar (approved):
Hard work! Thanks!

esanzgar:
For another PR, I suggest to add two parent levels on this menu, `Components` and `Styles`:

image

src/pattern-library/components/patterns/SharedOrganismPatterns.js
```diff
@@ -5,19 +5,116 @@ import {
...
+                </h2>
+              </header>
+              <div>
+                This is panel content. There is a header but no close
+                affordance.
+              </div>
+            </div>
+          </PatternExample>
+
+          <PatternExample details="Panel with title but no close button: className version">
```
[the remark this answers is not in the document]
lyzadanger: Removing this example.

styles/util/_layout.scss
lyzadanger: You're getting ahead of me! I was going to introduce additional utility classes as needed with modifiers indicating size. In the `client`, these look like, e.g., `.u-horizontal-rhythm--small`. We've been _sort of_ moving away from named unit sizes in `frontend-shared`, so I'm not sure how this will shake out precisely. But there will be one _class_ per "size".

Classes don't take parameters, but can wrap uses of mixins that provide different parameters. See the `.u-border--*` classes for examples.

I'll cross this bridge when we get closer to it...

lyzadanger:
@esanzgar Your feedback leaves me thinking there's a chunk of work here I'd like to encapsulate (i.e. follow-up PR): cementing how we want to handle spacing units and normalizing the layout mixins (especially the flex-related ones you've called out with their inconsistent parameters). There's a bit of nuance here because the `*-rhythm` and `row` and `column` mixins were copied over from the client, where they're already in use with the current parameters. I need to consider how to move forward and whether I need to concern myself with backwards-compatibility. So I'm going to hold on updating those parameters and their documentation in this PR, but try to get it done soon!

PR #48, opened 2021-05-05
Add HTMLAttributes to button props

LMS007:
The button should be able to set the `type` prop . Otherwise it defaults to "submit" and will submit any form that embeds it. `type` is a native prop on button and so it makes sense to just inherit the actual types from button from Preact. In doing so, we can remove some redundant props and replace them with HTMLInputElement props.

`size` is the one prop that conflicts because of the way Preact constructs HTMLInputElement--which is derived from HTMLElement and inherits the `size` prop even though it's not relevant for button. The simple fix is to omit `size` when pulling in HTMLInputElement.

This change will require adding type to each button so thats something we need to consider. One alternative could be to override type as optional and pass a default.

### DIFF
--- a/src/components/buttons.js
+++ b/src/components/buttons.js
@@ -5,22 +5,25 @@ import { SvgIcon } from './SvgIcon';
 /**
  * @typedef ButtonProps
  * @prop {import('preact').Ref<HTMLButtonElement>} [buttonRef]
- * @prop {import("preact").ComponentChildren} [children]
- * @prop {string} [className]
  * @prop {string} [icon] - Name of `SvgIcon` to render in the button
  * @prop {'left'|'right'} [iconPosition] - Icon positioned to left or to
  *   right of button text
- * @prop {boolean} [disabled]
  * @prop {boolean} [expanded] - Is the element associated with this button
  *   expanded (set `aria-expanded`)
  * @prop {boolean} [pressed] - Is this button currently "active?" (set
  *   `aria-pressed`)
- * @prop {(event: Event) => void} [onClick]
  * @prop {'small'|'medium'|'large'} [size='medium'] - Relative button size:
  *   affects padding
- * @prop {Object} [style] - Optional inline styles
  * @prop {string} [title] - Button title; used for `aria-label` attribute
  * @prop {'normal'|'primary'|'light'|'dark'} [variant='normal'] - For styling: element variant
+ */
+
+/**
+ * Fold in HTML button prop definitions into ButtonProps, but ignore `size` because it's inherited
+ * from HTMLElement and conflicts with the _ButtonProps.size prop above.
+ *
+ * @typedef {Omit<import('Preact').JSX.HTMLAttributes<HTMLButtonElement>, 'size'> } HTMLButtonElementProps
+ * @typedef {ButtonProps & HTMLButtonElementProps} ButtonBaseProps
  */
 
 /**
@@ -31,11 +34,11 @@ import { SvgIcon } from './SvgIcon';
  */
 
 /**
- * @typedef {ButtonProps & IconButtonBaseProps} IconButtonProps
+ * @typedef {ButtonBaseProps & IconButtonBaseProps} IconButtonProps
  */
 
 /**
- * @param {ButtonProps} props
+ * @param {ButtonBaseProps} props
  */
 function ButtonBase({
   buttonRef,
@@ -85,7 +88,7 @@ export function IconButton({ className = 'IconButton', ...restProps }) {
 /**
  * A labeled button, with or without an icon
  *
- * @param {ButtonProps} props
+ * @param {ButtonBaseProps} props
  */
 export function LabeledButton({
   children,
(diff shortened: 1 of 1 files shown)


src/components/buttons.js
```diff
@@ -54,6 +57,7 @@ function ButtonBase({
 
   return (
     <button
+      // @ts-ignore
```
LMS007, line 60: This error didn't make sense to me. It looks like after adding in the button props, ts thinks this is a RefObject<HTMLInputElement>|RefObject<HTMLButtonElement>  rather than RefObject<HTMLButtonElement> and so ts complains that the type is not the same.


```
^[[A^[[Asrc/components/buttons.js:60:7 - error TS2322: Type 'RefObject<HTMLButtonElement> | RefCallback<HTMLButtonElement> | RefObject<HTMLInputElement> | RefCallback<...> | undefined' is not assignable to type 'RefObject<HTMLButtonElement> | RefCallback<HTMLButtonElement> | undefined'.
  Type 'RefObject<HTMLInputElement>' is not assignable to type 'RefObject<HTMLButtonElement> | RefCallback<HTMLButtonElement> | undefined'.
    Type 'RefObject<HTMLInputElement>' is not assignable to type 'RefObject<HTMLButtonElement>'.
      Type 'HTMLInputElement' is not assignable to type 'HTMLButtonElement'.
        Types of property 'labels' are incompatible.
          Type 'NodeListOf<HTMLLabelElement> | null' is not assignable to type 'NodeListOf<HTMLLabelElement>'.
            Type 'null' is not assignable to type 'NodeListOf<HTMLLabelElement>'.

60       ref={buttonRef}
```

src/components/buttons.js
```diff
@@ -3,30 +3,33 @@ import classnames from 'classnames';
...
+ *
+ * @typedef {Omit<import('Preact').JSX.HTMLAttributes<HTMLInputElement>, 'size'> } HTMLInputElementProps
+ * @typedef {_ButtonProps & HTMLInputElementProps} ButtonProps
+ */
+
 /**
  * @typedef IconButtonBaseProps
  * @prop {string} icon - Icon is required for icon buttons
- * @prop {string} title - Title is required for icon buttons
+ * @prop {string} title - Button title; used for `aria-label` attribute
```
LMS007, line 32: I moved the description from the base to here since this is where it applies to now and we don't need to define this in the parent since its imported from HTMLButtonElement.  I left this prop here because its 1. required (unlike the its base prop) and 2. has special a function that's important to explain.

esanzgar (changes requested):
The `type` prop should be removed. Otherwise all the family of buttons must have the property:

`<IconButton **type="button"** icon="trash" title="Delete annotation"/>`

src/components/buttons.js
```diff
@@ -3,30 +3,34 @@ import classnames from 'classnames';
...
  *   right of button text
- * @prop {boolean} [disabled]
  * @prop {boolean} [expanded] - Is the element associated with this button
  *   expanded (set `aria-expanded`)
  * @prop {boolean} [pressed] - Is this button currently "active?" (set
  *   `aria-pressed`)
- * @prop {(event: Event) => void} [onClick]
  * @prop {'small'|'medium'|'large'} [size='medium'] - Relative button size:
  *   affects padding
  * @prop {Object} [style] - Optional inline styles
```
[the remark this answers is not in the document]
LMS007: Okay, I was only worried about the semantic implications of every button defaulting to type submit. DOing a quick test with voice over, it does not seem to care what the type is. Or rather, nothing changes in what it says, so maybe this is safe.

src/components/buttons.js
```diff
@@ -3,30 +3,34 @@ import classnames from 'classnames';
...
- * @prop {string} [title] - Button title; used for `aria-label` attribute
  * @prop {'normal'|'primary'|'light'|'dark'} [variant='normal'] - For styling: element variant
+ * @prop {string} type
```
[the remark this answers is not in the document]
LMS007: Okay, I was only worried about the semantic implications of every button defaulting to type submit. DOing a quick test with voice over, it does not seem to care what the type is. Or rather, nothing changes in what it says, so maybe this is safe.

src/components/buttons.js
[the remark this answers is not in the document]
LMS007: good catch!

src/components/buttons.js
```diff
@@ -5,37 +5,39 @@ import { SvgIcon } from './SvgIcon';
...
- * @prop {boolean} [disabled]
  * @prop {boolean} [expanded] - Is the element associated with this button
  *   expanded (set `aria-expanded`)
  * @prop {boolean} [pressed] - Is this button currently "active?" (set
  *   `aria-pressed`)
- * @prop {(event: Event) => void} [onClick]
  * @prop {'small'|'medium'|'large'} [size='medium'] - Relative button size:
  *   affects padding
- * @prop {Object} [style] - Optional inline styles
- * @prop {string} [title] - Button title; used for `aria-label` attribute
```
esanzgar, line 22: I would recommend keeping this line. It points to a different use between our custom button and the default ones.

esanzgar (approved):
I just leave a comment.

src/components/buttons.js
LMS007: will do, makes sense

PR #51, opened 2021-05-10
Avoid using alternative props for the same purpose

esanzgar:
I disallowed `aria-expanded`, `aria-pressed` and `aria-label` because
there are props to set these values (`expanded`, `pressed` and `title`,
respectively)


lyzadanger (approved):
Thank you; this is a nice little improvement.

PR #56, opened 2021-05-10
Split up pattern library routes: patterns, components

lyzadanger:
This PR divides up pattern library common routes to separate them into Patterns and Components, as per reviewer suggestion:

image

It also adjusts a bit of CSS styling those nav lists.


esanzgar (approved):
I left a couple of comments.

PR #57, opened 2021-05-10
Reduce selector insanity on `panel` pattern

lyzadanger:
I realized today that the new implementation of the `panel` pattern in SASS caused a whole glut of selectors to be generated for certain sub-elements in `panel`s. To wit, before:

Screen Shot 2021-05-10 at 4 06 06 PM

and after:

Screen Shot 2021-05-10 at 4 08 04 PM

Usage change (this isn't used anywhere else, so the change won't affect anything):

Before:

`<div className="panel--closeable"> ... </div>`

After:

`<div className="panel panel--closeable"> ...</div>`

Examples are updated accordingly.


esanzgar (approved):
Good that you notice this before we start using the pattern.

It is kind of crazy the amount of selectors generated. Something to watch in the future.

PR #61, opened 2021-05-13
Prettify one of the example of the LabeledCheckbox

esanzgar:
`jsxToString` output of the previous function wasn't super pretty. This change
displays the following instead, which is nicer:
```
<LabeledCheckbox name="test-alternative" position="before" onToggle={isChecked => setWantWatermelon(isChecked)}>
  <code>
    I want a watermelon
  </code>
</LabeledCheckbox>
```


lyzadanger (approved):
👍🏻

PR #62, opened 2021-05-13
Set default `type` property of buttons to "button"

robertknight:
Override the default "submit" value for the `type` property of buttons
to "button". This avoids a hazard where a button can unexpectedly submit
a containing form that it is not intended to.

This choice mirrors button components in several other major React UI
libraries (Material UI, Fluent UI, React Bootstrap), as well as the
`Button` component in the Hypothesis lms app.


lyzadanger (approved):
Thanks! LGTM.

PR #64, opened 2021-05-13
Correct `git push` instructions

robertknight:
The repository comes before the ref specs. Also the tag and the branch can be pushed in a single command.


robertknight:
I made these edits on GitHub originally. The result needed to be fixed up by running `yarn format` locally.

lyzadanger (approved):
Thanks for catching this!

PR #66, opened 2021-05-13
Prefix all utility and pattern classes with `.hyp-` and update docs

lyzadanger:
Prefixing utility and pattern classes can help guard against unexpected
conflict with pre-existing classnames in applications that use this
package.

The classnames that are prefixed in these changes are currently used only internally to the package and not part of its CSS API—yet. Once these classes are part of the main SASS styles, these prefixes will aid the project of replacing local application utility classes with these.

Slack thread with some related discussion: https://hypothes-is.slack.com/archives/C1M8NH76X/p1620923544133600 (private to Hypothesis team members).


LMS007 (approved):
👍

PR #67, opened 2021-05-18
Remove alignment rule from buttons; add columnar `actions` pattern

### LINKED ISSUE #63: Allow controlling alignment of text labels in `LabeledButton`
The UI mocks in https://github.com/hypothesis/lms/issues/2523#issuecomment-821160185 call for a vertical list of buttons with the left edges of the button labels vertically aligned. This is not currently supported by `LabeledButton`, which always centers the labels. Example usage from that issue:

lyzadanger:
This PR removes an unnecessary `justify-content` alignment rule from buttons styling to allow for left alignment of button text in cases where buttons are "stretched" wider than their content-sized width.

The removal of the rule has no effect on existing button patterns, as it didn't "do" anything for buttons that are content-sized.

This PR also adds a variant of the `actions` pattern, `actions--column`, to capture this usage of buttons stacked on top of each other. It can be viewed on the "Molecules" pattern-library page.

Fixes #63 

Screenshots from pattern library:

Existing labeled buttons unaffected:

New columnar `actions` pattern showing stretched buttons:

image

### DIFF
--- a/src/pattern-library/components/patterns/SharedMoleculePatterns.js
+++ b/src/pattern-library/components/patterns/SharedMoleculePatterns.js
@@ -57,7 +57,8 @@ export default function SharedMoleculePatterns() {
       <Pattern title="Actions">
         <p>
           The <code>actions</code> pattern presents a collection of actions
-          (typically buttons) spaced out in a row, aligned right.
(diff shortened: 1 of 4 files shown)


robertknight (approved):
LGTM

PR #69, opened 2021-05-18
Add clarification on availability of mixins

lyzadanger:


styles/README.md
```diff
@@ -18,7 +18,8 @@ The public SASS interface for the `frontend-shared` package is:
 
   `@use '@hypothesis/frontend-shared/styles/mixins/<mixin-module>';`
 
-  Individual reusable mixins, e.g. `focus`
+  - At this time, the only publicly-available mixins are `focus` and `buttons` (for button customizations only—use only if you know what you're doing). Do not use other mixins.
```
robertknight, line 21:
```suggestion
  - At this time, the only publicly-available mixin modules are `focus` and `buttons` (for button customizations only—use only if you know what you're doing). Do not use other mixins.
```

robertknight (approved):
LGTM. In future would be preferable to encode the private vs public separation in code or filename structure somehow, whether by using SASS's built-in facilities or some simple convention. I'll let you ruminate on how best to do this.

lyzadanger:
>  In future would be preferable to encode the private vs public separation in code or filename structure somehow

Definitely. I just wish I had more time to actually _work on this_ :)

PR #70, opened 2021-05-18
Establish basic color patterns

lyzadanger:
This PR establishes some very basic structure around managing colors within this package's internal SASS. This is part of ongoing work to start to establish some foundational bits and bobs of our design system—not creating a new system, but describing the existing one.

Changes relevant to package users: additional page visible in the pattern library (Colors) with some basic color swatches, and a  sub-heading changed from _Common Patterns_ to _Foundations_. No externally-consumable SASS or CSS is affected by these changes.

These changes are internal to the package:

* Establishes a 0 - 9 greyscale from existing color values. Internally some variable names changed, but no colors changed—just their references. Previously we had a 0-7 greyscale range, with some intermediate `grey-mid` and `grey-semi` values. One had to remember (or forget, every time) whether `grey-semi` was lighter or darker than `grey-4`. No longer. Values are lightest at 0 and darkest at 9.
* Standardizes naming for color variables
* Generates utility classes for foreground and background colors

There's a lot more to do here, but I wanted to keep things as simple as possible. These changes were motivated by some needs for `Dialog`—it felt like the right time to do this.

Screenshot from new pattern-library page:

image

There's a lot to do here (I left myself some TODOs, even), but it's a start.

PR #71, opened 2021-05-19
Refine pattern-library navigation

lyzadanger:
~~Currently depends on #70 for convenience reasons and to save me some rebase/merge time, but it doesn't need to if that becomes cumbersome.~~

Per dev feedback and obvious TODOs:

- Make entire nav item clickable
- Add navigation "on" state to show which page you're on
- DRY out route iteration for navigation rendering

After, showing navigation "on" state:

image

TODO later:

Routes could be better organized in general.

### DIFF
--- a/src/pattern-library/components/PlaygroundApp.js
+++ b/src/pattern-library/components/PlaygroundApp.js
@@ -1,3 +1,5 @@
+import classnames from 'classnames';
+
 import { SvgIcon } from '../../components/SvgIcon';
 
 // Design patterns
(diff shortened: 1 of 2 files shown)


styles/pattern-library.scss
```diff
@@ -17,6 +17,7 @@ body {
 
 .PlaygroundApp__sidebar h2,
 .Pattern > h2 {
+  margin: 0.5em 0;
```
lyzadanger, line 20: Just a tuning improvement in layout spacing.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -97,16 +99,25 @@ export default function PlaygroundApp({
   extraRoutes = [],
 }) {
   const allRoutes = routes.concat(extraRoutes);
-  const [route, navigate] = useRoute(baseURL, allRoutes);
-  const content = route ? (
-    <route.component />
+  const [activeRoute, navigate] = useRoute(baseURL, allRoutes);
```
lyzadanger, line 102: Renamed `route` to `activeRoute` here to avoid confusion as `route` is used for other references.

robertknight (approved):
LGTM 👍

Issue #53, opened 2021-05-10
Implement common Dialog component


lyzadanger:
A common `Dialog` preact component (and relevant design pattern(s)) is needed to address inconsistencies in patterns between applications. 

Part of #83

lyzadanger:
Just a note to outline some of my thinking here. I'm considering implementing the following components:

* `Dialog`: A special kind of `Panel` that is aware of routing focus, containing buttons and cancel/dismissal. This kind of `Dialog` could show up anywhere in the application UI, and doesn't overlay/obscure UI. It is plausible that some of the components in the `client` that are currently `Panel`s would be better served as being `Dialog`s when they need user focus and input.
* `Modal`: A special kind of `Dialog` that comes with a fullscreen overlay and positioning
* `ConfirmModal`: Akin to `ConfirmDialog` in the client: wraps a `Modal` with the specific "boolean-ness" of a confirm prompt.

lyzadanger:
`Dialog` is done and delivered. The next steps are `Modal` and `ConfirmModal`, which are in progress and I will track in a separate issue.

PR #73, opened 2021-05-19
Add `overlay` and `fixed-centered` layout patterns

lyzadanger:
Part of https://github.com/hypothesis/frontend-shared/issues/53

This is an unexciting PR that adds a couple of layout patterns for:

* Vertically and horizontally centering an element in the viewport
* A full-viewport semi-transparent overlay

A new Layout page is added to the Foundations section of the Pattern Library.

Acknowledgement: The Layout page is fairly fugly, and we're getting to the point where I want to go back and make some improvements, but I also didn't want to distract from moving toward a `Dialog` at this moment.

## To see it in action

Check out this branch, run `make dev`, go to the [Pattern Library's new Layout page](http://localhost:4001/ui-playground/shared-layout).

### DIFF
--- a/src/pattern-library/components/patterns/LayoutPatterns.js
+++ b/src/pattern-library/components/patterns/LayoutPatterns.js
@@ -1,0 +1,132 @@
+import { useState } from 'preact/hooks';
+
+import { LabeledButton } from '../../../';
+
+import {
+  PatternPage,
+  Pattern,
+  PatternExamples,
+  PatternExample,
+} from '../PatternPage';
+
+export default function LayoutPatterns() {
+  const [showExample1, setShowExample1] = useState(false);
+  const [showExample2, setShowExample2] = useState(false);
+  const [showExample3, setShowExample3] = useState(false);
+  return (
+    <PatternPage title="Layout">
+      <Pattern title="Fixed-Centered Positioning">
+        <p>
+          The <code>fixed-centered</code> layout pattern centers an element both
+          horizontally and vertically within the entire viewport.
+        </p>
+        <PatternExamples>
+          <PatternExample details="Centering an element vertically and horizontally within the viewport">
+            <div>
+              <LabeledButton
+                variant="primary"
+                onClick={() => setShowExample1(true)}
+              >
+                Show example
+              </LabeledButton>
+              <div
+                className="hyp-u-fixed-centered"
+                style={`width:450px;${
+                  showExample1 ? 'visibility:visible' : 'visibility:hidden'
+                }`}
+              >
+                <div className="hyp-card">
+                  <p>
+                    This is a card that is centered vertically and horizontally
+                    in the current viewport.
+                  </p>
+                  <div className="hyp-actions">
+                    <LabeledButton
+                      variant="primary"
+                      onClick={() => setShowExample1(false)}
+                    >
+                      Hide example
+                    </LabeledButton>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </PatternExample>
+        </PatternExamples>
+      </Pattern>
+      <Pattern title="Full-screen overlay">
+        <p>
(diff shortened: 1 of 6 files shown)


lyzadanger:
@esanzgar Gentle ping for you when you're back. These patterns will be used for the `Modal`.

src/pattern-library/components/patterns/LayoutPatterns.js
```diff
@@ -0,0 +1,132 @@
...
+              >
+                Show example
+              </LabeledButton>
+              <div
+                className="hyp-u-fixed-centered"
+                style={`width:450px;${
+                  showExample1 ? 'visibility:visible' : 'visibility:hidden'
+                }`}
+              >
+                <div className="hyp-card">
```
esanzgar, line 38: Maybe these two divs could be combined into one:

```suggestion
              <div
                className="hyp-u-fixed-centered hyp-card"
                style={`width:450px; ${
                  showExample1 ? 'visibility:visible' : 'visibility:hidden'
                }`}
              >
```

esanzgar: Maybe also for the ones below: `className="hyp-u-overlay hyp-u-fixed-centered"` or `className=hyp-u-overlay hyp-u-fixed-centered hyp-card"`

styles/mixins/_layout.scss
```diff
@@ -82,3 +83,27 @@ $-space: var.$layout-space;
...
+/**
+ * Semi-opaque overlay, full-viewport
+ */
+@mixin overlay {
+  z-index: 10;
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
```
esanzgar, line 107: I have seen also implemented this way:
```suggestion
  bottom: 0;
  right: 0;
```

styles/mixins/_atoms.scss
```diff
@@ -29,9 +29,9 @@ $-color-shadow: var.$color-shadow;
 @mixin shadow($active: false) {
   @if $active {
     /* offset-x | offset-y | blur-radius | spread-radius | color */
-    box-shadow: 0px 2px 3px 0px color.scale($-color-shadow, $alpha: -85%);
+    box-shadow: 0px 2px 3px 0px $-color-shadow;
   } @else {
     /* offset-x | offset-y | blur-radius | color */
-    box-shadow: 0px 1px 1px color.scale($-color-shadow, $alpha: -90%);
+    box-shadow: 0px 1px 1px $-color-shadow--active;
```
esanzgar, line 35: This looks the opposite with the else block (it feels it should this be `--no-active`).

Maybe sticking with `--dark` is better.

esanzgar (approved):
A couple of minor comments.

styles/mixins/_atoms.scss
lyzadanger: You're right, this is just...wrong. Will fix.

styles/mixins/_layout.scss
lyzadanger: Sure, this seems fine! I'll update.

src/pattern-library/components/patterns/LayoutPatterns.js
lyzadanger: I'd like to keep the two separate div elements because I want to prevent any case in which styles applied by (in this case) `hyp-card` are interfering with `hyp-u-fixed-centered`. `hyp-u-fixed-centered` is intended to be a container, while the `div.hyp-card` is the "thing inside the container."

This will mirror how this pattern will be used in the `Modal` component, which I expect will be something like:

```jsx
<>
  <div className="Modal__overlay" />
  <div className="Modal__container" />
  <div className="Dialog" />
</>
```

Where `Modal__overlay` inherits from `overlay` and `Modal__container` from `fixed-centered` and `Dialog` is inside the `fixed-centered` "container."

PR #75, opened 2021-05-21
Add Dialog Component

lyzadanger:
This PR adds a `Dialog` component and makes some supporting adjustments to patterns and the pattern library:

* Add a `Dialog` component, styles and tests
* Adjust some foundational shared patterns that are used by this component to, well, be a little better—see commit messages, but affected patterns include `card` and `frame`. No visible changes to consumers of the package.
* Add a page to the pattern library to demo `Dialog` components, including some provisional styling for a potential future pattern-library component

A `Dialog` is a panel-like component to use when user interaction is desired. It will route focus when opened, and has `aria-` attributes for a11y.

The `Dialog` component here is very similar to the `Dialog` component in the `client` project, with these differences:

* This version adds an (optional) `icon` prop
* This version uses the `panel` design pattern which increases visual consistency, instead of rolling its own styling
* This `Dialog` does not assume it's modal, and does not provide a backdrop or positioning. That will be supported in a forthcoming `Modal` component.

## To preview

Check out this branch, visit the pattern library, go to the ["Dialogs" page](http://localhost:4001/ui-playground/shared-dialog):

image

### DIFF
--- a/src/components/Dialog.js
+++ b/src/components/Dialog.js
@@ -1,0 +1,139 @@
+import classnames from 'classnames';
+import { useEffect, useLayoutEffect, useRef, useState } from 'preact/hooks';
+
+import { IconButton, LabeledButton } from '../';
+import { SvgIcon } from './SvgIcon';
+
+let idCounter = 0;
+
+/**
+ * Return an element ID beginning with `prefix` that is unique per component instance.
+ *
+ * This avoids different instances of a component re-using the same ID.
+ *
+ * @param {string} prefix
+ */
+function useUniqueId(prefix) {
+  const [id] = useState(() => {
+    ++idCounter;
+    return `${prefix}-${idCounter}`;
+  });
+  return id;
+}
+
+/**
+ * @typedef {import('preact').ComponentChildren} Children
+ *
+ * @typedef DialogProps
+ * @prop {Children} [buttons] -
+ *   Additional `Button` elements to display at the bottom of the dialog.
+ *   A "Cancel" button is added automatically if the `onCancel` prop is set.
+ * @prop {string} [cancelLabel] - Label for the cancel button
+ * @prop {Children} children
+ * @prop {string} [contentClass] - CSS class to apply to the dialog's content
+ * @prop {string} [icon] - Name of optional icon to render in header
+ * @prop {import("preact/hooks").Ref<HTMLElement>} [initialFocus] -
+ *   Child element to focus when the dialog is rendered.
+ * @prop {() => void} [onCancel] -
+ *   A callback to invoke when the user cancels the dialog. If provided, a
+ *   "Cancel" button will be displayed.
+ * @prop {'dialog'|'alertdialog'} [role] - The aria role for the dialog (defaults to" dialog")
+ * @prop {string} title
+ */
+
+/**
+ * HTML control that can be disabled.
+ *
+ * @typedef {HTMLElement & { disabled: boolean }} InputElement
+ */
+
+/**
+ * Render a "panel"-like interface with a title and optional icon and/or
+ * close button. Grabs focus on initial render, defaulting to the entire
+ * Dialog container element, or `initialFocus` HTMLElement if provided.
+ *
+ * @param {DialogProps} props
+ */
+export function Dialog({
+  buttons,
+  cancelLabel = 'Cancel',
+  children,
+  contentClass,
+  icon,
+  initialFocus,
+  onCancel,
+  role = 'dialog',
+  title,
+}) {
+  const dialogDescriptionId = useUniqueId('dialog-description');
+  const dialogTitleId = useUniqueId('dialog-title');
+
+  const rootEl = useRef(/** @type {HTMLDivElement | null} */ (null));
+
+  useEffect(() => {
+    const focusEl = /** @type {InputElement|undefined} */ (initialFocus?.current);
+    if (focusEl && !focusEl.disabled) {
+      focusEl.focus();
+    } else {
+      // Modern accessibility guidance is to focus the dialog itself rather than
+      // trying to be smart about focusing a particular control within the
+      // dialog. See resources above.
+      rootEl.current.focus();
+    }
+    // We only want to run this effect once when the dialog is mounted.
+    //
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  // Try to assign the dialog an accessible description, using the content of
+  // the first paragraph of text in it.
+  //
+  // A limitation of this approach is that it doesn't update if the dialog's
+  // content changes after the initial render.
+  useLayoutEffect(() => {
+    const description = rootEl.current.querySelector('p');
+    if (description) {
+      description.id = dialogDescriptionId;
+      rootEl.current.setAttribute('aria-describedby', dialogDescriptionId);
+    }
+  }, [dialogDescriptionId]);
+
+  return (
+    <div
+      aria-labelledby={dialogTitleId}
+      className={classnames(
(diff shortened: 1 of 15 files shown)


lyzadanger:
There's a minor test miss here that I will fix.

src/pattern-library/components/patterns/SharedMoleculePatterns.js
```diff
@@ -12,30 +12,35 @@ export default function SharedMoleculePatterns() {
     <PatternPage title="Molecules">
       <Pattern title="Frame">
         <p>
-          A <code>frame</code> has a border and a background, but no other
-          layout affordances.
+          A <code>frame</code> has a border, background, padding, and vertical
```
esanzgar, line 15: A frame is looking more to a card

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @typedef {import("preact").ComponentChildren} Children
+ *
+ * @typedef DialogProps
+ * @prop {Children} [buttons] -
+ *   Additional `Button` elements to display at the bottom of the dialog.
+ *   A "Cancel" button is added automatically if the `onCancel` prop is set.
+ * @prop {string} [cancelLabel] - Label for the cancel button
+ * @prop {import("preact").ComponentChildren} children
```
esanzgar, line 32: @prop {Children} children

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+ * @prop {string} [contentClass] - CSS class to apply to the dialog's content
+ * @prop {string} [icon] - Name of optional icon to render in header
+ * @prop {import("preact/hooks").Ref<HTMLElement>} [initialFocus] -
```
esanzgar, line 35: * @prop {import('preact').Ref<HTMLElement>} [initialFocus]

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+function useUniqueId(prefix) {
+  const [id] = useState(() => {
+    ++idCounter;
+    return `${prefix}-${idCounter}`;
+  });
+  return id;
+}
+
+/**
+ * @typedef {import("preact").ComponentChildren} Children
```
esanzgar, line 25: @typedef {import('preact').ComponentChildren} Children

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+  role = 'dialog',
+  title,
+}) {
+  const dialogDescriptionId = useUniqueId('dialog-description');
+  const dialogTitleId = useUniqueId('dialog-title');
+
+  const rootEl = useRef(/** @type {HTMLDivElement | null} */ (null));
+
+  useEffect(() => {
+    const focusEl = /** @type {InputElement|undefined} */ (initialFocus?.current);
```
esanzgar, line 74: I thought `initialFocus` type was HTMLElement.

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+  //
+  // A limitation of this approach is that it doesn't update if the dialog's
+  // content changes after the initial render.
+  useLayoutEffect(() => {
+    const description = rootEl.current.querySelector('p');
+    if (description) {
+      description.id = dialogDescriptionId;
+      rootEl.current.setAttribute('aria-describedby', dialogDescriptionId);
+    }
+  }, [dialogDescriptionId]);
```
esanzgar, line 99: I am not very convinced by this approach. I am concerned about collisions on `id`s. I will think over this in the weekend. I

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+    // We only want to run this effect once when the dialog is mounted.
+    //
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  // Try to assign the dialog an accessible description, using the content of
+  // the first paragraph of text in it.
+  //
+  // A limitation of this approach is that it doesn't update if the dialog's
```
esanzgar, line 91: That's a serious limitation to be aware.

esanzgar (commented):
Lyza, I provided an initial feedback to this PR.

src/pattern-library/components/patterns/SharedMoleculePatterns.js
lyzadanger: A frame gives you a border, background, padding and spacing. A card gives you dimensionality (shadow/hover shadow). These patterns are still settling out a bit. The answer to 

> If I want to create a new component do I based on a frame or a card?

is currently "neither" — these patterns are still private to the pattern library. Once they're not private anymore this question should have a clear(er) answer.

src/components/Dialog.js
lyzadanger: @esanzgar This was directly pulled over from the current implementation on the client. I can take a deeper look here...

src/components/Dialog.js
lyzadanger: I remember saying the same thing when I reviewed the original implementation of this on the `client` :) — I'm wondering if this component should take an optional prop for this?

src/components/Dialog.js
lyzadanger: Thanks for spotting that!

src/components/Dialog.js
lyzadanger: This gives me `Generic type 'Ref' requires 1 type argument(s).ts(2314)`

src/components/Dialog.js
lyzadanger: This gives me `Generic type 'Ref' requires 1 type argument(s).ts(2314)`

src/components/Dialog.js
lyzadanger: To confirm: single quotes here? Wish we could lint this!

src/components/Dialog.js
lyzadanger: I need some assistance! I inherited this code (it's from the `client`). I can see that this "cast" is intended to avoid a TS error on the line following:

`if (focusEl && !focusEl.disabled) {`

Only `InputElement`s have `disabled` props; `HTMLElement`s do not. I tried a few things but haven't been successful at finding a nice workaround. The intent here is: focus on whatever element was provided, but not if that element is disabled.  The provided ref/element might be an `InputElement`, but might not. I don't know the best way to solve this!

src/components/Dialog.js
esanzgar: I meant `import('preact').Ref<HTMLInputElement>` but it is actually a different type. So forget about this suggestion.

The only thing maybe, would be to change the double quote for single quotes.

src/components/Dialog.js
```diff
@@ -0,0 +1,139 @@
...
+}) {
+  const dialogDescriptionId = useUniqueId('dialog-description');
+  const dialogTitleId = useUniqueId('dialog-title');
+
+  const rootEl = useRef(/** @type {HTMLDivElement | null} */ (null));
+
+  useEffect(() => {
+    const focusEl = /** @type {InputElement|undefined} */ (initialFocus?.current);
+    if (focusEl && !focusEl.disabled) {
+      focusEl.focus();
```
esanzgar, line 76:
```suggestion
    const focusEl = initialFocus?.current;
    if (focusEl && !focusEl.hasAttribute('disabled')) {
      focusEl.focus();
```

esanzgar (approved):
As we mentioned, we can revisit this component on a future sprint.

lyzadanger:
Follow-up issue is here: https://github.com/hypothesis/frontend-shared/issues/77

PR #76, opened 2021-05-21
Simplify routing and pattern component names

lyzadanger:
This is a Friday-afternoon housekeeping PR that takes care of a few tasks I had in my backlog:

* Extract routes for the pattern library into a module and simplify their organization (and simplify the structure of `PlaygroundApp`)
* Shorten the module and component names for pattern components, as they were clunky

There is no user-visible change.


LMS007 (approved):
👍

Issue #74, opened 2021-05-20
Fill in some `layout`, `molecule` utility gaps and consolidate


lyzadanger:
I've identified a list of small tasks as I've been working on some things:

* [x] Migrate legacy layout mixins away from using `!important` in their default variants, and add `!`-variant classnames (and evaluate components that use those mixins/utilities to make sure they still look right)
* [x] Flesh out utility class variants for additional layout-row and -column alignment options (e.g. `.hyp-u-layout-row--right`, etc.)
* [x] Migrate the base padding from `card` to `frame` pattern—`frame` is useless without some padding

PR #78, opened 2021-05-25
Re-order (intended) CSS output and make SASS source directories reflect this

lyzadanger:
This PR reorders (intended) output of CSS to give some utility classes a fighting chance of applying, without having to resort to `!important` or other weapons. These utility classes are (still) only used within the package itself. Nothing changes within the package's API.

Motivations:

1. Find the minimally-nuclear approach to assuring that utility classes can override as needed.
2. Resolve dependency and ordering differences between "true" utility classes and mixins vs. "patterns"

With regard to the first item, the goal is that authors can do something like: `<div className="hyp-card hyp-u-bg-color--grey-2">` and feel confident that they'll get a card pattern, but with a different background color.

Previous to these changes, I'd taken a more nuclear approach, though I tried to make sure the author meant to "press the button" by creating additional, `!`-named utility classes that included an `!important` rule, e.g. `.hyp-!-u-bg-color--grey-2`. Clever, but let's try to do better.

First stop: simple CSS ordering. This PR updates the pattern-library SASS entry point (the only thing that consumes utility classes at this time) to include `utils` last.

This on its own didn't fix the specific example here: `<div className="hyp-card hyp-u-bg-color--grey-2">` because "molecule" and "organism" CSS styles (which include `hyp-card`) were getting output with utility styles, and by luck of the draw/alphabet, `hyp-card` was getting output after `hyp-u-bg-color--grey-2`.

Stepping back and looking at things: true util classes should be atomic. "Molecules" and "organisms" aren't utils, they're composite pattern styles, which in turn draw on more atomic patterns. This was already reflected in their dependencies (atomic mixins have no other mixin dependencies, but molecules and organisms depend on atomic mixins). Let's get molecules and organisms out of the atomic-utility-class mix, then, by putting them into their own `patterns` directory, which gets included in stylesheet output before `utils`.

Finally, update the README to explain WTH and remove the (now) unneeded color-utility-override classes and example in the pattern library.

There are still instances of `!important` in two layout mixins, which I'll review as part of https://github.com/hypothesis/frontend-shared/issues/74

p.s. We have a few other tricks up our sleeve that are less scorched-earth than `!important`, like leveraging the `:is` pseudo-select, should we run into further specificity wars. The arsenal is not depleted.

### DIFF
--- a/styles/README.md
+++ b/styles/README.md
@@ -57,15 +57,29 @@ Mixins in `mixins` and utility styles in `util` loosely apply [Atomic Design](ht
 
 The directory that a SASS module lives in dictates what it should provide (styles, mixins, variables, etc.) and what kind of dependencies it is allowed:
 
-| Directory    | Description                                     | Provides  | Dependencies          |
-| ------------ | ----------------------------------------------- | --------- | --------------------- |
-| `base`       | Global reset styles: element styles, typography | styles    | mixins, variables[^1] |
-| `components` | Styles for shared components                    | styles    | any                   |
-| `mixins`     | Mixins                                          | mixins    | mixins[^2], variables |
-| `util`       | Utility styles                                  | styles    | mixins                |
-| `variables`  | Variables                                       | variables | none                  |
+| Directory    | Description                  | Provides  | Dependencies          |
+| ------------ | ---------------------------- | --------- | --------------------- |
+| `base`       | Reset and element styles     | styles    | mixins, variables[^1] |
+| `components` | Styles for shared components | styles    | any                   |
+| `mixins`     | Mixins                       | mixins    | mixins[^2], variables |
+| `patterns`   | Pattern styles               | styles    | mixins                |
+| `util`       | Utility styles               | styles    | mixins                |
+| `variables`  | Variables                    | variables | none                  |
 
-Directories whose modules provide styles should contain an entry-point module (`index.scss`) for the convenience of consumers.
+## CSS Output and Ordering
+
+The current SASS entry point for the package only generates styles from `components`.
+
+Other CSS output is currently private to the package, and used (only) by the pattern library.
+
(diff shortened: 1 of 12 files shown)


styles/README.md
```diff
@@ -57,15 +57,30 @@ Mixins in `mixins` and utility styles in `util` loosely apply [Atomic Design](ht
...
-| Directory    | Description                                     | Provides  | Dependencies          |
-| ------------ | ----------------------------------------------- | --------- | --------------------- |
-| `base`       | Global reset styles: element styles, typography | styles    | mixins, variables[^1] |
-| `components` | Styles for shared components                    | styles    | any                   |
-| `mixins`     | Mixins                                          | mixins    | mixins[^2], variables |
-| `util`       | Utility styles                                  | styles    | mixins                |
-| `variables`  | Variables                                       | variables | none                  |
+| Directory    | Description                  | Provides  | Dependencies          |
+| ------------ | ---------------------------- | --------- | --------------------- |
+| `base`       | reset and element styles     | styles    | mixins, variables[^1] |
```
LMS007, line 62: All other descriptions start with a capitol

styles/README.md
```diff
@@ -57,15 +57,30 @@ Mixins in `mixins` and utility styles in `util` loosely apply [Atomic Design](ht
...
+Other CSS output is currently private to the package, and used (only) by the pattern library.
+
+The intended ordering of CSS output—to assure proper cascade/ordering—is:
+
+1. `base`
+2. `components`
+3. `patterns`
+4. `utils`
+
+Note `utils` in last position, to assure those classes can be used as override
```
LMS007, line 82: This does not read right to me, perhaps something like:

"...to assure those classes override others."

LMS007 (approved):
Test, and looks good. Just had a two small comments about text that could maybe be fixed up.

PR #79, opened 2021-05-26
Add styling for `dark` variant of `IconButton`

lyzadanger:
This PR adds colors (styling) for the `dark` variant of `IconButton`. We didn't have a use case for this variant until now, when it could be of use as a close button for the Notebook.

## To review

Check out this branch, `make dev` and go to the [Buttons page of your Pattern Library](http://localhost:4001/ui-playground/components-buttons).

Screen shot:

image

### DIFF
--- a/src/pattern-library/components/patterns/ButtonPatterns.js
+++ b/src/pattern-library/components/patterns/ButtonPatterns.js
@@ -81,6 +81,48 @@ export default function ButtonPatterns() {
           </PatternExample>
         </PatternExamples>
 
+        <h3>Dark variant</h3>
+        <p>
+          This variant is for use on darker (light grey vs. white) backgrounds.
+          Note that this button, unlike other <code>IconButton</code>s, has a
+          background color. This is to allow for a use case in which the dark{' '}
+          <code>IconButton</code> is initially fixed on a grey background but
+          floats on top of content when scrolled.
+        </p>
+
+        <PatternExamples>
+          <PatternExample
+            details="Basic usage"
+            style={{ backgroundColor: '#ececec' }}
+          >
+            <IconButton icon="cancel" title="Close" variant="dark" />
+          </PatternExample>
+          <PatternExample
+            details="Note that the button has a background color"
+            style={{ backgroundColor: '#ffffff' }}
+          >
+            <IconButton icon="cancel" title="Close" variant="dark" />
+          </PatternExample>
+          <PatternExample
+            details="Pressed"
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/patterns/ButtonPatterns.js
```diff
@@ -81,6 +81,48 @@ export default function ButtonPatterns() {
           </PatternExample>
         </PatternExamples>
 
+        <h3>Dark variant</h3>
+        <p>
+          This variant is for use on darker (light grey vs. white) backgrounds.
+          Note that this button, unlike other <code>IconButton</code>s, has a
+          background color. This is to allow for a use case in which the dark{' '}
+          <code>IconButton</code> is initially fixed on a grey background but
+          floats on top of content when scrolled.
```
esanzgar, line 90: That's nice that you made it work straight for the notebook, thanks! Do you think could be a problem in other context?

src/pattern-library/components/patterns/ButtonPatterns.js
```diff
@@ -81,6 +81,48 @@ export default function ButtonPatterns() {
...
+        </p>
+
+        <PatternExamples>
+          <PatternExample
+            details="Basic usage"
+            style={{ backgroundColor: '#ececec' }}
```
esanzgar, line 96: You could create a variable: `const backgroundColor = '#ececec';` and use to replace all the lines:
```suggestion
            style={{ backgroundColor }}
```

esanzgar (approved):
Thanks!

esanzgar:
Could this variant be useful for the `LabeledButton` or the `LinkButton`?

src/pattern-library/components/patterns/ButtonPatterns.js
lyzadanger: I appreciate you asking this! And I'll be honest to say: I don't know! There is a little bit of experimentation going on here. We might find that we want to harmonize this by adding an explicit background color for `LabeledButton`'s `dark` variant.

src/pattern-library/components/patterns/ButtonPatterns.js
lyzadanger: It's not DRY, but I'd like to leave it as-is as it's easy to understand and scan, and we may end up with examples on this page that use a different background color.

lyzadanger:
> Could this variant be useful for the LabeledButton or the LinkButton? 

They already exist! :) You can see them on the `Buttons` pattern library page.

Issue #86, opened 2021-06-02
Implement needed `Modal` components


lyzadanger:
Part of #83 

Implement `Modal` and `ConfirmModal` components in this package to replace variants in use in the `client`. 

* `Modal` builds on `Dialog` and adds a backdrop, positioning
* `ConfirmModal` builds on `Modal` and provides a boolean variant of `Modal` (yes/no options)

PR #88, opened 2021-06-03
`Modal`, `ConfirmModal` components and design patterns

lyzadanger:
This (initially draft) PR contains new `Modal` and `ConfirmModal` components and their corresponding supporting design patterns and pattern-library content.

`Modal` wraps a `Dialog` and adds a background overlay and a close-on-outside-click functionality. `Modal` + `Dialog` here = the `Dialog` component in the `client`. `ConfirmModal` is analogous to `ConfirmDialog` in the `client` and doesn't deviate too much from it.

I've added some responsiveness to the modal positioning and sizing in this variant. See implementation and comments for details (a test to see if the source is self-documenting and commented enough).

### In draft until...

* [x] Feedback gathered about general approach, design
* [x] Tests added (after that)

### To try it out

Check out this branch, run `make dev` and visit the new [Dialogs pattern library page](http://localhost:4001/ui-playground/components-dialogs) in your browser. The modal examples are last on the page.

### Screenshots

Note that the pattern-library examples include all supported Dialog bells and whistles, including an icon in the header, which may not be applicable to modals. But it's helpful to see everything in the mix...

`Modal`, narrow viewport (with initial-focus input):

The same `Modal`, wider viewport:

image

`ConfirmModal`, narrow viewport:

`ConfirmModal`, narrow viewport, with excessively long content (NB, this example is not included in source):

`ConfirmModal`, wider viewport:

C1268CBD-4F24-4A0E-A10F-037E7E8C80EA

Fixes https://github.com/hypothesis/frontend-shared/issues/86

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -1,0 +1,88 @@
+import { useRef } from 'preact/hooks';
+
+import { useElementShouldClose } from '../hooks/use-element-should-close';
+import { Dialog } from './Dialog';
+import { LabeledButton } from './buttons';
+
+/**
+ * @typedef {import('./Dialog').DialogProps} DialogProps
+ */
+
+/**
+ * @typedef ModalBaseProps
+ * @prop {() => void} onCancel - `onCancel` is required for Modals
+ */
+
+/**
+ * @typedef {DialogProps & ModalBaseProps} ModalProps
+ */
+
+/**
+ * A modal dialog. Presents a dialog with an overlay background. Will close
+ * if user clicks/taps outside of it.
+ *
+ * @param {ModalProps} props
+ */
+export function Modal({ children, onCancel, ...restProps }) {
+  const modalContainerRef = useRef(/** @type {HTMLDivElement | null} */ (null));
+
+  useElementShouldClose(modalContainerRef, true /* isOpen */, () => {
+    if (onCancel) {
+      onCancel();
+    }
+  });
+
+  return (
+    <>
+      <div className="Modal__overlay" />
+      <div className="Modal" ref={modalContainerRef}>
+        <Dialog onCancel={onCancel} {...restProps}>
+          {children}
+        </Dialog>
+      </div>
+    </>
+  );
+}
+
+/**
+ * @typedef ConfirmModalBaseProps
+ * @prop {string} message - Main text of the modal message
+ * @prop {string} confirmAction - Label for the "Confirm" button
+ * @prop {() => void} onConfirm - Callback invoked if the user clicks the "Confirm" button
+ * @prop {() => void} onCancel - Callback invoked if the user cancels
+ *
+ * @typedef {Omit<ModalProps, 'buttons' | 'children'> & ConfirmModalBaseProps} ConfirmModalProps
+ */
+
+/**
+ * A modal that emulates a `window.confirm` interface:
+ * request a boolean yes/no confirmation from the user.
+ *
+ * @param {ConfirmModalProps} props
+ */
+export function ConfirmModal({
+  message,
+  confirmAction,
+  onConfirm,
+  onCancel,
+  ...restProps
+}) {
+  return (
+    <Modal
+      onCancel={onCancel}
+      buttons={[
+        <LabeledButton
+          key="ok"
+          onClick={onConfirm}
+          variant="primary"
+          data-testid="confirm-button"
+        >
+          {confirmAction}
+        </LabeledButton>,
+      ]}
+      {...restProps}
+    >
+      <p>{message}</p>
+    </Modal>
+  );
+}
--- a/src/pattern-library/components/patterns/DialogPatterns.js
+++ b/src/pattern-library/components/patterns/DialogPatterns.js
@@ -1,51 +1,48 @@
 import { createRef, render } from 'preact';
 import { useState } from 'preact/hooks';
-import { Dialog, LabeledButton } from '../../../';
+import { ConfirmModal, Dialog, LabeledButton, Modal } from '../../../';
 
 import { PatternPage, Pattern } from '../PatternPage';
 
 /**
- * Render a Dialog within the `container`, and invoke
+ * Render a Dialog or Modal within the `container`, and invoke
  * `setDialogIsOpen` as needed to alert caller to state changes. Provides the
  * ability to open and close a Dialog demo. We don't want to render a Dialog
  * until it's opened because it will grab focus when it first mounts.
  *
  * @param {Object} options
- *   @prop {HTMLElement} container - Element to render the Dialog inside of
- *   @prop {(isOpen: boolean) => void} setDialogIsOpen - callback to call when
+ *   @param {import('preact').FunctionComponent} options.DialogComponent - Which Dialog
+ *   or Dialog-wrapping component (function) to use;
+ *   a reference to `Dialog`, `Modal`, `ConfirmModal`, e.g.
+ *   @param {HTMLElement} options.container - Element to render the Dialog inside of
+ *   @param {(isOpen: boolean) => void} options.setOpen - callback to call when
  *     the Dialog state changes: opened (true) or closed/removed (false)
+ *   @param {Object} [options.props] - Extra prop(s) for Dialog component
  */
-const showDialog = ({ container, setDialogIsOpen }) => {
+const showDialog = ({ DialogComponent, container, setOpen, props }) => {
   const initialFocusRef = createRef();
 
   const close = message => {
     if (message) {
       alert(message);
     }
-    render(null, container);
-    setDialogIsOpen(false);
+    if (container) {
+      render(null, container);
+    }
+    setOpen(false);
   };
 
-  const extraButtons = [
-    <LabeledButton key="maybe" onClick={() => close('You chose maybe!')}>
-      Maybe
-    </LabeledButton>,
-    <LabeledButton
-      key="yep"
-      onClick={() => close('You chose "Do it!"')}
-      variant="primary"
-    >
-      Do it!
-    </LabeledButton>,
-  ];
+  if (!container) {
+    return null;
+  }
 
-  render(
-    <Dialog
-      buttons={extraButtons}
+  return render(
+    <DialogComponent
       icon="edit"
       initialFocus={initialFocusRef}
       title="Basic dialog with icon"
       onCancel={() => close()}
+      {...props}
     >
       <div>
         <p>This is an example of a dialog.</p>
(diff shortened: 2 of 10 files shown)


src/components/Modal.js
```diff
@@ -0,0 +1,72 @@
+import { Dialog, LabeledButton, useElementShouldClose } from '../';
+
+import { useRef } from 'preact/hooks';
+
+/**
+ * @typedef {import('./Dialog').DialogProps} ModalProps
```
lyzadanger, line 6: `Modal` takes the same props as `Dialog`. I don't know if we have a convention about renaming like I've done here (or _not_ renaming).

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -1,51 +1,42 @@
 import { createRef, render } from 'preact';
 import { useState } from 'preact/hooks';
-import { Dialog, LabeledButton } from '../../../';
+import { ConfirmModal, Dialog, LabeledButton, Modal } from '../../../';
 
 import { PatternPage, Pattern } from '../PatternPage';
 
 /**
- * Render a Dialog within the `container`, and invoke
+ * Render a Dialog or Modal within the `container`, and invoke
```
lyzadanger, line 8: This code is merely supportive to the demos/examples on this page.

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -59,18 +50,67 @@ const showDialog = ({ container, setDialogIsOpen }) => {
...
       container: document.getElementById('#dialog1'),
-      setDialogIsOpen,
+      setIsOpen: setDialogIsOpen,
+      props: {
+        buttons,
+      },
+    });
+  };
+
+  const openModal = () => {
```
lyzadanger, line 90: I recognize that the amount of supporting code for these examples/demos is on the excessive side. I'm hoping that we can simplify this as we go, possibly integrating some of the utility functionality in the client's `prompts` module or something.

styles/mixins/patterns/_molecules.scss
```diff
@@ -59,3 +59,33 @@ $-space: var.$layout-space;
...
+  width: 90vw;
+  max-width: 90vw;
+  max-height: 90vh;
+
+  // For wider viewports:
+  // - positioning: keep horizontal centering but bring top nearer to top
+  //     of viewport.
+  // - sizing: Set a min-width to make sure the modal isn't too dinky. Set
+  //     a max-width to make sure it isn't too ginormously wide on wide screens.
+  @media screen and (min-width: 48rem) {
```
lyzadanger, line 80: Bah, I had it in an earlier comment here but then deleted it and neglected to re-add: this `48rem` magic is a rem-translation of the `768px` tablet breakpoint from the client. I will make sure to add commenting.

robertknight:
Some feedback on the pattern library unrelated to the changes in this PR: The different font sizes for the component description and example description looks a bit odd, possibly because they are visually part of the same region (that is to say, there is no border or background separating the two paragraphs):
 
Pattern library typography

lyzadanger:
> Some feedback on the pattern library unrelated to the changes in this PR: The different font sizes for the component description and example description looks a bit odd, possibly because they are visually part of the same region (that is to say, there is no border or background separating the two paragraphs):

Noted! There is _work to do_ here, for sure.

robertknight:
I noticed that the vertical positioning of modals is sensitive to the width of the dialog as well as the height. Wide viewport:

Narrow viewport

Same viewport height but narrower:

Wide viewport

To me it would make more sense if the vertical position depended only on the height.

src/pattern-library/components/patterns/MoleculePatterns.js
```diff
@@ -98,6 +101,58 @@ export default function MoleculePatterns() {
...
+              >
+                Show example
+              </LabeledButton>
+              <div
+                className="hyp-u-overlay"
+                style={
+                  showExample1 ? 'visibility:visible' : 'visibility:hidden'
+                }
+              >
+                <div className="hyp-modal">
```
robertknight, line 125: Ah - I think I'm beginning to appreciate how the molecules / utility classes / components fit together now. This starting to come together nicely.

styles/mixins/patterns/_molecules.scss
```diff
@@ -59,3 +59,33 @@ $-space: var.$layout-space;
...
+  // For narrower viewports/default:
+  // - positioning: centered vertically and horizontally in the viewport
+  // - sizing: fills up most of the available horizontal space
+  width: 90vw;
+  max-width: 90vw;
+  max-height: 90vh;
+
+  // For wider viewports:
+  // - positioning: keep horizontal centering but bring top nearer to top
+  //     of viewport.
```
robertknight, line 77: I noted this in another comment, but I think we might want to make the vertical centering behavior dependent on the height rather than the width of the viewport.

styles/mixins/patterns/_molecules.scss
```diff
@@ -59,3 +59,33 @@ $-space: var.$layout-space;
...
+  // - sizing: Set a min-width to make sure the modal isn't too dinky. Set
+  //     a max-width to make sure it isn't too ginormously wide on wide screens.
+  @media screen and (min-width: 48rem) {
+    top: 10vh;
+    transform: translate(-50%, 0);
+
+    width: auto;
+    min-width: 28rem;
+    max-width: 44rem; // default for older browsers
+    max-width: min(44rem, 90vw);
```
robertknight, line 87: Makes sense to have a fallback. From https://developer.mozilla.org/en-US/docs/Web/CSS/min()#browser_compatibility it looks like Safari has supported `min` for a long time but support in Chrome and Firefox is much more recent. That's unusual.

robertknight (commented):
Overall this looks great. I didn't have a lot of comments to add. What use cases did you have in mind for `Dialog`s that are not `Modal`s?

lyzadanger:
> I didn't have a lot of comments to add. What use cases did you have in mind for Dialogs that are not Modals?

I wanted to evaluate replacing a couple of what are now `Panel` elements in the client with `Dialog`—the salient difference being that `Dialog` can route focus and we had an a11y flag on a couple of the `Panel` elements in the client (e.g. the one that alerts you that you need to log in before you can annotate) because of focus issues.

src/components/Modal.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * A modal dialog. Presents a dialog with an overlay background. Will close
+ * if user clicks/taps outside of it.
+ *
+ * @param {ModalProps} props
+ */
+export function Modal({ children, onCancel, ...restProps }) {
+  const modalContainerRef = useRef(/** @type {HTMLDivElement | null} */ (null));
+
+  useElementShouldClose(modalContainerRef, true, () => {
```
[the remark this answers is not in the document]
lyzadanger: Noted and updated. This pattern is sensible and wise...

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -1,51 +1,42 @@
...
 
 /**
- * Render a Dialog within the `container`, and invoke
+ * Render a Dialog or Modal within the `container`, and invoke
  * `setDialogIsOpen` as needed to alert caller to state changes. Provides the
  * ability to open and close a Dialog demo. We don't want to render a Dialog
  * until it's opened because it will grab focus when it first mounts.
  *
  * @param {Object} options
+ *   @prop {import('preact').FunctionComponent} DialogComponent - Which Dialog
```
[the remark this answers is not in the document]
lyzadanger: One day, I'll summit the typing mountain and look out at the glorious view. Until then, thank you for being my faithful sherpa.

src/pattern-library/components/patterns/MoleculePatterns.js
lyzadanger: Well, that's good! They are quite useful at least in the pattern-library context at present...

styles/mixins/patterns/_molecules.scss
lyzadanger: It's good feedback. I'm working on splitting the media-query-based differentiation out into two now:

* Screens that are _wide_ enough: set a reasonable min- and max-width so that the modal isn't too dinky but also doesn't stretch way out
* Screens that are _tall_ enough: position the modal nearer to the top of the viewport

e.g. Narrow, but tall screen:

image

src/pattern-library/components/patterns/MoleculePatterns.js
```diff
@@ -8,6 +10,7 @@ import {
 import { IconButton, LabeledButton } from '../../../';
 
 export default function MoleculePatterns() {
+  const [showExample1, setShowExample1] = useState(false);
```
robertknight, line 13: "Example1" is a bit vague and could be misleading if you decide to re-order examples in the page. Maybe `showModalExample` or similar?

src/pattern-library/components/patterns/MoleculePatterns.js
```diff
@@ -98,6 +101,58 @@ export default function MoleculePatterns() {
...
+                variant="primary"
+                onClick={() => setShowExample1(true)}
+              >
+                Show example
+              </LabeledButton>
+              <div
+                className="hyp-u-overlay"
+                style={
+                  showExample1 ? 'visibility:visible' : 'visibility:hidden'
+                }
```
robertknight, line 123:
```suggestion
style={ visibility: showExample1 ? 'visible' : 'hidden' }
```

The use of strings as values for the `style` prop in Preact is technically supported, although I consider it preferable to pass an object of CSS property names and values instead (as if accessing the `element.style` property). React only supports object values as a form of XSS protection (eg. that could happen if using user input in the style string).

styles/mixins/patterns/_molecules.scss
```diff
@@ -59,3 +59,32 @@ $-space: var.$layout-space;
...
+  @media screen and (min-width: 48rem) {
+    width: auto;
+    min-width: 28rem;
+    max-width: 44rem; // default for older browsers
+    max-width: min(44rem, 90vw);
+  }
+
+  // For viewports with more vertical room, adjust vertical positioning to
+  // near the top of the viewport.
+  @media screen and (min-height: 48rem) {
```
robertknight, line 85: `48rem` seems quite tall as a minimum value for me. In Safari on my 13" MBP I can only just get `window.innerHeight` to be that tall if I make my window full screen.

styles/mixins/patterns/_molecules.scss
```diff
@@ -59,3 +59,32 @@ $-space: var.$layout-space;
...
 }
+
+// A responsive sized and positioned container for modal content
+@mixin modal {
+  // The modal container is positioned horizontally- and vertically-centered
+  @include layout.fixed-centered;
+
+  // This sizing applies to smaller viewports
+  width: 90vw;
+  max-width: 90vw;
```
robertknight, line 70: I noticed a jump as the dialog shrinks in size when going from a width that is less than the `min-width` media query below to one that is greater.

robertknight (approved):
I had some minor queries but overall this looks good. Please have a look through and make changes as you see fit.

src/components/Modal.js
```diff
@@ -0,0 +1,88 @@
+import { Dialog } from './Dialog';
+import { LabeledButton } from './buttons';
+import { useElementShouldClose } from '../hooks/use-element-should-close';
+
+import { useRef } from 'preact/hooks';
```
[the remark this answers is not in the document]
lyzadanger: Thanks for catching this; likely an artifact from when I was moving `import`s around to debug that `$imports.$mock` conundrum.

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -1,51 +1,48 @@
...
  * ability to open and close a Dialog demo. We don't want to render a Dialog
  * until it's opened because it will grab focus when it first mounts.
  *
  * @param {Object} options
- *   @prop {HTMLElement} container - Element to render the Dialog inside of
- *   @prop {(isOpen: boolean) => void} setDialogIsOpen - callback to call when
+ *   @param {import('preact').FunctionComponent} options.DialogComponent - Which Dialog
+ *   or Dialog-wrapping component (function) to use;
+ *   a reference to `Dialog`, `Modal`, `ConfirmModal`, e.g.
+ *   @param {HTMLElement|null} options.container - Element to render the Dialog inside of
```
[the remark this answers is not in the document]
lyzadanger: OK, yep, I'll move the casting to where I need it to clarify that `container` is required here.

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -113,6 +159,67 @@ export default function DialogPatterns() {
...
+          <div className="Example__content">
+            <div className="Example__usage">
+              <p>
+                Close the modal by clicking the close (X) button, the cancel
+                button or clicking anywhere outside of it.
+              </p>
+            </div>
+            <div className="Example__demo hyp-frame">
+              <div>
+                <div id="#modal1" />
```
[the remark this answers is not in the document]
lyzadanger: Deeerrrrrrrrpppppp.

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -113,6 +159,67 @@ export default function DialogPatterns() {
...
+                {!modalIsOpen && (
+                  <LabeledButton variant="primary" onClick={openModal}>
+                    Open modal
```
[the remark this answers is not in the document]
lyzadanger: Ah! The first example on this page (Dialog) gets confusing if you leave the button extant when it's open, but the examples added here don't (Modals), so I'll rectify this.

src/pattern-library/components/patterns/MoleculePatterns.js
lyzadanger: Roger.

src/pattern-library/components/patterns/MoleculePatterns.js
lyzadanger: Renamed.

styles/mixins/patterns/_molecules.scss
lyzadanger: OK! Let's try `32rem` for the height breakpoint for now. We can keep adjusting...

styles/mixins/patterns/_molecules.scss
lyzadanger: What would your preferred behavior be? I'm never sure how best to optimize for the "but it isn't animated/smooth when resizing a screen" issue...

PR #89, opened 2021-06-03
Add basic clean-theme support and update `card`, `frame` patterns

### LINKED ISSUE #85: Implement theme-clean variants of `card` pattern
There are some "clean-theme" affordances in the `client` version of the "card" pattern styling, which is used by several components, including panels and dialogs. Those affordances need to be understood, reproduced here and represented in pattern library examples of those patterns and components.

This needs to be done before wholesale replacement of Panel and Dialog components across various applications with shared versions.

Part of #83

lyzadanger:
This PR solves https://github.com/hypothesis/frontend-shared/issues/85 by:

* adding a `theme` mixin that other mixins can use to apply theme-specific styles
* Updating the `card` and `frame` mixins to theme for the `clean` theme

I'm hoping that the changes themselves will obviate further explanation here (that's the goal here, anyway).

~~To finish https://github.com/hypothesis/frontend-shared/issues/85, after #88 is resolved, we'll want to go back in and see how the modal components look in a clean theme. We may wish, e.g., to apply borders to modals even in the clean theme.~~ (This is done)

Fixes #85

### DIFF
--- a/styles/mixins/patterns/_molecules.scss
+++ b/styles/mixins/patterns/_molecules.scss
@@ -2,6 +2,7 @@
 
 @use '../atoms';
 @use '../layout';
+@use '../themes';
 
 $-border-radius: var.$border-radius;
 $-color-background: var.$color-background;
@@ -14,11 +15,17 @@ $-color-background: var.$color-background;
 /**
  * Give an element a border, background color and internal vertical spacing
  */
-@mixin frame($with-hover: false) {
+
+@mixin frame {
   @include layout.block;
   @include atoms.border;
   border-radius: $-border-radius;
   background-color: $-color-background;
+
+  @include themes.theme('clean') {
+    // A frame should not have any borders in the clean theme.
+    border: none;
+  }
 }
 
 /**
@@ -36,9 +43,13 @@ $-color-background: var.$color-background;
       @include atoms.shadow($active: true);
     }
   }
+
   width: 100%;
 
-  // TODO: Add clean-theme affordances
+  @include themes.theme('clean') {
+    // A card should have no shadows at all in the clean theme.
+    box-shadow: none;
+  }
 }
 
 /**
--- a/src/pattern-library/components/patterns/MoleculePatterns.js
+++ b/src/pattern-library/components/patterns/MoleculePatterns.js
@@ -29,6 +29,14 @@ export default function MoleculePatterns() {
               <div className="hyp-u-border">Child content in a frame.</div>
             </div>
           </PatternExample>
+          <PatternExample details="content in a frame: clean theme">
+            <div className="theme-clean">
+              <div className="hyp-frame">
+                Content within a frame in the clean theme. The frame itself has
+                no borders when in the clean theme.
+              </div>
+            </div>
+          </PatternExample>
         </PatternExamples>
       </Pattern>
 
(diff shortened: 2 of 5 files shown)


styles/mixins/patterns/_molecules.scss
```diff
@@ -15,13 +16,18 @@ $-space: var.$layout-space;
...
+@mixin frame {
   @include atoms.border;
   @include layout.vertical-rhythm;
 
   padding: $-space;
   border-radius: $-border-radius;
   background-color: $-color-background;
+
+  @include themes.theme('clean') {
+    // A frame should not have any borders in the clean theme.
```
lyzadanger, line 28: Both this comment and the one further down in this module are "x is 5"-type comments, but the point I'm trying to make is that we should _document_ theme deviations and their intent.

styles/mixins/patterns/_molecules.scss
```diff
@@ -15,13 +16,18 @@ $-space: var.$layout-space;
 /**
  * Give an element a border, background color and internal vertical spacing
  */
-@mixin frame($with-hover: false) {
```
lyzadanger, line 18: Derp. This argument was unused.

robertknight (approved):
The "clean" theme mainly exists to satisfy the needs of a specific customer in the client. We may need to have it in the pattern library for that reason but we probably don't want to encourage its use elsewhere. Perhaps this is just something that can be noted in the docs. I wonder whether this mechanism could potentially support something more generally useful like a dark theme in future?

lyzadanger:
> The "clean" theme mainly exists to satisfy the needs of a specific customer in the client.

And don't I know it! And there's more, because to get the intended full effect of the clean theme, you also need to configure the client with several `branding` configuration properties.

I'm going to think about this a little more. Maybe this customization belongs entirely in the client. There are kind of two problems involved here:

* Support for theming, in general. This approach would move toward that, but we don't _need_ to solve that problem yet, necessarily.
* Support for this specific, customer-related theming which only applies to the Sidebar (and in a couple of places, the Notebook). That needs to be solved before we can use shared Dialogs/Panels/Modals in the client, but doesn't need to be solved _here_ necessarily.

lyzadanger:
After some soul-searching and reflection, I would like to opt to move forward with this approach to theming, in which this package applies some clean-theme styles.

The “clean” theme is only relevant to the `client` application, true. 

But there’s no way to implement the theming affordances for the client-only _clean_ theme without leakage about that theming in some way:

* With this approach, this project’s package has to “know” about the clean theme and provide styling affordances for it.
* If we did the clean-theming in the `client` project, however, the `client` would have to “know” about the package enough to extend classes or mixins. And getting the theming right is easier if a lower-level mixin (in this case, the `card` and `frame` patterns) are themed. I don’t especially want to make those mixins part of the public API, especially not now.

In looking at those two, the first (this PR’s direction) seems less problematic to me at present, additionally because:

* The overall amount of stying needed to effect the clean theme is fairly minimal (there just isn’t that much of it). So, yes, this package knows a little about the clean theme, but it’s not a huge pile.
* It will centralize theming. Say we introduce a dark mode or something. We’ll be able to implement it with the same theming-mixin pattern and see how the two (dark vs. clean) play together (or don’t).
* No changes are required in the applications (i.e. `client`) to use the components and patterns; they just work out of the box.

A side note about Modals: Although Modals are styled with the `card` and `frame` patterns (through multi-level inheritance), which have clean-theme affordances, they will not apply clean-theme styles in the client when the clean theme is activated. That is, the `Modal` components look the same in the client whether or not clean theme is activated. This is actually _fine_, as modals look better with a border, even in the clean theme. The reason the styles aren’t picked up is that `Modal`s are attached to `document.body` when they’re rendered, and are not inside the application container that has the `theme-clean` class. tl;dr no user-facing changes will arise when these changes are released because the one shared component in current use that this would ostensibly affect is _not_ affected.

robertknight:
> The reason the styles aren’t picked up is that Modals are attached to document.body when they’re rendered, and are not inside the application container that has the theme-clean class.

This is potentially a little brittle. We can probably accept the risk for the moment, but if we can find a way to more directly enforce the non-usage of the "clean" theme inside modals that might prevent accidental breakage in future.

lyzadanger:
> This is potentially a little brittle. We can probably accept the risk for the moment, but if we can find a way to more directly enforce the non-usage of the "clean" theme inside modals that might prevent accidental breakage in future.

I'm working on replacing `Panel` components in the client right now with the shared variant, and will address this in that work (it sounds unrelated, but it actually is).

PR #90, opened 2021-06-04
Add prefixes to all component class names to prevent collisions

lyzadanger:
Problem:

* This package's SASS endpoint will generate CSS styles for all components—there is no supported ability for a package user to be selective about which components they want styles for at present[1]
* This package is increasingly containing generically-named components that have local variants of the same name in applications: `Panel`, `Dialog`, `Modal`.
* There is a danger (nay, more than a danger, a certainty) that there will be collisions after the next package release unless we take some sort of action with regard to the generated CSS class names.

This PR takes an approach of prefixing all component class names (utility and other classes are already prefixed). Details are in the commit message.

We may choose another tactic for component class-naming, I don't care much, but we need to address this in some way before we cut another release of this package.

Note that this PR is currently dependent on #88 — this is more of a convenience to save time going back and re-doing classnames for the modal components—but I can eliminate that dependency if it becomes troublesome.

[1] Changing this is certainly a viable plan, in the longer term


robertknight (approved):
Looks to me like a straightforward solution to the problem of potential conflicts between this library and applications.

Issue #68, opened 2021-05-18
Establish a basis for layout (spacing)


lyzadanger:
In the `client` project, we apply spacing using named spacing variables, e.g. `$layout-space--xxsmall` or `$layout-space--large`. This is better than chaos but has a few drawbacks:

* it's not linear. It goes:
    * xxsmall: 0.25 unit
    * xsmall: 0.5 unit
    * small: 0.75 unit
    * medium/default: 1 unit
    * large: 2 units
    * xlarge: 4 units
 * The size naming isn't particularly intuitive
 
In this repository, for now, we have a single layout unit variable. Utility and component styles and mixins are then multiplying that unit against whatever proportion they feel like, for example the current `actions` molecule mixin:

```scss
@mixin actions($direction: row) {
  @if $direction == row {
    @include layout.row(right);
    @include layout.horizontal-rhythm($-space * 0.75);
  } @else {
    @include layout.column;
    @include layout.vertical-rhythm($-space * 0.5);
  }
}
```

At least things are proportional against the same base unit, but we still have potential chaos.

We have an opportunity to design a better system here...

lyzadanger:
This is, in general, done.

PR #92, opened 2021-06-09
Introduce spacing-unit scale and spacing patterns

lyzadanger:
This PR:

* Defines a `rem`-based “spacing” system that can be used to apply space around and between elements, based on a scale from 0 to 9. 
* Replaces the existing “t-shirt-sizing”-like system in the client. Provides a `space` function (in the `variables` module) that returns an appropriate spacing value, versus individual SASS variables for each size.
* Disentangles void-space units (defined here) from future sizing units (font sizing, content/element sizing). These are somewhat conflated in the client’s existing design system.
* Replaces  `horizontal-rhythm` and `vertical-rhythm` mixins and utility classes with `horizontal-spacing` and `vertical-spacing` mixins and utility classes. These new variants obviate the need for the `!important` rule by leveraging specificity instead, have less clever naming, and don’t use a lobotomized owl selector anymore.
* Adds spacing examples to the Layout pattern-library page.
* Provides some foundational underpinnings for immediately-needed work, such as the `Checkbox` component layout and grid layouts needed for LMS.

There are no user-visible changes as a result of this PR, nor any changes to the public API.

Part of https://github.com/hypothesis/frontend-shared/issues/68

### Pattern Library examples

Here are a few screen shots from the [Pattern Library layout page](http://localhost:4001/ui-playground/foundations-layout).

NB: These examples are functional, if not elegant. I didn't want to put too much time in making them really pretty for now.

Screen Shot 2021-06-09 at 9 56 10 AM

Screen Shot 2021-06-09 at 9 56 41 AM

Screen Shot 2021-06-09 at 9 56 32 AM

### Background

This system bases spacing on a `1rem` base unit. In almost all cases, `1rem` = `16px`, but in the rare cases when a user has adjusted their browser’s base font size (through preferences, typically), our spacing will scale accordingly.

However, `rem`-based units give us more predictability than the previously-used `em`-based units, which could be fiddly to figure out in practice (we tend to have a lot of local `font-size` changes, e.g., in our apps). `rem`-based fonts and spacing are recommended for a11y, because of their proportional reaction to zooming.

By encapsulating the spacing scale within a  `space` function and spacing mixins, we give ourselves the opportunity to potentially overlay a responsive scale later (i.e. different space-unit values based on breakpoints), and developers don’t have to worry about variable naming or exact values—they can twiddle with the space units from 0 - 9 until they get something that looks right (or rely on the defaults for vertical and horizontal spacing).

### What's not included

To keep this PR changeset from being overwhelming, these tasks are left for future PRs:

* Padding classes, mixins, pattern-library examples—these changes contain margin-related spacing only
* Pattern-library font and other visual cleanup (it’s still fairly fugly, acknowledged, but getting a few more foundations in place will make it easier to address)
* Additional utility classes for margins (i.e. for applying margin to an element itself instead of these spacing classes, which apply margins to children)

The spacing system is entirely private to the package at present. The previous internal spacing was based on `1rem` also (i.e. this doesn’t represent a switch internally from `em`s to `rem`s), so there should be no user-visible changes to the pattern and component layout generated by this package. That makes the risks of trying it out a little lower than if we were exposing it. If it feels cumbersome, we can abort!

### Sources

I took the biggest inspirations from the GOV.uk design system and BBC Gel. 

* [Spacing – GOV.UK Design System](https://design-system.service.gov.uk/styles/spacing/) — 0 - 9 scale, but uses `px` units
* [BBC GEL | Spacing Units](https://www.bbc.co.uk/gel/guidelines/spacing-units) - explains the distinction between void spacing and sizing; uses `rem`
* * [Pattern Lab](https://boltdesignsystem.com/pattern-lab/?p=visual-styles-spacing-system) (Bolt) - Didn’t use this, but example of varying scales horizontally and vertically. This felt a little too complex.
* https://designsystem.gov.au/components/core/#spacing has some interesting API ideas about a spacing mixin
* [VTEX Styleguide](https://styleguide.vtex.com/#!/Spacing) - Sort of similar idea
* [Hudl Uniform](https://uniform.hudl.com/guidelines/layout/space/design) - Another argument for basing on `rem`
* [Bulb Design System](https://design.bulb.co.uk/styles/spacing) - A little cleverness in naming; they have something analogous to our spacing pattern: [Bulb Design System](https://design.bulb.co.uk/styles/spacing#vspace)
* [Spacing | Skyline, Benevity’s Design System.](https://skyline.benevity.org/design-language/spacing/)

And a comprehensive explanation of `rem` vs `em` vs `px` in scalability:

* [Pixels vs. Relative Units in CSS: why it’s still a big deal - 24 Accessibility](https://www.24a11y.com/2019/pixels-vs-relative-units-in-css-why-its-still-a-big-deal/)

### DIFF
--- a/src/pattern-library/components/patterns/LayoutPatterns.js
+++ b/src/pattern-library/components/patterns/LayoutPatterns.js
@@ -1,3 +1,4 @@
+import classnames from 'classnames';
 import { useState } from 'preact/hooks';
 
 import { LabeledButton } from '../../../';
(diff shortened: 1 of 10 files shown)


src/pattern-library/components/patterns/LayoutPatterns.js
```diff
@@ -9,12 +10,144 @@ import {
   PatternExample,
 } from '../PatternPage';
 
+function SquareBlock() {
```
lyzadanger, line 13: I will type/comment these demo components once the overall direction is approved.

styles/mixins/_layout.scss
```diff
@@ -107,3 +91,9 @@ $-color-overlay: var.$color-overlay;
   right: 0;
   background-color: $-color-overlay;
 }
+
+// A content block, with default padding and vertical spacing between elements
```
lyzadanger, line 95: I extracted this out of the `frame` molecule, just kind of adjusting things as it feels natural, but it may not stick around, we'll see.

styles/pattern-library.scss
```diff
@@ -101,10 +101,7 @@ pre {
   p {
     margin: 1em;
     margin-left: 0;
-  }
-
-  & > p:first-of-type {
-    font-size: 125%;
+    font-size: 1.25rem;
```
lyzadanger, line 104: I bumped this for immediate readability, but it still looks a little crap. More later...

styles/util/_layout.scss
```diff
@@ -32,6 +20,10 @@
   @include layout.column;
 }
 
+.hyp-u-stretch {
```
lyzadanger, line 23: Needed for an example.

styles/variables/_index.scss
```diff
@@ -1,8 +1,40 @@
...
+    /* 6 */ 1.5rem,
+    /* 7 */ 2rem,
+    /* 8 */ 4rem,
+    /* 9 */ 8rem
+  );
+  @if $space-unit == 0 {
+    // The non-space space
+    @return $-spaces-0;
+  } @else {
+    // This will throw an error if index is out of bounds
```
lyzadanger, line 37: I added this comment because I thought a question might come up here in the review cycle. Basically, SASS is going to throw if the index is invalid, and I didn't feel that defining our own error handling added much (as the error message is pretty clear).

styles/mixins/_layout.scss
```diff
@@ -48,39 +47,24 @@ $-color-overlay: var.$color-overlay;
...
- * This mixin uses `!important` such that it can compete with specificity
- * of reset rules that set some of our element's margins to 0. That allows
- * these rules—which are applied to a parent element—to be able to assert
- * margins, as it should be able to do.
- *
- * This mixin uses a "lobotomized owl" selector.
- * See https://css-tricks.com/lobotomized-owls/
- *
- * @param $size [$-space]: Spacing size (padding)
+ * @param {length} [$size] - Relative size of spacing, 0 - 9. Default 5
```
LMS007, line 52:
```suggestion
 * @param {number} [$size] - Relative size of spacing, 0 - 9. Default 5
```

To me it's clear from the description that this is a number range. Unless there is a specific reason to use "length" which I'm not sure is standard?

styles/mixins/_layout.scss
```diff
@@ -48,39 +47,24 @@ $-color-overlay: var.$color-overlay;
...
- * children (i.e. put equal space between children).
- *
- * This mixin uses `!important` such that it can compete with specificity
- * of reset rules that set some of our element's margins to 0. That allows
- * these rules—which are applied to a parent element—to be able to assert
- * margins, as it should be able to do.
+ * Put `$size` of horizontal space between all immediate children.
  *
- * @param $size [$-space * 0.5] - Size of horizontal margin between child elements
+ * @param {length} [$size] - Relative size of spacing, 0 - 9. Default 3
```
LMS007, line 63: same here

styles/mixins/_layout.scss
```diff
@@ -48,39 +47,24 @@ $-color-overlay: var.$color-overlay;
...
  */
-@mixin horizontal-rhythm($size: $-space * 0.5) {
-  & > * + * {
-    margin-left: $size !important;
+@mixin horizontal-spacing($size: 3) {
+  & > :not(:first-child) {
```
LMS007, line 66: Seems reasonable!

styles/variables/_index.scss
LMS007: That seems fine, thanks

LMS007 (approved):
Thanks Lyza. I gave this is a close look and the proposed changes are clear and accurate. I had some small suggestions below but those can be fixed up if you agree and then merged.

src/pattern-library/components/patterns/LayoutPatterns.js
```diff
@@ -9,12 +10,144 @@ import {
...
+        <div>{size}</div>
+      )}
+      <div className={classnames(layoutClass, sizeClass, 'hyp-u-border')}>
+        <SquareBlock />
+        <SquareBlock />
+        <SquareBlock />
+        <SquareBlock />
+        <SquareBlock />
+      </div>
+      {direction === 'horizontal' && (
```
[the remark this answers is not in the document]
lyzadanger: Interesting idea! The, ahem, layout of this layout-pattern page is a little in flux right now, but I'll keep this idea in my back pocket.

styles/mixins/_layout.scss
lyzadanger: I am sort of forcing a round peg into a square hole here, but I was trying to be explicit that the "type" (I can't put enough quotation marks around that here) should be a CSS `length`. https://developer.mozilla.org/en-US/docs/Web/CSS/length

src/pattern-library/components/patterns/LayoutPatterns.js
lyzadanger: ...And now done.

PR #95, opened 2021-06-10
Add styling, design pattern for checkbox components

lyzadanger:
Motivation: Define a design pattern for checkboxes to apply to a sprint objective in the `lms` project.

This started out sounding simple: I wanted to apply some horizontal-spacing between a checkbox input and its label, and get things vertically aligned correctly. It turns out this whole thing (styling native radio, checkbox inputs) is sort of a bottomless pit. After spending some time Googling and reading and cursing—some of the proposed workarounds made me cringe by how convoluted they are—I found a [comprehensive write-up from Sara Soueidan](https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/), a trusted name in the a11y and web-design space.

**In short: This implementation styles the appearance of a checkbox with an SVG image, but leaves the input element in place—the user still interacts with the input element.**

This PR contains a simplified variant of Sara’s solution. It:

* Allows us to style a checkbox (both checked and unchecked state) using a relatively simple SVG
* It makes the checkbox input element “not visible” in a way that still leaves it in place, such that the user is interacting with the checkbox input still,
* and the checkbox is accessible to screen readers and touch-screen-readers.
* Keyboard focus styling is present

Changes made to the existing `Checkbox` and `LabeledCheckbox` components to accommodate this:

* Sara makes a persuasive argument that wrapping a checkbox-label-text pairing inside of a `label` element makes an overall larger interaction area, so I’ve done that.
* I’ve removed the option (for now) to put the label before the checkbox (`position` prop). Not that we couldn’t support this, but we don’t have a use case yet, and I figure we can cross that bridge as needed.

This PR also introduces an `override` mixin (the need to override pattern-imposed styles came up in this work).

### See it

Check out this branch, `make dev` and head over to the [Forms Pattern-library page](http://localhost:4001/ui-playground/components-forms).

Current appearance:

image

In situ in LMS assignment configuration flow:

Screen Shot 2021-06-10 at 9 29 43 AM

Showing keyboard focus:

Screen Shot 2021-06-10 at 11 12 59 AM

One my styling variants used a brand-red when checked, but I felt that this was a bit too much:

Screen Shot 2021-06-10 at 9 21 38 AM

(We can totally tweak how that SVG looks as we see fit).

**Alert**: This PR should not be landed until #94 is resolved in some way, as it has danger of breaking apps that use it! 

Fixes https://github.com/hypothesis/frontend-shared/issues/26


src/components/Checkbox.js
```diff
@@ -17,13 +19,15 @@
...
  *
  * @typedef {Omit<CheckboxProps, 'children'> & LabeledCheckboxBaseProps} LabeledCheckboxProps
  */
 
 /**
- * A checkbox input
+ * A checkbox input.
+ *
+ * A checkbox component is a combination of an <input> element and a sibling
+ * <svg> element that is used for the visual appearance of the checkbox.
```
lyzadanger, line 30: This isn't quite accurate and is stale from a previous implementation cycle; this should read `<SvgIcon>` instead of `<svg> element`

styles/components/Checkbox.scss
```diff
@@ -0,0 +1,103 @@
...
+  // Size and align the span containing the <svg>
+  & + span {
+    @include base.override {
+      // The parent container is applying horizontal-spacing.
+      // Don't add left-margin to this element.
+      // The input element, which comes first, isn't actually
+      // visible, so this (second) element should have no left-margin.
+      margin-left: var.space(0);
+    }
+    & > svg {
```
lyzadanger, line 63: Discussed with @esanzgar on a call, but, noted: this is one level deeper of SASS nesting than we typically "allow." But given the complexity of this styling, and the "encapsulation" within component styles, I felt that it was easier to understand in this form.

styles/components/Checkbox.scss
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  // Set focus styles on SVG when checkbox is focused
+  &:focus {
+    & + span > svg {
+      @include focus.outline-rule;
+    }
+
+    // TODO polyfill/browser support
```
lyzadanger, line 79: Our focus mixins include a variant that selects on a classname, `.focus-visible`. That presumes the presence of some sort of polyfill that is applying that class? I'm not sure how all of the pieces fit together.

styles/mixins/_focus.scss
```diff
@@ -2,6 +2,11 @@
 
 $-color-focus-outline: var.$color-focus-outline;
 
+@mixin outline-rule($inset: false) {
```
lyzadanger, line 5: I couldn't use the mixins as written because I needed to craft a specific selector. Here I'm externalizing at least part of the rule-set so I can use that.

lyzadanger:
I have pushed an additional commit that registers the needed checkbox icon for these components. This resolves https://github.com/hypothesis/frontend-shared/pull/94 — this PR no longer has any dependencies.

lyzadanger:
Update: I've added the do-not-merge label because I've found an icon-registration issue when testing this branch's build against `lms` and `client` locally. I'll update as I learn more.

styles/mixins/_base.scss
```diff
@@ -0,0 +1,16 @@
...
+ * This mixin serves two purposes:
+ * - Allows developers to be explicit that the contained CSS rule-set overrides
+ *   styling imposed by another pattern
+ * - Increases the specificity of the contained rule-set to aid in overriding
+ *
+ * Current implementation: Use the :not pseudo-class to increase specifity.
+ */
+@mixin override() {
+  // The ID here will increase the specificity of the whole selector
+  &:not(#increase-specificity) {
```
esanzgar, line 13: Interesting.

styles/mixins/_focus.scss
```diff
@@ -2,6 +2,11 @@
...
+  outline: none;
+  box-shadow: 0 0 0 2px $-color-focus-outline if($inset, inset, null);
```
esanzgar, line 7: I am so behind with scss 😢 . I really need to learn more, for example here `if(....)`

esanzgar (approved):
All look well.

Thank you for the clear organisation of the commits.

In the checkbox test it is not longer necessary to wrap it on a `<div>` to pass the a11y test.

styles/mixins/_focus.scss
lyzadanger: Hah! This isn't mine, actually; it's just a moved rule that pre-exists these changes. I also have a little trouble human-parsing the `if...` here!

styles/mixins/_base.scss
lyzadanger: Yah. I want to be sure this isn't super confusing. So please speak up if it just seems crazy pants.

PR #96, opened 2021-06-11
Package to v3.1.1

lyzadanger:
Version the package.

This started as a version-bump to `3.1.0`, but a typo in the `CHANGELOG` led to `3.1.1` because the `3.0.0` tag had already been pushed.


robertknight (approved):
LGTM. In general I feel that straightforward package updates like this (ie. version bump + CHANGELOG update) do not require review, although feel free to request one if you want input on something specific.

lyzadanger:
> In general I feel that straightforward package updates like this (ie. version bump + CHANGELOG update) do not require review

I tend to agree, especially from a PITA perspective: any requested changes here would end up with a new commit hash, which would in turn muck up the tag that has already been pushed at this point. Let's discuss in our next team chat.

PR #109, opened 2021-06-16
Fix `IconButton` horizontal icon alignment on narrow screens

lyzadanger:
There is a media-query that will enforce a minimum width and height on
`IconButton`s to ensure a minimum tap-target size. Before these changes,
icons within `IconButton`s on narrow screens could appear off-center
horizontally.

The icon image is already vertically centered.
Add a horizontal-alignment rule here to ensure that the icon image
always appears in the center of the entire button.

Before, [Pattern-Library buttons page](http://localhost:4001/ui-playground/components-buttons) in narrow viewport:

Screen Shot 2021-06-16 at 1 41 16 PM

After:

Screen Shot 2021-06-16 at 1 40 55 PM

### DIFF
--- a/styles/components/buttons/_mixins.scss
+++ b/styles/components/buttons/_mixins.scss
@@ -94,6 +94,10 @@
         &--large {
           min-width: c.$touch-target-size;
           min-height: c.$touch-target-size;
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
👍🏼

Issue #84, opened 2021-06-02
Document package usage


lyzadanger:
Problem: This package's usage documentation hasn't been updated lately and it's getting out of date. Usage of this package will change again as we update implementation of components. Developers need to know how to use the contents of the package!

Solution: Add usage documentation as a pre-requisite before any updated components are shipped, to ensure it's crystal clear how to use components from the package.

Outcome: It is clear what is included in the package's API, and how to import/use components in applications.

----

Notes:

Review these items currently part of the package API:

- `normalizeKeyName`
- `icons` and registration functions
- `useElementShouldClose` hook

lyzadanger:
This is done!

PR #110, opened 2021-06-17
Add additional utility mixins and classes

lyzadanger:
This is a relatively dull PR that fills in some gaps for utility classes and mixins. These are all still private to the package.

This PR also includes a commit that updates the package README, which was embarrassingly out-of-date.

Motivation: fill in any obvious gaps, and reproduce utility classes/mixins that are in use in other applications.

No user-visible changes at all. All of these mixins and utility classes are still private to the package. But we're getting closer to being able to make some utilities public here...

Part of https://github.com/hypothesis/frontend-shared/issues/74

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,18 +14,12 @@ Your project must have Preact installed as a dependency.
 $ npm install @hypothesis/frontend-shared --save
 ```
 
-### In SASS modules
+### SASS
 
-To import default styling of frontend-shared components, include this line in the main project's SASS.
+To add styles for all shared components to your project's SASS:
 
 ```scss
 @use '@hypothesis/frontend-shared/styles';
-```
-
-Mixins can be imported directly
-
-```scss
-@use "@hypothesis/frontend-shared/styles/mixins" as mixins;
 ```
 
 ### In JS
--- a/styles/util/_atoms.scss
+++ b/styles/util/_atoms.scss
@@ -16,6 +16,10 @@
   @include atoms.border(top);
 }
 
+.hyp-u-border--bottom {
+  @include atoms.border(bottom);
+}
+
 .hyp-u-shadow {
   @include atoms.shadow;
 }
--- a/styles/util/_layout.scss
+++ b/styles/util/_layout.scss
@@ -1,38 +1,9 @@
 @use '../mixins/layout';
 
-// Establish a row flex container
-.hyp-u-layout-row {
-  @include layout.row;
-}
+/* Spacing classes */
 
-// Establish a row flex container with items justified and aligned center
-.hyp-u-layout-row--center {
-  @include layout.row($align: center, $justify: center);
-}
-
-// Establish a row flex container with items justified center
-.hyp-u-layout-row--justify-center {
-  @include layout.row($justify: center);
-}
-
-// Establish a column flex container
-.hyp-u-layout-column {
-  @include layout.column;
-}
-
-.hyp-u-stretch {
-  flex-grow: 1;
-}
-
-// Position the element centered vertically and horizontally in the entire viewport
-.hyp-u-fixed-centered {
-  @include layout.fixed-centered;
-}
-
-// A full-viewport-width and -height semi-opaque backdrop overlay
-.hyp-u-overlay {
-  @include layout.overlay;
-}
+// These classes are applied to a parent element to establish horizontal
+// or vertical spacing between immediate-child elements.
 
 // Apply default horizontal spacing
 .hyp-u-horizontal-spacing {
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -14,20 +14,14 @@ Your project must have Preact installed as a dependency.
...
+### SASS
 
-To import default styling of frontend-shared components, include this line in the main project's SASS.
+To add styles for all shared components to your project's SASS:
 
 ```scss
 @use '@hypothesis/frontend-shared/styles';
 ```
 
-Mixins can be imported directly
```
lyzadanger, line 25: Um, nope, no they cannot be. I don't know how I overlooked this...

styles/util/_atoms.scss
```diff
@@ -16,6 +16,10 @@
   @include atoms.border(top);
 }
 
+.hyp-u-border--bottom {
```
lyzadanger, line 19: This one was just missing, whoops.

styles/util/_layout.scss
```diff
@@ -33,27 +174,3 @@
 .hyp-u-overlay {
   @include layout.overlay;
 }
-
```
lyzadanger, line 36: These are just moved elsewhere in the module.

styles/util/_layout.scss
```diff
@@ -1,20 +1,159 @@
...
+.hyp-u-margin--top {
+  @include layout.margin($side: 'top');
+}
+
+.hyp-u-margin--bottom {
+  @include layout.margin($side: 'bottom');
+}
+
+// Generate margin utility classes for spacing units 0-9
+@for $space-unit from 0 through 9 {
```
robertknight, line 58: The combinatorial size here (10 spacing levels * 5 margin sides) is a bit unfortunate, especially since I'd guess that many of the combinations will not be used. What is the thinking behind utility classes for this vs `margin: var.size($size)` (or `margin-$side: $size`) in component-specific classes?

lyzadanger:
I've moved this into draft state because, as I've been working with these changes in another branch based on this branch, I'm noticing a couple of things that could be ergonomically improved, so I might as well take care of it while I'm in here...

robertknight (commented):
Historically we've styled most components with specific classes that re-use styling via mixins or variables, with the use of a small number of utility classes for very common sets of rules for elements that don't have other styling (eg. `u-layout-row` for `display: flex; flex-direction: row`). On the opposite end of the spectrum are Tailwind-like systems which use CSS classes heavily as design tokens and up putting the styling in the HTML as combinations of these tokens. This PR and other discussion has pushed us more in the direction of having more utility classes. What I'd like to understand, and apologies if there has been recent discussion that I've missed, is:

- How far are we going to push this? ie. when/where should utilities be used vs use of mixins or variables in component-specific styles
- What will be available to consumers outside the package vs only available internally?
- How can we avoid having a significant amount of unused CSS in the package?

styles/util/_layout.scss
lyzadanger: Yes. Hah. Not a yes or no question, I realize.

Your question gets right to the nugget of what I'm working through solving right now. In real-world use, what's the right balance between ergonomics, sensible performance/efficiency (i.e. the combinatorial size issue here), consistency (not having every component be authored differently based on developer whim) and elegance? It remains to be seen. Will `.hyp-u-padding--right--4` be a life-saver, or a ridiculous extra detail glutting our CSS builds?

lyzadanger:
@robertknight Pasting from Slack:

I'm c-n-ping a comment I left in a conversation in Slack for you:

That thinking is a big part of this exercise, and central to why I haven’t yet been willing to make anything—mixins nor utility classes—public yet. At present the only thing that could suffer from any CSS bloat is the pattern library, as these utility and pattern classes are not built into any application CSS.

I’m working toward finding the right balance to exactly the questions you posed in your comment.

Now that the epic for the shared Dialogs/Modals has been satisfied, a plausible next epic to work toward here would be the piece of the puzzle around sorting out the API, utility classes, and mixins, whatnot. https://github.com/hypothesis/frontend-shared/issues/84

Until that is solidified, my strategy has been to keep all mixins and utility classes private to the package such that:

* We aren’t generating a glut of unused or redundant CSS, and.
* We aren’t providing later maintenance angst for ourselves by making public things that we later regret

This gives some room for exploration within the package itself, and helps me case-study test different approaches. To wit, I’m building out some needed pattern-library improvements on a branch based on this branch, and find myself coming back to this (utility-class) branch and making little improvements based on how this feels in a “real” use situation.

Put yet another way, this project is more tricky than most in terms of finding the right line between thinking forward (putting a bunch of stuff in place I think we might need, and finding that some, but not all of it, is useful) and letting things fall into place organically (but having some gaps and lurches). I’m doing my best to bridge this wiggliness by allowing myself some space for experimentation within the package itself, but being extra conservative with what’s exposed “publicly.”

robertknight (approved):
That sounds reasonable. Based on the understanding that the balance between utility/design-token vs component-specific classes is still in flux and therefore the classes are currently private to the package.

PR #112, opened 2021-06-24
Allow for overflow constraints on content within Modals

lyzadanger:
This PR updates the modal pattern such that it is possible to constrain content within its contained dialog using `overflow` rules. There is an additional "Long modal" example on the [Pattern Library's "Dialogs" page](http://localhost:4001/ui-playground/components-dialogs).

Problem we're trying to solve here: sometimes modals contain content that is too tall for the modal container, and app devs should be able to apply `overflow` rules to that content to allow it to scroll if it's too tall (versus making the modal container too tall for the viewport). To enable this, it is necessary to apply some additional height and layout rules to the modal container's immediate child element (the immediate parent of the elements that would want overflow control).

In an immediate sense, we need these constraints to be able to use the `Dialog` in the LMS' VitalSource content-picking flow. Here's what that looks like in situ (i.e. with these changes applied). You can see that the content is scrolling correctly:

Screen Shot 2021-06-24 at 8 47 01 AM

(_Note_: The shared `Modal` is narrower than the LMS-local variant, but that's a function of the content within the modal—the modal default/preferred width is sensible in most cases. We can manage for width locally if we feel it is too narrow in this particular context).

I found these changes tricky to articulate, and probably re-wrote SASS comments about 15 times. 

This PR also rectifies an oversight in which the `Dialog` component assumed the presence of the `cancel` icon.

Part of https://github.com/hypothesis/lms/issues/2870

More examples below. Note that I am unable to capture the scrollbars themselves as they disappear when I grab screenshots.

Scrollable content on narrow screen:

image

On a larger screen:

image

### DIFF
--- a/src/components/Dialog.js
+++ b/src/components/Dialog.js
@@ -1,8 +1,14 @@
 import classnames from 'classnames';
 import { useEffect, useLayoutEffect, useRef, useState } from 'preact/hooks';
 
-import { IconButton, LabeledButton } from '../';
-import { SvgIcon } from './SvgIcon';
+import { IconButton, LabeledButton } from './buttons';
+import { registerIcons, SvgIcon } from './SvgIcon';
+
+// Register the checkbox icon for use
+registerIcons({
+  /** @ts-ignore - TS doesn't understand require here */
+  'hyp-cancel': require('../../images/icons/cancel.svg'),
+});
 
 let idCounter = 0;
 
@@ -121,7 +127,7 @@ export function Dialog({
         </h2>
         {onCancel && (
           <div className="Hyp-Dialog__close">
-            <IconButton icon="cancel" title="Close" onClick={onCancel} />
+            <IconButton icon="hyp-cancel" title="Close" onClick={onCancel} />
           </div>
         )}
       </header>
--- a/src/components/Dialog.js
+++ b/src/components/Dialog.js
@@ -131,7 +131,7 @@ export function Dialog({
           </div>
         )}
       </header>
-      <div>{children}</div>
+      {children}
       <div className="Hyp-Dialog__actions">
         {onCancel && (
           <LabeledButton data-testid="cancel-button" onClick={onCancel}>
--- a/styles/mixins/patterns/_molecules.scss
+++ b/styles/mixins/patterns/_molecules.scss
@@ -74,7 +74,32 @@ $-color-background: var.$color-background;
   }
 }
 
-// A responsive sized and positioned container for modal content
+/**
+ * A modal container, with responsive positioning and sizing.
+ *
+ * A modal should contain one immediate-child element; in most cases, an
+ * element that applies the dialog[1] pattern. Content within the dialog
+ * (grandchild elements of modal) can be managed with `overflow` rules so as not
+ * to exceed the size constraints of the modal within the viewport.
+ *
+ * Example structure follows:
+ *
+ * <modal>
+ *   <dialog>
+ *     <div /> <!-- dialog can contain any arbitrary content -->
+ *     <div style="overflow: auto">
+ *       This content will scroll vertically if it is too tall for the
+ *       available space in the modal.
+ *     </div>
+ *     <div />
+ *     ...
+ *   </dialog>
+ * </modal>
+ *
+ * [1]: The modal's immediate child need not specifically apply the dialog
+ *      pattern, but any element that needs to manage overflow within the modal
+ *      must be a grandchild of the modal container element.
+ */
 @mixin modal {
   // The modal container is positioned horizontally- and vertically-centered
   @include layout.fixed-centered;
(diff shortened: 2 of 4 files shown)


src/components/Dialog.js
```diff
@@ -1,8 +1,14 @@
 import classnames from 'classnames';
 import { useEffect, useLayoutEffect, useRef, useState } from 'preact/hooks';
 
-import { IconButton, LabeledButton } from '../';
-import { SvgIcon } from './SvgIcon';
+import { IconButton, LabeledButton } from './buttons';
```
lyzadanger, line 4: We decided recently that it's better if components import each other directly versus going through the package entry point.

src/components/Dialog.js
```diff
@@ -1,8 +1,14 @@
...
+import { registerIcons, SvgIcon } from './SvgIcon';
+
+// Register the checkbox icon for use
+registerIcons({
+  /** @ts-ignore - TS doesn't understand require here */
+  'hyp-cancel': require('../../images/icons/cancel.svg'),
+});
```
lyzadanger, line 11: When trying out the `Dialog` in my local `lms`, it broke because it assumed the presence of a `cancel` icon. Whoops! We'd already run into this before with the `Checkbox` component, and I overlooked making sure other components registered their own icons. We're using the `hyp-` prefix here to avoid icon collisions.

src/components/Dialog.js
```diff
@@ -121,11 +127,11 @@ export function Dialog({
         </h2>
         {onCancel && (
           <div className="Hyp-Dialog__close">
-            <IconButton icon="cancel" title="Close" onClick={onCancel} />
+            <IconButton icon="hyp-cancel" title="Close" onClick={onCancel} />
           </div>
         )}
       </header>
-      <div>{children}</div>
+      {children}
```
lyzadanger, line 134: This wrapping element needed to be removed such that `{children}`'s elements were immediate children of the `Dialog`, for overflow reasons. I will double-check this change against the `client` to assure it hasn't caused any layout changes there.

src/pattern-library/components/patterns/DialogPatterns.js
```diff
@@ -189,6 +289,34 @@ export default function DialogPatterns() {
...
+            doesn&apos;t exceed available viewport space.
+          </p>
+          <div className="Example__content">
+            <div className="Example__usage">
+              <p>
+                To make something in a modal scroll-able, apply{' '}
+                <code>overflow: auto</code> to the element you wish to contain.
+                This element needs to be an immediate-child element of the{' '}
+                <code>Modal</code>.
+              </p>
```
lyzadanger, line 306: This isn't quite right, argh. I'll fix after getting feedback on the overall approach.

styles/mixins/patterns/_molecules.scss
```diff
@@ -74,7 +74,32 @@ $-color-background: var.$color-background;
   }
 }
 
-// A responsive sized and positioned container for modal content
+/**
```
lyzadanger, line 77: I can't express how hard this was to explain. Please let me know if you can think of any way to make this clearer...

src/components/Dialog.js
esanzgar: Good to know, thanks.

styles/mixins/patterns/_molecules.scss
```diff
@@ -74,7 +74,32 @@ $-color-background: var.$color-background;
...
+ * A modal container, with responsive positioning and sizing.
+ *
+ * A modal should contain one immediate-child element; in most cases, an
+ * element that applies the dialog[1] pattern. Content within the dialog
+ * (grandchild elements of modal) can be managed with `overflow` rules so as not
```
esanzgar, line 82: Could `Modal` component create this direct-child element and set the `overflow` style?

esanzgar (commented):
Lyza, I left a couple of comments.

Maybe next step could be to work to cancel the scrolling on the host page.

I also notice that Safari sets a focus ring around the modal, but not Chrome or Firefox.

styles/mixins/patterns/_molecules.scss
lyzadanger: I want to give authors flexibility over which elements scroll or don't scroll. Not all modals need scrolling content, and not all scrolling content will be the entirety of the modal content, if that makes sense. These rules merely make it _possible_ to apply `overflow`, if that makes sense. I don't want to dictate what overflows.

esanzgar: Yes, it makes sense. Your approach is more flexible but it requires more from the developer.

My comment assumed that the most likely scenario is that the scrollable content will be the entirety of the modal.

styles/mixins/patterns/_molecules.scss
esanzgar: I didn't find any trouble understanding your explanation. I think the explanation is very good.

esanzgar (approved):
Thank you for finding this awesome solution to this tricky CSS problem.

esanzgar:
I have tested this solution on an small iframe and it worked very well.

PR #113, opened 2021-06-24
Refine structure and naming of Pattern-library components and sections

lyzadanger:
This PR consolidates the components used for the Pattern Library pages and adds a new "group" of pages to the Pattern Library. 

The motivation for this change is that I'm working on some input and input-with-button design patterns and there is a component naming conflict: `FormPatterns` was "already taken." This PR's changes moves the former `FormPatterns` to `FormComponents` so that I can use `FormPatterns` for actual, well, form patterns.

There are here three types of pattern-library page components:

* `Foundations`, naming convention for components `*Foundations` — These are the atomic building blocks of the design system, expressed in utility CSS classes prefixed `hyp-u`(note: These utility classes are still private to the package for the moment). Note that the existing `*Foundations` pattern-library pages do not provide examples for all of the existing atoms (yet). But it's a start, so far.
* `Patterns`, naming convention for components `*Patterns` — These are composite design patterns built by combining atomic patterns. These are expressed in CSS class names prefixed with `hyp-` (no `u`). These classes are still private to the package. We're moving in a direction that most, if not all, patterns will have a corresponding component. Patterns here are referred to using lower-case: `modal`, `panel`, etc.
* `Components`, naming convention for components `*Components` — These are, duh, component examples. Typically each component implements a design pattern, e.g. `Panel` is built on `panel`. 

There are no actual changes to any content or anything; just adjusting how these examples are organized within the Pattern Library.

This PR currently depends on #112 for purely convenience/ergonomic reasons, but could be made independent if that PR gets mired.

Pattern Library sections after these changes:

image

### DIFF
--- a/src/pattern-library/components/PlaygroundApp.js
+++ b/src/pattern-library/components/PlaygroundApp.js
@@ -41,6 +41,7 @@ export default function PlaygroundApp({
 
   const routeGroups = [
     { title: 'Foundations', routes: getRoutes('foundations') },
+    { title: 'Patterns', routes: getRoutes('patterns') },
     { title: 'Common Components', routes: getRoutes('components') },
   ];
 
--- a/src/pattern-library/components/patterns/ButtonComponents.js
+++ b/src/pattern-library/components/patterns/ButtonComponents.js
@@ -7,7 +7,7 @@ import {
   PatternExample,
 } from '../PatternPage';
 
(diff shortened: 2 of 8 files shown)


LMS007 (approved):

lyzadanger:
Rebased on `main`.

PR #130, opened 2021-06-29
Add `inputRef` to `TextInput` component

lyzadanger:
Our newly-established convention is that components that wrap a single HTML element should take a `ref` prop named per the type of element rendered. Update `TextInput` accordingly, following the method that `buttons` components use.


LMS007 (approved):
Looks good Lyza

Issue #125, opened 2021-06-28
Rename "organisms" patterns


lyzadanger:
In #122 , we're renaming `molecules` to `containers`.

Currently, the `organisms` patterns only contain the `panel` pattern. Perhaps rename to `panel` (and create new, separate pattern collections/modules for future complex patterns)?

PR #131, opened 2021-06-29
Rename `organisms` patterns to `panels`

lyzadanger:
This PR renames all things that were called "organisms" to "panels". Justifications:

- We're moving away from atomic nomenclature as we grow
- `panel` is the only pattern here currently, and it's complex enough
  to warrant its own pattern module
- The correspondence to the `Panel` and derivative components is
  clarified

Pattern Library, after:

image

Fixes https://github.com/hypothesis/frontend-shared/issues/125


styles/README.md
```diff
@@ -45,14 +45,6 @@ Note that all SASS modules in this project's package can be imported, by deeper
 
 Isolate your customizations within a single module in the consuming application SASS (avoid adding dependencies to this repository's SASS in multiple places for ease of maintenance later). See the `client` for how buttons are customized in SASS. It is the hope that customization will become a rarer exception as these patterns get more established.
 
-### Q: How much do I need to know about Atomic Design to get my job done here?
```
lyzadanger, line 48: I pulled this entire section because the only thing left is `atoms`, and that may not be long for this world.

LMS007 (approved):
👍

PR #132, opened 2021-06-29
Add `spinner` pattern and `Spinner` component

lyzadanger:
This PR adds a shared `spinner` pattern and corresponding `Spinner` component. 

Motivation: The spinners are styled differently in our applications, and their usage within a single application is also inconsistent. In the immediate term, I'd like to use a loading spinner within a component on LMS, so this feels like the right time to drop this in.

~~This implementation is copied over from the `client`, with a little bit of simplification, and I made the `Spinner` component take a `classes` prop for better developer control.~~

This implementation is a variant of the `lms` `Spinner`, using an animated SVG.

~~To implement this, I had to update the `sass` dependency in this package. Needed the `math.div` function. The version here is in line with what the `client` and `lms` have.~~

The above is no longer the case, since the implementation is SVG-based, but there's no harm in bumping SASS.

~~This PR currently depends on #131 for my own convenience, but it doesn't strictly need to if that one gets mired.~~

## Try it out

On this branch, run `make dev` and visit the Pattern Library's [spinner pattern](http://localhost:4001/ui-playground/patterns-spinners) and [Spinner components](http://localhost:4001/ui-playground/components-spinner) pages.

Screen shot (doesn't show animation):

image

### DIFF
--- a/images/icons/spinner.svg
+++ b/images/icons/spinner.svg
@@ -1,0 +1,36 @@
+<!--
+By Sam Herbert (@sherb), for everyone. More @ http://goo.gl/7AJzbL
+
+From https://raw.githubusercontent.com/SamHerbert/SVG-Loaders/master/svg-loaders/tail-spin.svg
+-->
+<svg width="38" height="38" viewBox="0 0 38 38" xmlns="http://www.w3.org/2000/svg">
+    <defs>
+        <linearGradient x1="8.042%" y1="0%" x2="65.682%" y2="23.865%" id="a">
+            <stop stop-color="#222" stop-opacity="0" offset="0%"/>
+            <stop stop-color="#222" stop-opacity=".631" offset="63.146%"/>
+            <stop stop-color="#222" offset="100%"/>
+        </linearGradient>
+    </defs>
+    <g fill="none" fill-rule="evenodd">
+        <g transform="translate(1 1)">
+            <path d="M36 18c0-9.94-8.06-18-18-18" id="Oval-2" stroke="currentColor" stroke-width="2">
+                <animateTransform
+                    attributeName="transform"
+                    type="rotate"
+                    from="0 18 18"
+                    to="360 18 18"
+                    dur="0.9s"
+                    repeatCount="indefinite" />
+            </path>
+            <circle fill="#fff" cx="36" cy="18" r="1">
+                <animateTransform
+                    attributeName="transform"
+                    type="rotate"
+                    from="0 18 18"
+                    to="360 18 18"
+                    dur="0.9s"
+                    repeatCount="indefinite" />
+            </circle>
+        </g>
+    </g>
+</svg>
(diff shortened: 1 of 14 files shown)


styles/mixins/patterns/_spinners.scss
```diff
@@ -0,0 +1,76 @@
+// CSS Spinner modified from http://dabblet.com/gist/7615212
```
lyzadanger, line 1: Pretty much a direct c-n-p from the client.

src/components/Spinner.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * Loading indicator.
+ *
+ * @param {SpinnerProps} props
+ */
+export function Spinner({ classes = '' }) {
+  return (
+    <span className={classnames('Hyp-Spinner', classes)}>
```
LMS007, line 15: Currently LMS uses an SVG. I don't have a strong preference for either, but I think the svg was added after the css spinner in the client. We could discuss tomorrow and pick one or othe other and then replace as we move forward.

LMS007 (commented):
No objections to this PR other than deciding which is the right spinner to move forward with. LMS vs Client. The advantage of the SVG spinner is perhaps a whole lot less css, but it may be hard to tweak, slow down, etc.

lyzadanger:
@LMS007 Wow, I'm so glad you pointed the SVG implementation out from LMS! That is dead simple implementation; let me poke around and see if it can work for our needs. I tend to find the client one more visually attractive—that's why I used it here—but it was daft of me to not at least take a look at the `lms` variant. The implementation of that one is so simple that it may outweigh the slightly more attractive (but super complex) `client` version. I'm going to pop this PR into draft while I look into it.

lyzadanger:
OK! Reboot! Thank you @LMS007 for the heads up; I've opted for the _much_ simpler LMS variant here, which uses an animated SVG. While I'm still not 100% in love with the exact visual styling of the animated SVG per our own design style (p.s. I did take a look at other options in the originating [SVG repo](https://github.com/SamHerbert/SVG-Loaders) and this seemed to be one of the best alternatives). There's nothing wrong with the SVG! I just think the visual style of the (complexly-styled) version in the client is a little more in line with our design language.

The good news is that it's dead simple to drop in a different SVG in future if we find or make one that fits our style better! Another argument for centralizing this component...

For developer ergonomics, I added, in this latest implementation, a `size` property, as the loading spinner is often displayed large, in relation to the local font-size. I've added `large` and `xlarge` sizes because they're immediately useful. Whether we need to add `small` size(s) remains to be seen, but I didn't add them for now.

images/icons/spinner.svg
```diff
@@ -0,0 +1,36 @@
+<!--
```
lyzadanger, line 1: Copied from LMS

LMS007 (approved):
GTG! I think the SVG is just a bit more simple so that may be worth it alone.

The pattern lib is really getting flushed out, nice to see it grow.

LMS007:
> er ergonomics, I added, in this latest implementation

yep, I think the size prop is useful, thx

PR #140, opened 2021-07-09
Change `error` prop to `isError`

### LINKED ISSUE #136: Change name of `error` prop in `TextInput`
The `error` prop for `TextInput` is intended as a boolean; when `true`, it should indicate that something is wrong with the `input`. `TextInput` will apply error styling (presently a red outline; this may change) but has no logic about errors beyond that.

It has become our convention not to use `is` or `has` boolean prefixes on component properties that map directly to HTML attributes (e.g. `disabled`, `pressed`). However, `error` is not an HTML attribute on `input` elements, and its naming is ambiguous at present.

`isError` is a likely candidate for the rename.

This will be a breaking change. At time of writing this issue, `lms` is the only project using `TextInput`. It's the task of whoever picks this up to determine where `TextInput` is used, and whether it's plausible to take care of this as a breaking change, or whether some form of additive props/deprecation needs to happen. With luck, the former.

### LINKED ISSUE #138: `text-input.is-error` removal of border causing layout wiggle
The styling for the error state on the `text-input` pattern (used by the `TextInput` component) has a `border: none` rule that can cause a tiny layout wiggle for inputs that have a border in their non-error state (e.g. the `TextInputWithButton` component).

lyzadanger:
This PR contains two commits, both related to `TextInput`/`text-input`:

* The naming of the `error` prop on `TextInput` was ambiguous. ~~`isError`~~ `hasError` communicates
its boolean-ness better and doesn't masquerade as a native HTML element
attribute. It also better matches the applied stateful classname: ~~`.is-error`~~ `.has-error`
* Don't remove `border` when styling a text input in an error state. The error outline is inset, so should be independent of borders. Removing these borders here caused a tiny, tiny layout wiggle when `TextInput` was used on `lms`. Adds a pattern-library example that allows toggling between error and non-error states to demonstrate that there is no layout wiggle.

Screen grab of updated `TextInput` example in pattern library:

image

The prop name change is a _breaking change_. The only known use of this component is in one component in `lms`. I'll handle this manually, if that's OK (bump the package and update the component usage in one go).

Fixes #136
Fixes #138

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -7,7 +7,7 @@ import classnames from 'classnames';
  * @prop {string} [classes] - Additional CSS classes to apply
  * @prop {import('preact').Ref<HTMLInputElement>} [inputRef] - Optional ref for
  *   the rendered `input` element.
- * @prop {boolean} [isError] - There is an error associated with this input. Will
+ * @prop {boolean} [hasError] - There is an error associated with this input. Will
  *   set some error styling.
  */
 
@@ -32,12 +32,16 @@ import classnames from 'classnames';
 export function TextInput({
   classes = '',
   inputRef,
-  isError = false,
+  hasError = false,
   ...restProps
 }) {
   return (
     <input
-      className={classnames('Hyp-TextInput', { 'is-error': isError }, classes)}
+      className={classnames(
+        'Hyp-TextInput',
+        { 'has-error': hasError },
+        classes
+      )}
       {...restProps}
       ref={inputRef}
       type="text"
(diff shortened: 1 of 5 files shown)


src/components/TextInput.js
```diff
@@ -5,10 +5,10 @@ import classnames from 'classnames';
  *
  * @typedef TextInputBaseProps
  * @prop {string} [classes] - Additional CSS classes to apply
- * @prop {boolean} [error] - There is an error associated with this input. Will
- *   set some error styling.
  * @prop {import('preact').Ref<HTMLInputElement>} [inputRef] - Optional ref for
  *   the rendered `input` element.
+ * @prop {boolean} [isError] - There is an error associated with this input. Will
```
robertknight, line 10: Perhaps this should be `hasError` in the sense that "the input _has a validation error_" as opposed to "the input _is in an error state_"?

robertknight (commented):
I suggested that `hasError` might read better, since what the prop typically means is that the input has a validation error. Otherwise LGTM.

robertknight (approved):

PR #141, opened 2021-07-09
Remediate broken tests using `assert.exists`

### LINKED ISSUE #137: Vet component tests for erroneous use of `assert.exists`
Due to an oversight, some components have tests that use `assert.exists('some string')`, which _will always pass_. These need to be found and replaced with real tests.

lyzadanger:
This assertion does not do what I thought it did. Blush.

[`assert.exists`](https://www.chaijs.com/api/assert/#method_exists) asserts that something is neither `null` nor `undefined`. I don't even know how I got into a place of thinking these tests would do what I thought they would, which is checking to make sure a DOM element/React Wrapper corresponding to the given selector exists. Sheesh.

Fixes #137

### DIFF
--- a/src/components/test/Spinner-test.js
+++ b/src/components/test/Spinner-test.js
@@ -8,18 +8,18 @@ describe('Spinner', () => {
   const createSpinner = (props = {}) => mount(<Spinner {...props} />);
 
   it('renders', () => {
-    createSpinner();
-    assert.exists('span.Hyp-Spinner--medium');
+    const wrapper = createSpinner();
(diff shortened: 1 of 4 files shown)


LMS007 (approved):
Awe right! `wrapper.find` always returns an object so any assert.exists would always pass on that value.

In a few of these, it looks like it was just testing a string, but honestly if you didn't read it carefully, it looked like a valid test.  This is a perfect example of where tests don't help much because nothing tests the tests! At some point we just have to trust we right code right :)

In any case, good catch!

PR #142, opened 2021-07-09
Update structure of checkbox styling to match other patterns

### LINKED ISSUE #116: Adjust structure of `Checkbox` styling
The `Checkbox` component isn't quite in line with the emerging convention of this package, which is:

* Design pattern first, and the design pattern "owns" the style rules
* Component SASS `@include`s the design pattern's mixin(s)

There is no pattern for `Checkbox` yet, and presently the component defines all the rules in its own SASS module. Make it match the others.

lyzadanger:
This yawns-ville housekeeping PR adjusts the styling structure for checkboxes to match the way we're doing it for other patterns and components, which is, to wit:

* Underlying pattern mixins (`checkbox`, `label`, here)
* Pattern classes (`.hyp-checkbox`, `.hyp-label` here), which use pattern mixins
* Examples of patterns on a `*Patterns` pattern-library page
* Component classes, which use pattern mixins
* Example of components on `*Components` pattern-library page (were already present)

Previously, the styling implementation was in the component's SCSS module. In general, component SCSS modules should be dead simple and use the underlying pattern mixins as needed.

Diff here is a result of moving things around. The SCSS implementation here is not new. Checkbox pattern examples have been added to `FormPatterns`.

Fixes #116


styles/mixins/patterns/_forms.scss
```diff
@@ -88,3 +91,104 @@ $-color-text--disabled: var.$color-text--disabled;
...
+    }
+  }
+
+  // Set focus styles on SVG when checkbox is focused
+  &:focus {
+    & + span > svg {
+      @include focus.outline-rule;
+    }
+
+    // TODO polyfill/browser support
```
robertknight, line 170: Both the client and lms apps load the [standard polyfill](https://www.npmjs.com/package/focus-visible) for this pseudo-class, so I think we could write code in this package to assume that either the browser natively supports `:focus-visible` or this specific polyfill is loaded by the application consuming the package.

styles/mixins/patterns/_forms.scss
```diff
@@ -88,3 +91,104 @@ $-color-text--disabled: var.$color-text--disabled;
...
+  cursor: pointer;
+  // Style the checkbox such that it is not visible to sighted users but can
+  // be interacted with via screen-reader or touch-screen-reader.
+  position: absolute;
+  // Leaving a 1em size here allows users with touch-screen-readers to find
+  // and interact with the element.
+  width: 1em;
+  height: 1em;
+  // Some screen readers won't read out elements with 0 opacity
+  opacity: 0.00001;
```
robertknight, line 143: Good documentation here.

PR #144, opened 2021-07-14
Library layout components, simplified

lyzadanger:
This PR proposes a new collection of rather simple components for rendering Pattern Library pages: `Page`, `Pattern`, `Example` and `Demo`. These components represent an improvement on existing pattern-library components:

* The structure is simpler and more obvious
* The layout is responsive
* The styling is simpler, has no leakage into patterns, and uses and follows some existing underlying patterns
* Source JSX for demos is optional, and is put into a separate tab so as not to clutter the interface as much

This PR contains a component module with these components (`Library`), styles, and an example of one of the pattern-library pages converted to use the new components. 

This represents a coalescing of “lessons learned” from the first crop of pattern-library components. This was a quick implementation based on the growing set of underlying tools and patterns!

### Before

"Container components" page before these changes:

Screen Shot 2021-07-14 at 10 15 57 AM

Layout breaks on narrow screens:

Screen Shot 2021-07-14 at 10 17 11 AM

### After

Screen Shot 2021-07-14 at 10 14 05 AM

The layout is now responsive:

Screen Shot 2021-07-14 at 10 15 28 AM

### Proposed Model

In this model, a `Page` contains zero or more `Pattern` sections. A `Pattern` contains zero or more `Example`s. And an `Example` contains zero or more `Demo`s. 

`Demo` takes a `withSource` `boolean` prop. When `true`, it will render into a tabbed format. Here's what it looks like with the "Source" tab activated:

Screen Shot 2021-07-14 at 10 14 31 AM

When `false`, it will not provide source (second "Demo" here does not have source):

Screen Shot 2021-07-14 at 10 15 02 AM

All of these components are in a single module `Library`. Because the expectation is to import the set of components at once (e.g. `import * as Library from ‘../Library’;`) I’ve elected to leave the module name upper-case. Our convention for component modules that export multiple components has been to this point lower-case, however, this usage model is different. Discussion welcomed!

### Previous Model and shortcomings

(For reference if interested).

The previous model is `PatternPage`, `Pattern`, `PatternExamples`, `PatternExample`.

* `PatternExamples` uses a tabular layout, which is most often fine, but can cumbersome to manage in layout. And it’s not responsive.
* Some of the styling uses element selectors and/or relies on `base` selector styles. This can cause leakage into rendered patterns, and other complexity. The styling is also rather verbose and ad-hoc as it has developed over time.
* `PatternExample` will always render its source, which is sometimes not what you’d want. The “middle column” of an example row is kind of useless at present. Authors are forced to put example explanations into a `title` prop, which is awkward.
* `PatternExamples` doesn’t have much “meaning” as a component, and is used primarily to render a wrapping table. 

### Where to?

Assuming this kind of approach is reasonable, next steps would be to:

- [ ] Convert other pattern-library pages to use the Library components
- [ ] Make a package release providing access to the new Library components
- [ ] Update any third-party pattern-library pages (known one in the `client`, will have to verify that LMS does not have any) to use Library components
- [ ] Remove excess styling and unused `PatternPage` components


styles/pattern-library.scss
```diff
@@ -217,3 +219,81 @@ body {
     background-color: var.$color-grey-1;
   }
 }
+
+$-library-vertical-spacing: 7;
+.LibraryPage {
```
esanzgar, line 224: I wonder if `.LibraryPage` should belong to `styles/component/Library.css`

styles/pattern-library.scss
```diff
@@ -217,3 +219,81 @@ body {
...
+  // Make sure the content width is not too terribly wide: it gets hard to read
+  max-width: 75rem;
```
esanzgar, line 226: 👍🏼

esanzgar (approved):
I would suggest to export the components from `Library.js` in this way:

```
export default {
  Page,
  Pattern,
  Example,
  Demo,
};
```

So it can be imported without using the asterisk:
`import Library from './Library``

On a personal note, I prefer UIs with less borders and backgrounds, like this:

image

src/pattern-library/components/Library.js
```diff
@@ -0,0 +1,157 @@
...
+  const notDemos = kids.filter(kid => !demos.includes(kid));
+
+  return (
+    <div className="LibraryExample">
+      <div className="LibraryExample__container">
+        <div className="LibraryExample__content">
+          {title && <h3 className="LibraryExample__heading">{title}</h3>}
+          {notDemos}
+        </div>
+        <div className="LibraryExample__content">{demos}</div>
```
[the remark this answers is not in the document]
lyzadanger: Great catch!

src/pattern-library/components/Library.js
```diff
@@ -0,0 +1,157 @@
...
+      </div>
+      <div className="LibraryDemo__container">
+        {visibleTab === 'demo' && (
+          <div className="LibraryDemo__demo" style={style}>
+            <div className="LibraryDemo__demo-content">{children}</div>
+          </div>
+        )}
+        {visibleTab === 'source' && (
+          <div className="LibraryDemo__source">
+            <ul>{source}</ul>
```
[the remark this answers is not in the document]
lyzadanger: I believe that was the first implementation we tried. It broke down in some use cases, but darned if I didn't document which. I'll give it a brief think.

styles/pattern-library.scss
lyzadanger: I think it's worthwhile to consider some SASS module re-org after we get these new components used broadly and we're ready to remove the older support...I'll keep it in mind. I do want to keep pattern-library-related styling clearly separated from shared-component styling, however.

lyzadanger:
@esanzgar 

> On a personal note, I prefer UIs with less borders and backgrounds, like this:

I think that's helpful feedback. This particular page of patterns is a little hard to assess regarding "frames" because several of the components are frames themselves. Which leads to a sense of overwhelming frame-i-ness. If I might ask permission to take a look at this during the conversion of the next set of pages to this model? That way I can see how "demo" layout feels in real-world demo use cases and try to see if I can find a simpler style.

Issue #126, opened 2021-06-28
Figure out what's going on with `focus` and `button` styling


lyzadanger:
Although the button patterns and components use the `outline-on-keyboard-focus` focus mixin, they don't appear to have any focus rings at all in the pattern library. However, these same buttons _do_ have keyboard-focus outlines when used in the `client`:

![image](https://user-images.githubusercontent.com/439947/123652073-fc9e6680-d7f9-11eb-997f-a9c06efb4db2.png)

A little bit of sleuthing is needed here: why are these getting outlines when used in the `client` but not in the pattern-library pages?

PR #145, opened 2021-07-14
Make `:focus-visible` rules work when polyfill not present

lyzadanger:
When the [`focus-visible` polyfill](https://github.com/WICG/focus-visible) is loaded—which it is in our LMS and client projects—that polyfill will assign a `.focus-visible` class to elements whose heuristics match the intent of `:focus-visible`, whether or not the browser natively supports `:focus-visible` or not.

The following rule, as previously written, was meant to suppress outlines on elements if they didn't have the `.focus-visible` class (which is applied/removed by the polyfill):

```
&:focus:not(.focus-visible) {
  @include outline--hide;
}
```

The problem with this selector is that, if the polyfill is not present, it will always match—because the `.focus-visible` class is
never present—causing all focus outlines to be suppressed, even when the element is `:focus-visible`. That is, the previous selector was coupled to the presence of the polyfill.

When present, the polyfill also applies a `.js-focus-visible` class to a container element. We can make the selector here more specific, and include that classname, such that it will only match when the polyfill is actually present:

`.js-focus-visible &:focus:not(.focus-visible)`

Focus outlines "worked" before in the `lms` and `client` projects because the polyfill was also present, but did not work in this project, where we don't load the polyfill. Now this should work correctly in both contexts, and is independent of the polyfill.

Fixes https://github.com/hypothesis/frontend-shared/issues/126

### Test it out

Pull this branch, visit the [Buttons pattern-library page](http://localhost:4001/ui-playground/components-buttons) and try tabbing around between buttons. You should see a focus outline when you tab with a keyboard.

Compare against `main` where buttons don't show an outline.

### DIFF
--- a/styles/mixins/_focus.scss
+++ b/styles/mixins/_focus.scss
@@ -39,9 +39,13 @@ $-color-focus-outline: var.$color-focus-outline;
  */
 @mixin outline-on-keyboard-focus($inset: false) {
   @include outline($inset);
-  // Selector for browsers using JS polyfill, which adds the "focus-visible"
-  // class to a keyboard-focused element.
-  &:focus:not(.focus-visible) {
+
+  // For browsers using the https://github.com/WICG/focus-visible polyfill,
+  // add a selector for the classname assigned by that polyfill to indicate
+  // :focus-visible equivalence for browsers that don't natively support it.
+  // The `.js-focus-visible` class indicates the presence of the polyfill,
+  // and is assigned to a container element.
+  .js-focus-visible &:focus:not(.focus-visible) {
     @include outline--hide;
   }
 


styles/mixins/_focus.scss
```diff
@@ -39,9 +39,10 @@ $-color-focus-outline: var.$color-focus-outline;
  */
 @mixin outline-on-keyboard-focus($inset: false) {
   @include outline($inset);
+
   // Selector for browsers using JS polyfill, which adds the "focus-visible"
   // class to a keyboard-focused element.
-  &:focus:not(.focus-visible) {
+  .js-focus-visible &:focus:not(.focus-visible) {
```
LMS007, line 45: `.js-focus-visible`. Is a top level class correct? If thats the case, for clarify I think braces help here as well as putting the & on a second line. 

```
.js-focus-visible {
  &:focus:not(.focus-visible) {
    ...
  }
}

**Review by LMS007 — APPROVED** [review]
Yes, this makes more sense, the rule was too generic and I like this is a tiny tweak to fix it. I'm still wondering why Safari works without the polyfill, but I guess this means it has partial support for the pseudo selector. I wonder we can 100% ditch the polyfill then? Not a decision for today tho!

Great fix!

File: styles/mixins/_focus.scss [hunk 2]
```diff
@@ -39,9 +39,10 @@ $-color-focus-outline: var.$color-focus-outline;
  */
 @mixin outline-on-keyboard-focus($inset: false) {
   @include outline($inset);
+
   // Selector for browsers using JS polyfill, which adds the "focus-visible"
```
robertknight, line 43: I suggest we replace the phrase "JS polyfill" here with a link to the specific polyfill we assume is being used: https://github.com/WICG/focus-visible. The README for this project has an example which matches what you are doing here.

robertknight (approved):
Great PR description. The change looks good. I had a suggestion to link to the polyfill's GitHub project in the code comments.

styles/mixins/_focus.scss
lyzadanger: Great suggestion! Comment updated and expanded.

Issue #143, opened 2021-07-13
Provide ability to opt out of `Modal`/`Dialog` focus handling


lyzadanger:
`Dialog` and `Modal` (which wraps `Dialog`) have some handy focus-routing features that aid with a11y and usability. At present, the options are:

* Provide `Modal`/`Dialog` with an `initialFocus` property value, a `ref`. Assuming this `ref` has  a truthy `current` attribute, it will be focused when the `Modal` or `Dialog` is opened.
* Do nothing, or provide a `ref` with a non-truthy `current`. The `Modal`/`Dialog` will focus its container (`div`) for you.

There are some cases in which neither of these quite fit the bill, as in a case I've run into in which a given `Modal`'s contents themselves are built via several sub-components, making passing a `ref` to `Modal` cumbersome and fractious. In my case, I want to manage focus on a text input that is a couple of levels nested, and only focus it in some cases.

To provide for use cases like this, it would be nice to have a way to "opt out" of any focus handling at the `Modal`/`Dialog` level and allow manual control. 

Related to https://github.com/hypothesis/lms/issues/2954

PR #146, opened 2021-07-15
Add ability to opt out of focus control for `Dialog`, `Modal`

lyzadanger:
In some cases, it's useful to be able to opt out of the automatic focus handling of `Dialog` and `Modal` components. From https://github.com/hypothesis/frontend-shared/issues/143 :

> There are some cases in which neither of these quite fit the bill, as in a case I've run into in which a given Modal's contents themselves are built via several sub-components, making passing a ref to Modal cumbersome and fractious. In my case, I want to manage focus on a text input that is a couple of levels nested, and only focus it in some cases.

> To provide for use cases like this, it would be nice to have a way to "opt out" of any focus handling at the Modal/Dialog level and allow manual control.

This PR introduces the ability to set the `initialFocus` ref to `null` to opt out of automatic focus control, and allow for manual focus control.

There are three commits on this branch:

1. Convert the dialog-components pattern-library page to use the new Library components — "boilerplate" conversion
2. Update `Dialog` to handle `null` for `initialFocus`
3. Add example of manual focus control to the dialog-components pattern-library page

Fixes https://github.com/hypothesis/frontend-shared/issues/143

Screenshot of [updated Dialogs component page](http://localhost:4001/ui-playground/components-dialogs) after these changes:

image

### DIFF
--- a/src/components/test/Dialog-test.js
+++ b/src/components/test/Dialog-test.js
@@ -118,18 +118,16 @@ describe('Dialog', () => {
       );
     });
 
-    it('focuses the dialog if `initialFocus` ref is `null`', () => {
-      const wrapper = mount(
-        <Dialog initialFocus={{ current: null }} title="My dialog">
+    it('does not set focus if `initialFocus` is set to `null`', () => {
+      const focusedBefore = document.activeElement;
+      mount(
+        <Dialog initialFocus={null} title="My dialog">
           <div>Test</div>
         </Dialog>,
         { attachTo: container }
       );
 
-      assert.equal(
-        document.activeElement,
-        wrapper.find('[role="dialog"]').getDOMNode()
-      );
+      assert.equal(document.activeElement, focusedBefore);
     });
 
     it('focuses the dialog if `initialFocus` element is disabled', () => {
--- a/src/pattern-library/components/patterns/DialogComponents.js
+++ b/src/pattern-library/components/patterns/DialogComponents.js
@@ -1,6 +1,14 @@
 import { createRef, render } from 'preact';
 import { useState } from 'preact/hooks';
-import { ConfirmModal, Dialog, LabeledButton, Modal } from '../../../';
+import {
+  ConfirmModal,
+  Dialog,
+  LabeledButton,
+  Modal,
+  TextInputWithButton,
+  TextInput,
+  IconButton,
+} from '../../../';
 
 import Library from '../Library';
 
@@ -72,6 +80,8 @@ export default function DialogComponents() {
   const [dialogIsOpen, setDialogIsOpen] = useState(false);
   // Basic Modal example
   const [, setModalIsOpen] = useState(false);
+  // Modal with manual focus control
+  const [, setFocusModalIsOpen] = useState(false);
   // Modal with overflowing content example
   const [, setLongModalIsOpen] = useState(false);
   // ConfirmModal example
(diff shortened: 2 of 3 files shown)


src/pattern-library/components/patterns/DialogComponents.js
```diff
@@ -66,6 +74,19 @@ const showDialog = ({ DialogComponent, container, setOpen, props }) => {
 };
 
 export default function DialogComponents() {
+  // Dialog/Modal state for each of the examples
```
lyzadanger, line 77: I concede that the code to handle dialog demos on this pattern-library page is long and wander-y. I considered trying to restructure this and/or extract some into helpers, but felt overwhelmed, so I'll come back to it later.

src/components/test/Dialog-test.js
```diff
@@ -118,18 +118,15 @@ describe('Dialog', () => {
...
           <div>Test</div>
         </Dialog>,
         { attachTo: container }
       );
 
-      assert.equal(
-        document.activeElement,
-        wrapper.find('[role="dialog"]').getDOMNode()
-      );
+      assert.equal(document.activeElement, document.body);
```
robertknight, line 129: This assumes that `document.activeElement` was `document.body` prior to running the test. A safer approach would be to capture the value before mounting and check that it doesn't change.

src/pattern-library/components/patterns/DialogComponents.js
```diff
@@ -66,6 +74,19 @@ const showDialog = ({ DialogComponent, container, setOpen, props }) => {
...
+  // Basic Dialog example
+  const [dialogIsOpen, setDialogIsOpen] = useState(false);
+  // Basic Modal example
+  const [, setModalIsOpen] = useState(false);
+  // Modal with manual focus control
+  const [, setFocusModalIsOpen] = useState(false);
+  // Modal with overflowing content example
+  const [, setLongModalIsOpen] = useState(false);
+  // ConfirmModal example
+  const [, setConfirmModalIsOpen] = useState(false);
```
robertknight, line 88: What is the rationale for these write-only state values? Is it just to provide a dummy callback to pass to `showDialog`'s `setOpen` prop? If so you could make that optional instead.

robertknight (commented):
The functional changes to `Dialog` look good. I spotted a possible hazard in the tests that could cause flakiness and also had a query about the use of write-only state variables in the pattern library.

src/pattern-library/components/patterns/DialogComponents.js
lyzadanger: I mentioned [in another comment](https://github.com/hypothesis/frontend-shared/pull/146#discussion_r670539602) that this needs some attention (i.e. these demos need refactor). I'd like to tackle this separately so I don't get too mired.

src/components/test/Dialog-test.js
lyzadanger: Good catch. Will update.

src/pattern-library/components/patterns/DialogComponents.js
robertknight: OK, but I didn't grok the thinking behind this code. The only scenario in which write-only state is useful that I can see is if you want to force a re-render even though nothing the component directly depends on changed.

robertknight (approved):

lyzadanger:
Here's a follow-up issue for the convoluted examples code: https://github.com/hypothesis/frontend-shared/issues/148

PR #149, opened 2021-07-19
Bump package to v3.6.0

lyzadanger:
Note that there was a fumble in the `3.5.0` release, which was correctly tagged, but the `CHANGELOG` and `package.json` were not correctly updated. Currently trying to debug how that happened, and what I can put in place to avoid it happening again. I try to be careful when juggling version tags and commit hashes and whatnot.


lyzadanger:
Per team convention for minor releases, I am going to merge this without review.

PR #150, opened 2021-07-20
Fix icon naming for Spinner component

lyzadanger:
Derp/facepalm territory. The `Spinner` component indeed registered its needed icon but then used `spinner` in its rendered output. This was a bit of a trainwreck; what happened was:

* This project already had a registered `spinner` icon pre-existing the addition of the `spinner` pattern and `Spinner` component, so `spinner` was present/registered in this project, quietly lurking
* There was a faulty test here on `Spinner` that checked for `spinner` instead of `hyp-spinner`
* LMS—the first project this component was intended for—also registers a `spinner` icon, so this wasn't caught
* Today I tried starting to replace `Spinner` components in the client, and whammo. No `spinner` icon in that project.

Also discovered a silly test in the `Thumbnail` component which broke when I changed the icon naming.

Follow-up issue forthcoming to not register or use any non-prefixed (`hyp-`) icons in the pattern library to avoid this conflict in the future.

### DIFF
--- a/src/components/test/Spinner-test.js
+++ b/src/components/test/Spinner-test.js
@@ -12,6 +12,11 @@ describe('Spinner', () => {
     assert.isTrue(wrapper.find('SvgIcon.Hyp-Spinner--medium').exists());
   });
 
+  it('uses the registered `hyp-spinner` icon', () => {
+    const wrapper = createSpinner();
+    assert.equal(wrapper.find('SvgIcon').props().name, 'hyp-spinner');
+  });
+
   it('applies additional classes', () => {
     const wrapper = createSpinner({ classes: 'foo bar' });
     assert.isTrue(wrapper.find('SvgIcon.Hyp-Spinner--medium.foo.bar').exists());
--- a/src/components/test/Thumbnail-test.js
+++ b/src/components/test/Thumbnail-test.js
@@ -28,7 +28,7 @@ describe('Thumbnail', () => {
   context('when in loading state', () => {
     it('renders a loading spinner', () => {
       const wrapper = createComponent({ isLoading: true });
-      assert.isTrue(wrapper.find('SvgIcon[name="spinner"]').exists());
+      assert.isTrue(wrapper.find('Spinner').exists());
     });
 
     it('does not render content', () => {
--- a/src/components/Spinner.js
+++ b/src/components/Spinner.js
@@ -22,5 +22,7 @@ registerIcons({
  */
 export function Spinner({ classes = '', size = 'medium' }) {
   const baseClass = `Hyp-Spinner--${size}`;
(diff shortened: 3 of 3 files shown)


src/components/test/Spinner-test.js
```diff
@@ -12,6 +12,11 @@ describe('Spinner', () => {
     assert.isTrue(wrapper.find('SvgIcon.Hyp-Spinner--medium').exists());
   });
 
+  it('uses the registered `hyp-spinner` icon', () => {
+    const wrapper = createSpinner();
+    assert.equal(wrapper.find('SvgIcon').props().name, 'hyp-spinner');
+  });
+
```
lyzadanger, line 19: Work tracked for follow up: Don't register any icons as part of test bootstrapping; tests should fail if a component references an icon it doesn't register itself, even if a test is not written for the icon itself (i.e. a component will fail to render in test because the icon will be missing).

src/components/test/Thumbnail-test.js
```diff
@@ -28,7 +28,7 @@ describe('Thumbnail', () => {
   context('when in loading state', () => {
     it('renders a loading spinner', () => {
       const wrapper = createComponent({ isLoading: true });
-      assert.isTrue(wrapper.find('SvgIcon[name="spinner"]').exists());
+      assert.isTrue(wrapper.find('Spinner').exists());
```
lyzadanger, line 31: Yeah, I can't make a good excuse as to why `Thumbnail` was testing for the presence of an icon instead of the Spinner component. It may have been that I implemented `Thumbnail` before `Spinner` was a component (i.e. it existed as a pattern but hadn't been component-ized yet).

src/components/test/Spinner-test.js
```diff
@@ -12,6 +12,11 @@ describe('Spinner', () => {
     assert.isTrue(wrapper.find('SvgIcon.Hyp-Spinner--medium').exists());
   });
 
+  it('uses the registered `hyp-spinner` icon', () => {
+    const wrapper = createSpinner();
+    assert.equal(wrapper.find('SvgIcon').props().name, 'hyp-spinner');
+  });
```
robertknight, line 18: We don't typically write manual tests for static markup. In this case you're trying to enforce that components only depend on internally registered icons rather than ones that are exported from the package but not registered internally. That's something that really needs to be solved in a different way.

robertknight (commented):
The code change in `Spinner.js` is fine, but I don't think the unit test is useful as it is just testing static markup. Guaranteeing that components in this package only use icons that they register is something that needs to be done a different way. It seems we both have some ideas around that. We can talk through approaches next week. The immediate change in this PR can be landed independent of that though.

PR #152, opened 2021-07-20
Prefix all project icon names with `hyp-`

### LINKED ISSUE #151: All icons in this project should have `hyp-` prefix
To avoid confusion, all icons in this package should be registered with a `hyp-` prefix. This eliminates collisions with registered icons in other projects and can help prevent situations like https://github.com/hypothesis/frontend-shared/pull/150 from arising.

This will require updates to the main `icons` module, as well as usage in pattern-library components.

lyzadanger:
This PR prefixes the names of all icons used in this project with `hyp-` to avoid namespace collisions with other projects, and ambiguity. These are boilerplate/grep changes to update name references throughout the pattern library.

It also updates test bootstrapping such that it doesn't register the icons in the `icons` module. Instead, it registers a single "test" icon for convenience, and leaves it up to components to register the icons they require.

The icons registered in the project's main `icons` module are intended for use within pattern-library examples (comment added to that effect). They should not ever be relied upon for any public component.  Components must register any icons they use (using the `hyp-` prefix). This last part was already the convention in play, but the changes in this PR should help prevent pilot error in this regard.

There are no changes here to anything publicly available (components, e.g.). This will not affect users of the package, but is intended to prevent mess-ups like #150 from happening in the first place.

Depends on #150 because of the spinner-icon rename in that branch

Fixes #151

### DIFF
--- a/src/components/test/Dialog-test.js
+++ b/src/components/test/Dialog-test.js
@@ -44,7 +44,7 @@ describe('Dialog', () => {
   });
 
   it('renders an icon', () => {
-    const wrapper = mount(<Dialog title="Test dialog" icon="edit" />);
+    const wrapper = mount(<Dialog title="Test dialog" icon="hyp-edit" />);
     const icon = wrapper.find('SvgIcon[data-testid="header-icon"]');
     assert.isTrue(icon.exists());
   });
--- a/src/components/test/Dialog-test.js
+++ b/src/components/test/Dialog-test.js
@@ -44,7 +44,7 @@ describe('Dialog', () => {
   });
 
   it('renders an icon', () => {
-    const wrapper = mount(<Dialog title="Test dialog" icon="hyp-edit" />);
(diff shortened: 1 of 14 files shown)


robertknight:
I had a quick initial look over this. Using a common prefix for names of icons in this package is one way that we could avoid conflicts. An alternative approach might involve revisiting the way icon registration works to avoid the general issue with a global namespace. Let's catch up next week and discuss some options.

lyzadanger:
> I had a quick initial look over this. Using a common prefix for names of icons in this package is one way that we could avoid conflicts. An alternative approach might involve revisiting the way icon registration works to avoid the general issue with a global namespace. 

Would it be possible to land this as an interim solution so I can proceed with several PRs that are based off of this work? This change doesn't affect consumers of the package, so doing this won't _hurt_ anything.

lyzadanger:
Sounds like we're probably going to take another approach. Closing this.

PR #155, opened 2021-07-26
Remove remaining prop-types usage

robertknight:
We stopped using prop-types some time ago in favor of TypeScript + JSDoc, but it was still used by the `SvgIcon` component and `mockImportedComponents` utility in this package.

1. Remove `prop-types` usage from `SvgIcon` and remove the dependency from package.json
2. Update `mockImportedComponents` to latest version from client repo, that no longer depends on the `propTypes` property

### DIFF
--- a/src/hooks/test/use-element-should-close-test.js
+++ b/src/hooks/test/use-element-should-close-test.js
@@ -21,7 +21,6 @@ describe('useElementShouldClose', () => {
   ];
 
   // Create a fake component to mount in tests that uses the hook
-  // eslint-disable-next-line react/prop-types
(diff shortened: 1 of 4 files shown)


src/hooks/test/use-element-should-close-test.js
```diff
@@ -21,7 +21,6 @@ describe('useElementShouldClose', () => {
   ];
 
   // Create a fake component to mount in tests that uses the hook
-  // eslint-disable-next-line react/prop-types
```
lyzadanger, line 24: Yay

lyzadanger (approved):
Thanks!

PR #163, opened 2021-08-04
Revise icon registration API to prevent conflicts

robertknight:
Revise the API used to register icons with `SvgIcon` to prevent conflicts between different packages using the component.

Previously the keys of registered icons were developer-chosen strings. This could lead to conflicts if different packages tried to use the same string for different icons. The new approach in this commit is for the developer to call a `registerIcon(name, svgMarkup)` function which registers an icon and returns a key for use with `<SvgIcon name={key}
..>`. `registerIcon(...)` creates a new symbol and uses it as the key for the map entry, then returns the key. It could just as well generate some other type of unique key, but a symbol seems the natural choice for this purpose.

An added bonus of the new approach is that it makes it easier to find all the places that use a particular icon, since you can just ask your editor / IDE for all references to the variable holding the key returned by `registerIcon(...)`.

Previous API:

```js
registerIcons({ 'cancel': require('path/to/icon.svg') })

<SvgIcon name="cancel" .../>
```

New API:

```js
const cancelIcon = registerIcon('cancel', require('path/to/icon.svg'))

<SvgIcon name={cancelIcon} .../>
```

_In the second API `cancelIcon` holds the unique key and the "cancel" string passed as the first argument to `registerIcon` is just descriptive for use in error messages etc. If two different components registered different icons using the same descriptive name, they won't conflict._

To enable gradual migration, the `registerIcons` function and support for string values for the `name` prop is maintained, but deprecated.

### DIFF
--- a/src/components/SvgIcon.js
+++ b/src/components/SvgIcon.js
@@ -4,7 +4,7 @@ import { useLayoutEffect, useRef } from 'preact/hooks';
 /**
  * Object mapping icon names to SVG markup.
  *
- * @typedef {Object.<string,string>} IconMap
+ * @typedef {Map<string|symbol, string>} IconMap
  */
 
 /**
@@ -17,13 +17,13 @@ import { useLayoutEffect, useRef } from 'preact/hooks';
  *
  * @type {IconMap}
  */
-let iconRegistry = {};
+const iconRegistry = new Map();
 
 /**
  * @typedef SvgIconProps
- * @prop {string} name - The name of the icon to display.
+ * @prop {string|symbol} name - The name of the icon to display.
  *   The name must match a name that has already been registered using the
- *   `registerIcons` function.
+ *   `registerIcon` or `registerIcons` functions.
  * @prop {string} [className] - A CSS class to apply to the `<svg>` element.
  * @prop {boolean} [inline] - Apply a style allowing for inline display of icon wrapper.
  * @prop {string} [title] - Optional title attribute to apply to the SVG's containing `span`.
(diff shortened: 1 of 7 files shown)


robertknight:
A question you might ask reading this is "why do we have to register icons at all?" Why can't we just do:

```js
const cancelIcon = require('path/to/icon.svg');

<SvgIcon svg={cancelIcon} .../>
```

The primary reason is to ensure that only markup from trusted sources gets rendered. The thinking is that it is easier to check that `registerIcon` is always called with markup from a safe source than to check all props passed to `<SvgIcon>`. The other reason is that having a short, descriptive name associated with the icon is useful for debugging, tests etc.

lyzadanger:
Right now, many modules in our applications reference icons that are registered in other modules. For example, in this project, the pattern-library pages make use of icons that have been registered "by the application" during bootstrap, using (proposed-to-be-deprecated) string references. Is the intent that, in the future, any module that wishes to use icons would need to register them directly? That is, how do you envision the future of app-wide "icon registries?" Do they go away? Are there efficiency issues with re-registering the same markup in multiple modules (e.g. three modules use the "edit" icon, so they each register it?)? Or were you envisioning some way that registered-icon Symbols/keys could be shared across modules in an app?

It's a bit loosey-goosey that modules currently reference icons that they don't register (can lead to bugs), so I'm not opposed to modules having to register or somehow "declare" their intent to use an icon. But I do want to be sure it's relatively ergonomic and not an unreasonable performance hit.

lyzadanger:
> The primary reason is to ensure that only markup from trusted sources gets rendered. The thinking is that it is easier to check that registerIcon is always called with markup from a safe source than to check all uses of <SvgIcon>. The other reason is that having a short, descriptive name associated with the icon is useful for debugging, tests etc.

We should add some commenting to this effect when we next touch this component. I think it's really helpful context for the Next Dev.

src/components/SvgIcon.js
```diff
@@ -17,13 +17,13 @@ import { useLayoutEffect, useRef } from 'preact/hooks';
  *
  * @type {IconMap}
  */
-let iconRegistry = {};
+const iconRegistry = new Map();
 
 /**
  * @typedef SvgIconProps
- * @prop {string} name - The name of the icon to display.
+ * @prop {string|symbol} name - The name of the icon to display.
```
esanzgar, line 24: Do you support `string` to transition from the old `registerIcons` to the new `registerIcon`?

PR #165, opened 2021-08-05
Do not hard-code `localhost` in web template asset paths

lyzadanger:
Use relative links so that the Pattern Library web site may be accessed by devices on the same network.

Before, I could access the landing page of the Pattern Library from another device on my home network, but styles were not accessible, as they were referenced with a full `localhost` path. Pages in the Pattern Library were not accessible because the script path was also hard-coded to `localhost`.

After these changes, sweet bowls of cherries.


robertknight (approved):
LGTM.

PR #161, opened 2021-08-03
Convert container-patterns page to newer Library components

lyzadanger:
Part of overall conversion to newer components. Convert container patterns page to newer Library components for clarity, responsive layout.

After:

image

Part of #162

PR #166, opened 2021-08-05
Add "scrollbox" pattern

lyzadanger:
This PR introduces a "scrollbox" container pattern. 

A scrollbox is a scrolling "frame" that shows scroll-hint shadows at the top and bottomof the frame if:
 *  The content height exceeds the frame height: i.e. it is overflowing and can be scrolled, and
 *  The content is scrollable in the shadow's direction (up or down)

This is a CSS-only implementation of hinting behavior to help users see when content is overflowing its container and can be scrolled (thanks, browsers, for hiding scrollbars, is awesome).

This pattern will help with overflowing Table content, as it is not always obvious when said content is vertically scrollable otherwise. This pattern has been tested in Firefox, Chrome and Safari (desktop and iOS) (_updated_: this has also been tested in Opera). Safari's behavior is different than Chrome and Firefox's, as shown and explained below and in code comments. It relies on the `background-attachment` CSS rule, which, except for bugginess in Safari (with the `local` value), has had support in all major browsers for quite some time: https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment

Example of `scrollbox` with overflowing content (Chrome, Firefox, Opera behavior):

overflow-list-example

Safari has a bug with `background-attachment: local` detailed in code comments. Its behavior is different, but I think acceptable: if content is scrollable, the bottom shadow will always show. Top shadows never show. As in other browsers, the bottom shadow will only show if the content overflows (so we don't end up with meaningless shadows):

overflow-safari-behavior

Finally, here is an example of a `table`-pattern prototype in a `scrollbox`, to demonstrate the intended application of this pattern:

table-prototype-with-scrollbox

The CSS to achieve this is moderately arcane, but I've attempted to comment thoroughly and link to resources.

Depends on https://github.com/hypothesis/frontend-shared/pull/161

### DIFF
--- a/src/pattern-library/components/patterns/DialogComponents.js
+++ b/src/pattern-library/components/patterns/DialogComponents.js
@@ -160,7 +160,7 @@ export default function DialogComponents() {
 
   const openLongModal = () => {
     const children = (
-      <div style={{ overflow: 'auto' }}>
+      <div className="hyp-scrollbox">
         <p>
           Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis a sapien
           cursus, fringilla diam posuere, varius urna. Phasellus dictum sodales
--- a/styles/mixins/patterns/_containers.scss
+++ b/styles/mixins/patterns/_containers.scss
@@ -149,3 +149,65 @@ $-color-background: var.$color-background;
     }
   }
 }
+
+/**
+ * A scrolling "frame" that shows scroll-hint shadows at the top and bottom
+ * of the frame if:
+ *  - The content height exceeds the frame height: i.e. can be scrolled, and
+ *  - The content is scrollable in the shadow's direction (up or down)
+ *
+ * Shadows are not visible once the frame has been scrolled all the way in the
+ * shadow's direction. Shadows are not visible if the height of the content
+ * does not overflow the frame (is not scrollable).
+ *
+ * The shadow hinting is handled by four positioned background gradients:
+ *   - One gradient each at top and bottom of frame that obscure the shadow hints
+ *     (shadow covers). These use `background-attachment: local`, which makes
+ *     their position fixed to the _content_ within the scrollbox.
+ *   - One gradient each at the top and the bottom of the frame that are the
+ *     shadow hints (shadows). These use `background-attachment: scroll` such
+ *     that they are always positioned at the top and the bottom of the
+ *     _scrollbox_ frame. When these positions align with the positions of the
+ *     shadow covers--at the top and the bottom of the overflowing content--
+ *     they will be obscured by those shadow covers.
+ *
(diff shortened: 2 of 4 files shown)


styles/mixins/patterns/_containers.scss
```diff
@@ -149,3 +150,66 @@ $-color-background: var.$color-background;
...
+ *   - Only the bottom scroll-hint shadow appears if content overflows
+ *   - The bottom scroll-hint shadow is always present, even if content is
+ *     fully scrolled
+ */
+@mixin scrollbox {
+  overflow: auto;
+  @include atoms.border;
+
+  background:
+		// Shadow covers
```
lyzadanger, line 192: Weird indent, will fix.

src/pattern-library/components/patterns/DialogComponents.js
```diff
@@ -160,7 +160,7 @@ export default function DialogComponents() {
 
   const openLongModal = () => {
     const children = (
-      <div style={{ overflow: 'auto' }}>
+      <div className="hyp-scrollbox">
```
lyzadanger, line 163: This demonstrates that this pattern can be used within a Modal to constrain to the modal dimensions (without setting explicit dimensions).

styles/mixins/patterns/_containers.scss
```diff
@@ -1,6 +1,7 @@
 @use '../../variables' as var;
 
 @use '../atoms';
+@use '../focus';
```
lyzadanger, line 4: Oops. Not needed; will remove.

styles/mixins/patterns/_containers.scss
```diff
@@ -149,3 +149,65 @@ $-color-background: var.$color-background;
...
+        rgba(0, 0, 0, 0.05) 5px,
+        rgba(255, 255, 255, 0) 70%
+      )
+      0 100%;
+  background-repeat: no-repeat;
+  background-color: $-color-background;
+  background-size: 100% 40px, 100% 40px, 100% 14px, 100% 14px;
+
+  background-attachment: local, local, scroll, scroll;
+}
```
robertknight, line 213: For the effect to work, if I understand correctly, the shadow covers background layer need to be rendered above the shadows. I was surprised when reading this because the shadow covers layer was declared first, and in the DOM later elements are stacked above earlier ones by default. Apparently [CSS backgrounds](https://css-tricks.com/stacking-order-of-multiple-backgrounds/) don't work that way though.

robertknight (approved):
LGTM. Great documentation 👍

styles/mixins/patterns/_containers.scss
lyzadanger: Yes, me too! It seemed backwards.

Issue #129, opened 2021-06-29
Add typing for "base" props for certain types of components


lyzadanger:
As we build out the collection of shared components, we're running into some component "types":

* Logic-free "container" components. These components are simple and serve to apply the correct design pattern via styling. These components always take `children` and `classes:string` props.
* Components that wrap single HTML elements, e.g. `*Button`, `TextInput`. These components should always take a `ref`, and should also take `classes` to give developers flexibility.

At some point we should centralize base-props types for these types of components.

lyzadanger:
Another piece of this: common test suites for components that take these common props.

lyzadanger:
This is really part of https://github.com/hypothesis/frontend-shared/issues/177

PR #170, opened 2021-08-10
Add Scrollbox container component

lyzadanger:
This PR adds a `Scrollbox` container component based on the `scrollbox` pattern. It also extracts a partial pattern for a `sticky-header`, as inspired by [this comment](https://github.com/hypothesis/frontend-shared/pull/168#discussion_r685978251) in a previous PR. The only place a sticky header is currently used in our applications is in a table, so I haven't fully fleshed out this pattern, but it exists as a stub and the table-heading styling makes use of it now.

This `Scrollbox` component will be used by forthcoming `Table` component.

You can see this component in action on the [Container Components](http://localhost:4001/ui-playground/components-containers) pattern-library page.

Depends on https://github.com/hypothesis/frontend-shared/pull/169 as it builds on a pattern-library page modified in that branch.

image

### DIFF
--- a/src/components/containers.js
+++ b/src/components/containers.js
@@ -47,3 +47,20 @@ export function Actions({ children, direction = 'row', classes = '' }) {
   const baseClass = `Hyp-Actions--${direction}`;
   return <div className={classnames(baseClass, classes)}>{children}</div>;
 }
+
+/**
+ *
+ * @typedef ScrollboxBaseProps
+ * @prop {boolean} [withHeader=false] - Provide layout affordances for a sticky
+ *   header in the scrollable content
+ */
+
+/**
+ * Render a scrollable container to contain content that might overflow.
+ *
+ * @param {ScrollboxBaseProps & ContainerProps} props
+ */
+export function Scrollbox({ children, classes = '', withHeader = false }) {
+  const baseClass = withHeader ? 'Hyp-Scrollbox--with-header' : 'Hyp-Scrollbox';
+  return <div className={classnames(baseClass, classes)}>{children}</div>;
+}
--- a/src/components/test/containers-test.js
+++ b/src/components/test/containers-test.js
@@ -1,6 +1,6 @@
 import { mount } from 'enzyme';
 
-import { Frame, Card, Actions } from '../containers';
+import { Frame, Card, Actions, Scrollbox } from '../containers';
 
 describe('Frame', () => {
   const createComponent = (props = {}) =>
@@ -68,5 +68,32 @@ describe('Actions', () => {
     const wrapper = createComponent({ direction: 'column' });
 
     assert.isTrue(wrapper.find('div.Hyp-Actions--column').exists());
+  });
+});
+
+describe('Scrollbox', () => {
+  const createComponent = (props = {}) =>
+    mount(
+      <Scrollbox {...props}>
+        <div>This is content inside of a Scrollbox</div>
+      </Scrollbox>
+    );
+
+  it('renders children inside of a div with appropriate classnames', () => {
+    const wrapper = createComponent();
+
+    assert.isTrue(wrapper.find('.Hyp-Scrollbox').exists());
+  });
+
+  it('applies extra classes', () => {
+    const wrapper = createComponent({ classes: 'foo bar' });
+
+    assert.isTrue(wrapper.find('div.Hyp-Scrollbox.foo.bar').exists());
+  });
+
+  it('applies header-affordance layout class if `withHeader`', () => {
+    const wrapper = createComponent({ withHeader: true });
+
+    assert.isTrue(wrapper.find('div.Hyp-Scrollbox--with-header').exists());
   });
 });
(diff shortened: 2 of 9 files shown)


src/components/containers.js
```diff
@@ -47,3 +47,20 @@ export function Actions({ children, direction = 'row', classes = '' }) {
...
+ * @prop {boolean} [withHeader=false] - Provide layout affordances for a sticky
+ *   header in the scrollable content
+ */
+
+/**
+ * Render a scrollable container to contain content that might overflow.
+ *
+ * @param {ScrollboxBaseProps & ContainerProps} props
+ */
+export function Scrollbox({ children, classes = '', withHeader = false }) {
```
esanzgar, line 63: The classes in these two components will be ignored, isn't it? `<Scrollbox className="myClass" />` or `<Scrollbox class="myClass" />`

Wouldn't be better to support and extend the native `class` and `className` props?

src/components/test/containers-test.js
```diff
@@ -70,3 +70,30 @@ describe('Actions', () => {
...
+    mount(
+      <Scrollbox {...props}>
+        <div>This is content inside of a Scrollbox</div>
+      </Scrollbox>
+    );
+
+  it('renders children inside of a div with appropriate classnames', () => {
+    const wrapper = createComponent();
+
+    assert.isTrue(wrapper.find('div').first().hasClass('Hyp-Scrollbox'));
```
esanzgar, line 85:
```suggestion
    assert.isTrue(wrapper.find('div.Hyp-Scrollbox').exists());
```

src/components/test/containers-test.js
```diff
@@ -70,3 +70,30 @@ describe('Actions', () => {
...
+  });
+
+  it('applies extra classes', () => {
+    const wrapper = createComponent({ classes: 'foo bar' });
+
+    assert.isTrue(wrapper.find('div.Hyp-Scrollbox.foo.bar').exists());
```
esanzgar, line 91: If you want to check the correct class order, you can use this complicated test:
```suggestion
    assert.deepEqual(
      [
        ...wrapper
          .find('div.Hyp-Scrollbox.foo.bar')
          .getDOMNode()
          .classList.values(),
      ],
      ['Hyp-Scrollbox', 'foo', 'bar' ]
    );
```

src/components/containers.js
esanzgar: I see that other components also uses `classes`.

styles/mixins/patterns/_containers.scss
```diff
@@ -235,3 +235,26 @@ $-header-height: var.$touch-target-size;
 @mixin scrollbox--with-header {
   @include scrollbox($shadow-top-position: $-header-height);
 }
+
+/**
+ * A sticky container that is sized one-touch-unit high. This is currently
+ * expanded upon by table styling, but not used in other patterns (yet).
+ */
+@mixin sticky-header {
```
esanzgar, line 243:
```suggestion
@mixin sticky-scrollbox-header {
```

The association of the two names could help to understand when to use the pattern.

styles/mixins/patterns/_containers.scss
```diff
@@ -235,3 +235,26 @@ $-header-height: var.$touch-target-size;
...
+  position: sticky;
+  top: 0;
+  height: $-header-height;
+
+  @include layout.padding;
+  @include atoms.border(bottom);
+
+  background-color: var.$color-grey-1;
```
esanzgar, line 251: Maybe the colour could be taken out of this pattern, so the pattern only deals about positioning and size.

esanzgar (approved):
I added a few comments.

I would suggest to add use the new `<Scrollbox>` component in the `LongModalExample` (`DialogComponents.js`):

```
        <Scrollbox>
          <LoremIpsum />
        </Scrollbox>
```

src/components/containers.js
lyzadanger: There is an API emerging for the components in this package. There is an issue that sort of captures the need to formalize it: https://github.com/hypothesis/frontend-shared/issues/129

I think we're moving towards all (or nearly all, still needs assessment) components accepting a few common props. The most evident one is `classes`, which is a string of any classes to append to the outermost element rendered by the component. I chose the name `classes` for some coherent reason, which I am having trouble tracking down now! The train has somewhat left the station on this (or, at least, will be a breaking change if we decide to use `className` instead). I'll see if I can't track down the origin of this naming convention; I swear I didn't make it up.

Another common prop taken by non-container components is a `ref`, and the naming convention we settled on was `<elementName>Ref`, e.g. `inputRef`, `buttonRef`.

Anyway, this definitely needs some clarity/formalization....

src/components/test/containers-test.js
lyzadanger: I've added a note to https://github.com/hypothesis/frontend-shared/issues/129 that could help to address this: I'd like to craft some common/reusable tests for components that take these common props.

styles/mixins/patterns/_containers.scss
lyzadanger: You're right to note that this pattern isn't quite normalized where we need it yet. I spent a few minutes trying to see if I could re-jigger it (also taking the comment below about color into consideration), but the pattern lacks clarity out of context. What I'd like to propose is that we carry this forward into the next set of work, which involves a `Table` component that uses the `Scrollbox` component as a wrapper. Then we can scrutinize this pattern in better context and make some improvements.

PR #172, opened 2021-08-11
Make all container components take `containerRef`; consolidate tests

lyzadanger:
Per some verbal discussion today, there is a props API naturally shaping up with some of the container components. There's an immediate need for a `ref` on the `Scrollbox` container (for `Table`), so I went ahead and added the `containerRef` prop to all container components. Tests are consolidated here, as well, to DRY out common container tests.

### DIFF
--- a/src/components/containers.js
+++ b/src/components/containers.js
@@ -4,12 +4,16 @@
 
 import classnames from 'classnames';
 
+import { downcastRef } from '../util/typing';
+
 /**
  * @typedef {import('preact').ComponentChildren} Children
  *
  * @typedef ContainerProps
  * @prop {Children} children
  * @prop {string} [classes] - Additional CSS classes to apply
+ * @prop {import('preact').Ref<HTMLElement>} [containerRef] - Access to the
+ *  wrapping element.
  */
 
 /**
@@ -17,8 +21,15 @@ import classnames from 'classnames';
  *
  * @param {ContainerProps} props
  */
-export function Frame({ children, classes = '' }) {
-  return <div className={classnames('Hyp-Frame', classes)}>{children}</div>;
+export function Frame({ children, classes = '', containerRef }) {
+  return (
+    <div
+      className={classnames('Hyp-Frame', classes)}
+      ref={downcastRef(containerRef)}
+    >
+      {children}
+    </div>
+  );
 }
 
 /**
@@ -26,8 +37,15 @@ export function Frame({ children, classes = '' }) {
  *
  * @param {ContainerProps} props
  */
-export function Card({ children, classes = '' }) {
-  return <div className={classnames('Hyp-Card', classes)}>{children}</div>;
+export function Card({ children, classes = '', containerRef }) {
+  return (
+    <div
+      className={classnames('Hyp-Card', classes)}
+      ref={downcastRef(containerRef)}
+    >
+      {children}
+    </div>
+  );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/util/typing.js
```diff
@@ -0,0 +1,20 @@
...
+ *   return <div ref={myRef}>Hi there</div>;
+ * }
+ *
+ * The <div> wants {Ref<HTMLDivElement>}, while the component accepts the more
+ * generic {Ref<HTMLElement>}. TS' invariance in this case gets in our way.
+ * We want to keep the flexibility of accepting an {Ref<HTMLElement>} here.
+ */
+export function refElement(arg) {
+  return /** @type {import('preact').Ref<any>} */ (arg);
+}
```
robertknight, line 20: This explanation is clear although I think the term "downcast" probably best describes what we're doing with the input. I took a shot at creating a generic version of this that has the proper types:

```js
/**
 * @template T
 * @typedef {import('preact').Ref<T>} Ref
 */

/**
 * Helper for downcasting a ref to a more specific type, where that is safe
 * to do.
 *
 * This is mainly useful to cast a generic `Ref<HTMLElement>` to a more specific
 * element type (eg. `Ref<HTMLDivElement>`) for use with the `ref` prop of a JSX element.
 * Since Preact only writes to the `ref` prop, such a cast is safe.
 *
 * @template T
 * @template {T} U
 * @param {Ref<T>|undefined} ref
 * @return {Ref<U>|undefined}
 */
export function downcastRef(ref) {
  return /** @type {Ref<U>|undefined} */ (ref);
}
```

The above works in place of the current `refElement` calls, if you adjust the name. The type variables T and U could also be given more descriptive names like `Base` and `Derived`.

src/components/containers.js
```diff
@@ -4,30 +4,48 @@
...
 
+import { refElement } from '../util/typing';
+
 /**
  * @typedef {import('preact').ComponentChildren} Children
  *
  * @typedef ContainerProps
  * @prop {Children} children
  * @prop {string} [classes] - Additional CSS classes to apply
+ * @prop {import('preact').Ref<HTMLElement>} [containerRef] - Access to the
```
robertknight, line 15: What are the use cases you have in mind for `containerRef`?

robertknight (commented):
I left some comments after an initial look through. The main feedback I had was a suggestion regarding the `refElement` helper name and a query about the use cases we have for the `containerRef` prop.

src/components/containers.js
lyzadanger: E.g. https://github.com/hypothesis/frontend-shared/pull/174/files#diff-39adc517b1457f125e1e729d1039b7ce6a3a4c39317a18af693ff418d5cf6073R153

src/util/typing.js
lyzadanger: > This explanation is clear although I think the term "downcast" probably best describes what we're doing with the input.

Heh, this is why I specifically asked for vocabulary when we chatted about this. I'll use your suggestions here.

lyzadanger:
@robertknight Have at! I've integrated your version.

src/components/test/containers-test.js
```diff
@@ -1,99 +1,70 @@
...
+        ],
+        [className, 'foo', 'bar']
+      );
+    });
+
+    it('passes along a `ref` to the input element through `containerRef`', () => {
+      const containerRef = createRef();
+      createComponent(Component, { containerRef });
+
+      assert.isTrue(containerRef.current instanceof Node);
```
robertknight, line 39: `assert.instanceOf(object, constructor)` will give a better error if this check fails.

robertknight (approved):
LGTM

src/components/test/containers-test.js
lyzadanger: Oh, that is a nice assertion to use, thanks.

PR #176, opened 2021-08-31
Make all Library pages import from package entry point

### LINKED ISSUE #124: Make pattern library component modules import from package entry point
Most Pattern-library page-components import their relevant components from the component modules directly. While the components themselves and their tests should avoid importing from the package's entry point, importing from there within pattern-library modules could serve as a sanity check to assure that those components are exported as expected in the package.

e.g. this could help prevent: https://github.com/hypothesis/frontend-shared/pull/123

lyzadanger:
Pattern Library page component modules should import common components from the package's entry point. This makes the Pattern Library more clearly a consumer of the package, and serves as an additional layer of assurance that we are exporting package components correctly.

Also: make imports more consistent in these modules (ordering, alphabetization...)

Fixes #124

### DIFF
--- a/src/pattern-library/components/patterns/ContainerPatterns.js
+++ b/src/pattern-library/components/patterns/ContainerPatterns.js
@@ -1,8 +1,7 @@
 import { useState } from 'preact/hooks';
 
-import Library from '../Library';
-
 import { IconButton, LabeledButton } from '../../../';
+import Library from '../Library';
 
 import { SampleListElements } from './samples';
 
--- a/src/pattern-library/components/patterns/DialogComponents.js
+++ b/src/pattern-library/components/patterns/DialogComponents.js
@@ -1,16 +1,17 @@
 import { createRef } from 'preact';
 import { useState } from 'preact/hooks';
+
+import Library from '../Library';
+
 import {
   ConfirmModal,
   Dialog,
+  IconButton,
   LabeledButton,
   Modal,
-  TextInputWithButton,
   TextInput,
-  IconButton,
+  TextInputWithButton,
 } from '../../../';
-
-import Library from '../Library';
 
 import { LoremIpsum } from './samples';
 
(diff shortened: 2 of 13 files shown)


src/pattern-library/components/patterns/ContainerPatterns.js
```diff
@@ -1,9 +1,9 @@
 import { useState } from 'preact/hooks';
 
-import Library from '../Library';
-
 import { IconButton, LabeledButton } from '../../../';
 
+import Library from '../Library';
```
esanzgar, line 5: I believe the newline above should be removed. Both exports starts with `'../[...]'`, so they should be in the same section.

src/pattern-library/components/patterns/DialogComponents.js
```diff
@@ -1,17 +1,18 @@
...
+  IconButton,
   LabeledButton,
   Modal,
-  TextInputWithButton,
   TextInput,
-  IconButton,
+  TextInputWithButton,
 } from '../../../';
 
-import Library from '../Library';
```
esanzgar, line 13: This placement of the import was correct, except for the newline above.

src/pattern-library/components/patterns/FormPatterns.js
```diff
@@ -1,7 +1,7 @@
-import Library from '../Library';
-
 import { IconButton, SvgIcon } from '../../../';
 
+import Library from '../Library';
```
esanzgar, line 3: Same about the newline above.

esanzgar (approved):
👍🏼

src/pattern-library/components/patterns/FormPatterns.js
esanzgar: Also applies to other changes below.

lyzadanger:
@esanzgar I've removed the extra line breaks per your suggestion. Thanks!

PR #183, opened 2021-09-02
Create types using `tsconfig.json`

esanzgar:
Problem: the `tsc` command that generate the type declaration didn't
include all the options used for type checking the project.

Solution: use `tsc --build` command for both, type checking, and the
generation of the type declaration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,7 @@
   },
   "scripts": {
     "build-js": "babel src --out-dir lib --source-maps --ignore '**/test' --ignore '**/karma.config.js'",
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
I can confirm that building the package with these settings fixes the problematic typing for the `selectedItem` prop on the `Table` component, and that linking my local `lms` branch against a build of this eliminates the TS error I was encountering earlier today on that prop.

I'm not certain if there are other things I should keep an eye out for, or other risks raised by this change, but from what I can see here, it looks like it does what it is supposed to!

esanzgar:
The only minor difference regarding this change is that checking the types will generate type declaration. Before this change, it didn't do that. But I think we can live with that.

PR #186, opened 2021-09-03
Replace `<Fragment>` for `<>`

esanzgar:
We prefer `<></>` JSX syntax to `<Fragment></Fragment>`. It also avoids
an extra import.


lyzadanger (approved):
👍🏻 Thanks for keeping us to our conventions!

Issue #177, opened 2021-09-01
Establish common props API for shared components


lyzadanger:
Problem: Shared components have certain common props but these are inconsistently named and supported. This slows us down because it's hard to know which components do what and what their props are called.

Solution: Apply a consistent, typed set of shared props to components as each is converted and updated.

Outcome: Props available in updated components should be consistent and predictable for dev users.

Not part of this work:

* Documenting components or their props. Important work, just not part of this task.

lyzadanger:
There is a rough consensus around this common API now.

PR #188, opened 2021-09-07
Add support for `classes` prop to Button components

lyzadanger:
This PR makes Button components accept `classes`.

No two ways about it, the Button components were created before many lessons were subsequently learned :). 

Before these changes, an author could provide `className` to a `Button` component and it would _replace_ the Button's own CSS classes. That support is still present, but this PR adds a `classes` prop that behaves as it does elsewhere. It doesn't touch `className`, as that is being used widely, and I'll reconsider that as a separate issue.

Getting this in place will allow app customizations of button styling to be less horrifyingly complex, and has some immediate-term usefulness.

Part of #177

### DIFF
--- a/src/components/buttons.js
+++ b/src/components/buttons.js
@@ -5,6 +5,10 @@ import { SvgIcon } from './SvgIcon';
 /**
  * @typedef ButtonProps
  * @prop {import('preact').Ref<HTMLButtonElement>} [buttonRef]
+ * @prop {string} [classes] - Optional CSS class name(s) to use _in addition_
+ *   to the button component's own clases
+ * @prop {string} [className] - Optional CSS class name that will _replace_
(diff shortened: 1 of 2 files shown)


esanzgar (approved):
I don't like too much the name `classes` because I tend to reserve plurals for `Array`. For people like me, you may be tempted to use it like this by the name: `classes={['a', 'b']}`.

I guess at this time changing this is no possible or desirable.

lyzadanger:
> I don't like too much the name classes because I tend to reserve plurals for Array. For people like me, you may be tempted to use it like this by the name: classes={['a', 'b']}.

Hmmm, I hear you. I was following the lead from MaterialUI on this prop naming, FWIW. e.g. https://material-ui.com/api/button/ (or any other component) takes `classes`

PR #189, opened 2021-09-08
Add support for other "text-like" input types to TextInput

lyzadanger:
When working on converting LMS' `URLPicker` over to use the shared `Modal`, I was also wanting to convert it to use `TextInput` for its URL-input field. However, `URLPicker` relies on browser validation on `input type="url"`, and `TextInput` wasn't allowing any `type` except text.

This PR adds support for the most "text-like" input `type`s, per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input — we could open this up to more in the future if we wanted.

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -9,13 +9,12 @@ import classnames from 'classnames';
  *   the rendered `input` element.
  * @prop {boolean} [hasError] - There is an error associated with this input. Will
  *   set some error styling.
+ * @prop {'email'|'search'|'text'|'url'} [type="text"] - Set the <input> type:
+ *   restricted to the "text-like" type values.
  */
 
 /**
- * `TextInput` sets the `type` attribute on the `input`, but any other valid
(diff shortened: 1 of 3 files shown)


src/components/TextInput.js
```diff
@@ -9,13 +9,12 @@ import classnames from 'classnames';
  *   the rendered `input` element.
  * @prop {boolean} [hasError] - There is an error associated with this input. Will
  *   set some error styling.
+ * @prop {'email'|'search'|'text'|'url'} [type="text"] - Set the <input> type:
```
esanzgar, line 12: We could add 'password' too (while not having a specific component for that).

Something to notice about the 'search' is that in Chrome it adds a 'X' cancel symbol. I am not sure if we mind that:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#differences_between_search_and_text_types

esanzgar (approved):
👍🏼

PR #192, opened 2021-09-10
Serve pattern library at web root

### LINKED ISSUE #40: Fix routing and `baseURL` for locally-served pattern library
When working on #20 I ran into a wrinkle with the preact router in which trying to set the `baseURL` to `/` blew up on me:

Screen Shot 2021-04-27 at 1 09 44 PM

We should track that down, because serving the pattern library at the web root seems a whole lot more convenient.

lyzadanger:
Make the Pattern Library serve at the web root instead of under
`ui-playground`. This is convenient and eliminates the need to have
an index route/template.

Fixes #40


robertknight (approved):
LGTM 👍

PR #199, opened 2021-09-23
Change SVG icon imports from CommonJS to ES imports

robertknight:
This will facilitate an upcoming migration to a new module bundler (I
have been testing Rollup, but it likely applies to others as well) which
speaks ES modules natively.


lyzadanger (approved):
I've tested a build of this branch against both the `lms` and `client` applications. "Does not appear to break."

robertknight:
> "Does not appear to break."

That's the kind of ringing endorsement I aspire to 😉

Issue #139, opened 2021-07-09
Modals should default to NOT close on external click, focus events


lyzadanger:
There are cases when we use a Modal component in which it shouldn't be possible to close the modal without choosing an adventure (i.e. clicking a button).

 "Cancel" may still be valid, and may still close the modal. But in some cases, it's problematic to have clicks outside of the modal component close the modal—e.g. in certain LMS contexts—or it's "too easy" to accidentally close a modal. And/or the close button (X) is misleading or distracting.

Before we do this, we need to look into proper focus-trapping (see Keyboard Support section of https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html).

One implementation approach that's been looked at here involves updating the `useElementShouldClose` hook to be configurable.

lyzadanger:
Now part of #77

PR #201, opened 2021-09-29
Provide more options for closing Modals and Dialogs

lyzadanger:
This PR contains updates to `Modal`, `Dialog` and the `useElementShouldClose` hook to support several use cases in our applications.

Fixes https://github.com/hypothesis/frontend-shared/issues/139
Part of https://github.com/hypothesis/lms/issues/3139
Related to / a dependency for https://github.com/hypothesis/lms/issues/3106

### Dialog with a close button but no cancel button

I want to be able to replace the `SidebarPanel` `Panel` components in the client with `Dialog` for consistency and the nice focus routing that `Dialog` provides:

image

That UI renders a close button but does not render any buttons in the panel/dialog itself. Previous to these changes, the `Dialog` component would always automatically render both a close and a cancel button if an `onCancel` (i.e. close callback) is provided. Now there is granular control via two new props: `withCloseButton` and `withCancelButton` (which default to `true` such that default behavior is unchanged).

#### Similarly, a modal with a cancel button but no close button

In LMS, the `OAuth2RedirectErrorApp` modals have a cancel button (it's labeled "Close" here but it is a "cancel button" from the perspective of the Dialog) but no close ("X") button:

OAuthErrorApp-scope-keys

These changes support this.

### Modal that doesn't close when you click outside of it

When working within assignment-configuration Modals in the LMS, it's easy to accidentally click outside the Modal and close it (e.g. when selecting text in an input field, or navigating across several file folders in the file picker). These changes include a new `Modal` prop: `closeOnExternalInteraction`. When disabled, the Modal (assuming it has an `onCancel` callback) will still close when `ESC` is pressed, but not on other external events.

### Modal that doesn't close at all

This use case is specific to the LMS `ErrorDialogApp` use case, in which a "modal" is displayed but provides no further functionality:

ErrorDialogApp-reused_tool_guid

These changes include the ability to not provide an `onCancel` to the `Modal`.

## Testing

These changes include a commit that adds several [new Dialog examples to the pattern library](http://localhost:4001/components-dialogs).

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -10,11 +10,10 @@ import { LabeledButton } from './buttons';
 
 /**
  * @typedef ModalBaseProps
- * @prop {boolean} [closeOnExternalInteraction=true] - By default, a modal
- *   with a provided `onCancel` callback will close when a user clicks outside
- *   the modal, or focus is moved outside of the modal. In some cases, this
- *   may not be desireable. When this option is disabled, the modal may still
- *   be closed by pressing ESC.
+ * @prop {boolean} [closeOnExternalInteraction=false] - By default, a modal
+ *   with a provided `onCancel` callback will close when a user presses ESC.
+ *   Enabling this option will also close the modal if the user clicks outside
+ *   of the modal, or if focus is moved outside of the modal.
  *
  * @typedef {ModalBaseProps & DialogProps} ModalProps
  */
@@ -29,7 +28,7 @@ import { LabeledButton } from './buttons';
 export function Modal({
   children,
   onCancel,
-  closeOnExternalInteraction = true,
+  closeOnExternalInteraction = false,
   ...restProps
 }) {
   const modalContainerRef = useRef(/** @type {HTMLDivElement | null} */ (null));
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -10,26 +10,33 @@ import { LabeledButton } from './buttons';
 
 /**
  * @typedef ModalBaseProps
- * @prop {() => void} onCancel - `onCancel` is required for Modals
- */
-
-/**
- * @typedef {DialogProps & ModalBaseProps} ModalProps
+ * @prop {boolean} [closeOnExternalInteraction=true] - By default, a modal
+ *   with a provided `onCancel` callback will close when a user clicks outside
+ *   the modal, or focus is moved outside of the modal. In some cases, this
+ *   may not be desireable. When this option is disabled, the modal may still
+ *   be closed by pressing ESC.
+ *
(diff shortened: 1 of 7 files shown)


lyzadanger:
Note: there is a one-line test miss. I'll fix.

src/components/Modal.js
```diff
@@ -10,26 +10,33 @@ import { LabeledButton } from './buttons';
...
- * @prop {() => void} onCancel - `onCancel` is required for Modals
- */
-
-/**
- * @typedef {DialogProps & ModalBaseProps} ModalProps
+ * @prop {boolean} [closeOnExternalInteraction=true] - By default, a modal
+ *   with a provided `onCancel` callback will close when a user clicks outside
+ *   the modal, or focus is moved outside of the modal. In some cases, this
+ *   may not be desireable. When this option is disabled, the modal may still
+ *   be closed by pressing ESC.
```
robertknight, line 17: Hmm. How strong a reason do we have to continue supporting closing the dialog when clicking outside of it? IIRC the original reason for this was to avoid needing to do focus trapping (ie. preventing focus from moving outside of the dialog). If we are going to prevent dialogs closing by interacting with external elements then I think we'll need to implement focus trapping as well.

robertknight (commented):
Hi Lyza - I've had an initial look through this. One question that comes to mind is whether any of the supported variations of the dialog will become unnecessary in future (eg. should we continue to support closing on external interactions)? Also if we're going to add a mode where external interactions don't close the dialog then I think we'll need to implement focus trapping (ie. so keyboard focus cycles only through controls in the modal when opened).

> In LMS, the OAuth2RedirectErrorApp modals have a cancel button (it's labeled "Close" here but it is a "cancel button" from the perspective of the Dialog) but no close ("X") button:

Should this dialog have a close button as well? I can't remember what the reason was for omitting in this context.

src/components/Modal.js
lyzadanger: Ooh! Thank you for thinking this through and not letting me proceed blindly.

Let's see... the close behavior as initially implemented in this hook was intended for things like menus which _should_ go away if the user clicks or focuses away, but for modals, this behavior as default (for modals/dialogs) may not make sense.

I just did a quick sweep of all uses of `<Modal>`, `<ConfirmModal>` and `<Dialog>` in `client` and `lms` and am of the opinion that close-on-external-click-or-focus should _not_ be the default behavior. (ESC should still close, however).

Unless you demur, I'm going to go ahead and make this change.

lyzadanger:
Closing this PR in favor of splitting this work into two:

* Control over rendered dialog buttons (close, cancel)
* Control over events that close a modal: needs further think and focus management

PR #203, opened 2021-09-30
Allow configuration of close, cancel buttons on Dialog

lyzadanger:
This PR contains updates to `Dialog` to address a couple of application use cases dealing with close and cancel buttons.

Part of https://github.com/hypothesis/lms/issues/3139
Related to / a dependency for https://github.com/hypothesis/lms/issues/3106

### Dialog with a close button but no cancel button

I want to be able to replace the `SidebarPanel` `Panel` components in the client with `Dialog` for consistency and the nice focus routing that `Dialog` provides:

image

That UI renders a close button but does not render any buttons in the panel/dialog itself. Previous to these changes, the `Dialog` component would always automatically render both a close and a cancel button if an `onCancel` (i.e. close callback) is provided. Now there is granular control via two new props: `withCloseButton` and `withCancelButton` (which default to `true` such that default behavior is unchanged).

#### Similarly, a modal with a cancel button but no close button

In LMS, the `OAuth2RedirectErrorApp` modals have a cancel button (it's labeled "Close" here but it is a "cancel button" from the perspective of the Dialog) but no close ("X") button:

OAuthErrorApp-scope-keys

These changes support this.

### DIFF
--- a/src/components/Dialog.js
+++ b/src/components/Dialog.js
@@ -48,6 +48,11 @@ function useUniqueId(prefix) {
  *   "Cancel" button will be displayed.
  * @prop {'dialog'|'alertdialog'} [role] - The aria role for the dialog (defaults to" dialog")
  * @prop {string} title
+ * @prop {boolean} [withCancelButton=true] - If `onCancel` is provided, render
+ *   a Cancel button as one of the Dialog's buttons (along with any other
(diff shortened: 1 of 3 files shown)


robertknight (approved):
LGTM

Issue #208, opened 2021-10-08
Normalize and reset: stop the insanity!


lyzadanger:
The time is unfortunately nigh to unravel the different approaches to CSS normalization (making sure browser default quirks are accounted for), CSS reset (actually resetting styles on some or all elements) and element styling (applying styles to HTML elements, without regard for classnames) across our apps.

### The current situation

* Client
  * Normalization: Some c-n-p'ed normalize rules that stretch way back until about [this commit](https://github.com/necolas/normalize.css/blob/master/normalize.css) from a decade ago. There was some rumor that there might be some sort of legacy-technical reason for not using `normalize.css` directly, but in digging back through a lot of git history I didn't see anything too clarifying. These rules are out of date and incomplete compared to `normalize.css` itself.
  * Reset: Extensive, manual. Most block elements have their margins, padding and borders reset. Also sets `box-sizing: border-box` on `*`.
  * Elements: Primarily links (`<a>`) and paragraph (`<p>`) styles; makes all button cursors `pointer`.
* LMS
  * Normalization: `normalize.css` without any customization
  * Reset: None, except CSS generated for _back-end_ use resets `box-sizing: border-box` universally. The front-end CSS does _not_.
  * Elements: none.
* Pattern library/this project
  * Modeled after the client
  * Exported Pattern Library CSS includes its own resets and element styles — baddddd. It should only contain the CSS relevant to styling the `Library.*` components themselves.

### How to eject

I believe LMS is closer to where we want to go, which I'd propose is:

* Rely on `normalize.css`, which is maintained by smart people, for normalization, without being additionally clever
* Pare back resets and/or remove completely. [^1]
* Remove element styles per our CSS authoring standards of using low-specificity class name selectors.
* Break up pattern-library styling such that exported styles only contain true `Library.*` styles

[^1]: I recognize that there are cases where we want certain element-like styles to "apply automatically," a good example being, say, rendered markdown HTML. In this case, we want `<a>` tags to "look like" styled/Hypothesis links. But I would suggest we handle this with a wrapping pattern, e.g. `formattedText` or something, applied to the wrapping/parent element, which could encapsulate some of these styles.

#### Steps

* [x] Replace custom normalization in this package with `normalize.css`. This has no effect on consuming applications, but will help to spot if there is any effect on component styles or pattern styles. This should be fairly easy.
* [x] Remove resets from this package. Like above, this has no effect on consuming applications. Vet patterns to make sure they aren't relying on any resets.
[...]

PR #209, opened 2021-10-08
Update CSS normalization

lyzadanger:
Part of https://github.com/hypothesis/frontend-shared/issues/208

Replace custom, legacy normalization rules with `normalize.css`, similar to how it's done in the `lms` app.

### DIFF
--- a/styles/base/_reset.scss
+++ b/styles/base/_reset.scss
@@ -1,57 +1,5 @@
+@use 'normalize.css/normalize';
 @use '../mixins/reset';
-
-/*
-  Consistency fixes
-  adopted from http://necolas.github.com/normalize.css/
-  */
-
-* {
-  box-sizing: border-box;
-}
-
-html {
-  height: 100%;
-  text-size-adjust: 100%;
-}
-
-body {
-  min-height: 100%;
-  font-size: 100%;
-  margin: 0;
-}
-
-sub,
-sup {
-  font-size: 75%;
-  line-height: 0;
-  position: relative;
-}
-
-sup {
-  top: -0.5em;
-}
-sub {
-  bottom: -0.25em;
-}
-
-pre {
-  white-space: pre;
-  white-space: pre-wrap;
-  word-wrap: break-word;
-}
-
-b,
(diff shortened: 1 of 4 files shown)


styles/base/_reset.scss
```diff
@@ -1,58 +1,6 @@
+@use 'normalize.css/normalize';
 @use '../mixins/reset';
 
-/*
-  Consistency fixes
-  adopted from http://necolas.github.com/normalize.css/
```
robertknight, line 5: I note this URL was obsolete (the current one is https://necolas.github.io/normalize.css/).

styles/base/index.scss
```diff
@@ -1,2 +1,13 @@
...
+* {
+  box-sizing: border-box;
+}
+
+button,
+html input[type='button'],
+input[type='reset'],
+input[type='submit'] {
+  cursor: pointer;
+}
```
robertknight, line 13: Should these rules not go in `_elements.scss`?

styles/base/index.scss
```diff
@@ -1,2 +1,13 @@
 @use 'reset';
 @use 'elements';
+
+* {
+  box-sizing: border-box;
+}
+
+button,
+html input[type='button'],
```
robertknight, line 9: Why is the `html` ancestor needed for this particular rule?

robertknight (commented):
I had a couple of queries, otherwise LGTM.

styles/base/index.scss
lyzadanger: My plan is to entirely obliterate `_elements` in a few steps from now, so this is a place to isolate remaining things that I need to take a second look at.

styles/base/index.scss
lyzadanger: Yeah, I haven't had a chance to track down this history of these rules. I have the same question.

styles/base/_reset.scss
lyzadanger: Yeah, I found that out the hard way, too

PR #212, opened 2021-10-15
Split pattern library SASS and remove remaining resets and element styles

lyzadanger:
Depends on https://github.com/hypothesis/frontend-shared/pull/211

Part of https://github.com/hypothesis/frontend-shared/issues/208

This PR accomplishes a few tasks toward establishing a common styling baseline (reset, normalize, element styles) across our applications. It removes the remaining resets and element styles from the SASS modules in this package, and splits the generated pattern-library stylesheets into two[^1]. It also opportunistically fixes a glitch with the page-not-found handling.

User-visible changes are minor. Removing resets effectively restores a few sensible browser defaults for headings, lists, etc. Styling is simplified for a number of headings, etc. (this is a rather red diff). Any user-visible changes are confined to the pattern library itself.

These changes work toward a goal that a pattern library served within an external application should show what patterns will look like _in that app_, without any shared-package assumptions about resets or whatnot.

[^1]: Ultimately, external applications can opt to serve their local pattern library with a combination of the app's own styles and _just_ the pattern-library styles for the pattern-library components themselves. This will help ensure that what one sees in an app's instance of the pattern library is really how those patterns will look in that app.


styles/library.scss
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  &__heading3 {
+    font-weight: normal;
+    font-style: italic;
+  }
+
+  &__heading4 {
+    font-style: italic;
+  }
```
lyzadanger, line 22: I'm not married to this naming convention going forward; it's an interim step of isolating these styles.

styles/library.scss
```diff
@@ -0,0 +1,174 @@
+/**
+ * This stylesheet contains styles for pattern-library components.
```
lyzadanger, line 2: These styles we extracted from `pattern-library.scss`

PR #213, opened 2021-10-18
Convert bundling to use Rollup

robertknight:
Change bundling in development and when running tests to use ~~Browserify~~ _Rollup_
rather than ~~Rollup~~ _Browserify_. This adapts the changes that were recently made in
the lms [1] and client [2] projects.

I'm aware that there is currently a fair amount of duplication (~120 lines or so) in the helper functions in gulpfile.js. I'm planning to look at extracting these into a separate package future (say `@hypothesis/frontend-build` or something) to simplify the per-project configuration.

[1] https://github.com/hypothesis/lms/pull/3221
[2] https://github.com/hypothesis/client/pull/3840

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -14,6 +14,13 @@ function bundleConfig(name, entryFile) {
       chunkFileNames: '[name].bundle.js',
       entryFileNames: '[name].bundle.js',
     },
+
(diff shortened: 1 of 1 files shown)


robertknight:
I have pushed a change which suppresses the warning globally. This is OK for the moment since the issue existed with Browserify as well, it's just that Browserify didn't warn about it.

PR #215, opened 2021-10-19
Simplify gulpfile using functions from @hypothesis/frontend-build

robertknight:
Use functions from the @hypothesis/frontend-build project to handle building of JS and CSS bundles, and running tests. This allows the project's gulpfile to be significantly simplified.

As part of this change, gulpfile.js has been converted to a native ES module (`.mjs` extension). This is because @hypothesis/frontend-build is a native ES module package and only ES modules can directly import other ES modules.


src/karma.config.js
```diff
@@ -24,7 +24,7 @@ module.exports = function (config) {
       // CSS bundle relied upon by accessibility tests (eg. for color-contrast
       // checks).
       {
-        pattern: '../build/styles.css',
+        pattern: '../build/styles/index.css',
```
esanzgar, line 27: Was this previously wrong?

PR #217, opened 2021-10-19
v2: Revise icon registration API

lyzadanger:
This PR builds on work in https://github.com/hypothesis/frontend-shared/pull/163 . It aims to alleviate problems with key conflicts when registering icons for use.

Loose ends are tied up, and a few additions and changes are here versus #163. I'll call those out in code comments.

The biggest discussion item is that I'm not convinced we should necessarily deprecate `registerIcons`. The symbol-keyed registry used here in this common package will obviate the original problem when this package is consumed. On the app side, continuing to use `strings` as keys could be ergonomically preferable to managing references to symbols. I grok that there isn't really any overhead to re-registering an icon, but it feels a little awkward. Perhaps we could see how things shake out as we start updating app icon modules?

These changes also move the project's `icons` module into `pattern-library`, as that is the only place that it is currently used. 

There are several bits of follow-up work to this, including figuring out how the API should look for using common icons out of this package...


src/components/SvgIcon.js
```diff
@@ -75,18 +75,39 @@ export function SvgIcon({ name, className = '', inline = false, title = '' }) {
...
+ *
+ * Returns a symbol that can be passed as the `name` prop to `SvgIcon` in order
+ * to render this icon.
+ *
+ * @param {string|symbol} name - A name for this icon
+ * @param {string} markup - SVG markup for the icon
+ * @return {symbol}
+ */
+export function registerIcon(name, markup) {
+  const key = typeof name === 'string' ? Symbol(name) : name;
```
lyzadanger, line 89: Accepting `symbol` here allows for re-registering an icon, as an analog to the `reset` option in `registerIcons`. The need specifically came up in a test: tests were calling `registerIcons` with `reset=true` and there needed to be a way to restore the removed icons.

src/components/SvgIcon.js
```diff
@@ -75,18 +75,39 @@ export function SvgIcon({ name, className = '', inline = false, title = '' }) {
...
+  return key;
+}
+
 /**
  * Register icons for use with the `SvgIcon` component.
  *
- * @param {IconMap} icons
- * @param {object} options
+ * @deprecated Prefer the `registerIcon` function instead which will return a
+ * key that does not conflict with existing icons.
```
lyzadanger, line 98: As I mentioned in the PR description, I'm not 100% sold that we do want to deprecate this, but I have an open mind.

src/components/test/Checkbox-test.js
```diff
@@ -65,7 +65,7 @@ describe('LabeledCheckbox', () => {
   it('renders a checkbox SvgIcon for styling the checkbox visually', () => {
     const wrapper = createComponent();
     const icon = wrapper.find('SvgIcon');
-    assert.equal(icon.props().name, 'hyp-checkbox');
+    assert.equal(icon.props().name.toString(), Symbol('checkbox').toString());
```
lyzadanger, line 68: This kosher?

src/components/test/SvgIcon-test.js
```diff
@@ -12,9 +17,13 @@ describe('SvgIcon', () => {
   // Global icon set that is registered with `SvgIcon` outside of these tests.
   let savedIconSet;
 
-  beforeEach(() => {
-    savedIconSet = availableIcons();
+  before(() => {
+    // Clone the returned Map so that we're not affected by changes to the
+    // Map it references
+    savedIconSet = new Map(availableIcons());
```
lyzadanger, line 23: Ran into some "interesting" behavior here during my travels.

test/bootstrap.js
```diff
@@ -9,13 +9,6 @@ import { Adapter } from 'enzyme-adapter-preact-pure';
 
 configure({ adapter: new Adapter() });
 
-// Make all the icons that are available for use with `SvgIcon` in the actual
-// app available in the tests. This enables validation of icon names passed to
-// `SvgIcon`.
-import testIcons from '../src/icons';
-import { registerIcons } from '../src/components/SvgIcon';
-registerIcons(testIcons);
-
```
lyzadanger, line 18: Components should be registering their needed icons, so let's get rid of these training wheels.

src/components/buttons.js
```diff
@@ -27,15 +27,16 @@ import { SvgIcon } from './SvgIcon';
 
 /**
  * Fold in HTML button prop definitions into ButtonProps, but ignore `size` because it's inherited
- * from HTMLElement and conflicts with the _ButtonProps.size prop above.
+ * from HTMLElement and conflicts with the _ButtonProps.size prop above. Ignore `icon`
+ * because it is typed to `string` only and we need to be able to accept {string|symbol}
  *
- * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLButtonElement>, 'size'> } HTMLButtonElementProps
+ * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLButtonElement>, 'size' | 'icon'> } HTMLButtonElementProps
```
lyzadanger, line 33: Figuring THIS out was fun. Thank goodness for good IDE integrations.

src/pattern-library/icons.js
```diff
@@ -0,0 +1,27 @@
+// @ts-nocheck
```
lyzadanger, line 1: This module has just been moved.

src/components/test/Checkbox-test.js
robertknight: Yes, I think this is fine, as long as we only care about the symbol's description here rather than symbol equality.

src/components/test/SvgIcon-test.js
robertknight: `availableIcons` is only used in tests, and is thus not performance sensitive, so we could make it return a map copy itself.

PR #233, opened 2021-11-01
Add appropriate ARIA attributes for button role="tab"

lyzadanger:
It turns out that different ARIA attributes are required to denote
"on"-ness of buttons with a role of "tab." The `aria-pressed` attribute
is disallowed, and the `aria-selected` attribute is required.

Add some logic to support this.

I discovered this problem when attempting to transition the custom `button` in the client's `SelectionTabs` component over to a `LabeledButton`. These buttons are `role="tab"` within a `div role="tablist"` and a11y tests do not pass without this change (they do pass after this change).


robertknight (approved):
Looks correct according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role.

Issue #194, opened 2021-09-15
Add `FullScreenSpinner` component


lyzadanger:
The LMS application uses a `FullScreenSpinner` component, which does what it sounds like it would: renders a spinner centered in the viewport, with a semi-opaque overlay behind it.

Let's add that to this collection of components, using the standardized `spinner` pattern.

Issue #229, opened 2021-10-27
Epic: Establish patterns for shared icons, Icon component


lyzadanger:
This epic tracks the set of tasks needed to migrate to a set of shared icons (in this project's package), as well as a new `Icon` component to replace the `SvgIcon` component.

Why?

* Make this project the source of truth and reference for Hypothesis' icon set. Each of our apps (before this epic's work is done) has their own icon set and source files. These icons have some commonalities, and also some inconsistencies. There's duplication of source. And there's no one place of reference to see or access the entire set. Apps no longer need to maintain their own stables of icon source, but can import what they need from this package.
* A new `Icon` component brings our icon implementation into our convention standards and remediates some challenges with `SvgIcon`. The `SvgIcon` component has a convention mismatch with our naming conventions, so we've been planning to rename it. But there are also some challenges with the way that `SvgIcon` is structured and styled, which has required some hacky CSS workarounds. The new `Icon` component is also in line with our standard component API (e.g. takes standard props).
* Establishing icon patterns in this central, shared place will give us the foundational tools to prototype new, needed patterns that involve icons.

This work includes:

* [x] refactoring how icon registration works to avoid conflicts
* [x] Establishing an `Icon` component to replace `SvgIcon`
* [x] Using shared icons, wherever possible in LMS
* [x] Using shared icons, wherever possible in the client
* [x] Using `Icon` instead of `SvgIcon` in LMS
* [x] Using `Icon` instead of `SvgIcon` in the client
* [x] Deprecating `SvgIcon`
* [ ] Adding an `Icon Component` pattern library page
* [ ] Update `SvgIcon` to `Icon` in the pattern library itself
* [ ] Update `IconButton` to use `Icon` instead of `SvgIcon`
* [ ] Sort out inconsistencies between local and shared icons of the same name (e.g. annotate, sort)
* [ ] Evaluate and/or add utility classes, patterns or mixins for icons.

lyzadanger:
This epic as written is no longer applicable, but some of the tasks have been completed as spec'ed. I'm closing now, and will likely track `Icon`-related needs as a separate ticket/project.

Issue #224, opened 2021-10-26
Deprecate SvgIcon and add IconComponents pattern library page


lyzadanger:
Now that `Icon` is available, follow up:

* Deprecate he `SvgIcon` component
* Move icon-registration functions into `Icon`
* Add a pattern-library page for the `Icon` component

lyzadanger:
Related work is now tracked in #451

PR #237, opened 2021-11-05
Update Spinner and add FullScreenSpinner pattern and component

lyzadanger:
This PR updates `Spinner` to use `Icon` instead of `SvgIcon` and adds a new `FullScreenSpinner` component and associated pattern, as well as examples in the pattern library.

Motivation: LMS' local `FullScreenSpinner` is the last use of `SvgIcon` in that project, and updating components here to use `Icon` instead of `SvgIcon` is part of icon-transition work.

Fixes https://github.com/hypothesis/frontend-shared/issues/194
Part of https://github.com/hypothesis/frontend-shared/issues/229

To see the new pattern and component in action, spin up your local and visit the [new pattern page](http://localhost:4001/patterns-spinners) and the [new component page](http://localhost:4001/components-spinner).

### DIFF
--- a/src/components/Spinner.js
+++ b/src/components/Spinner.js
@@ -2,7 +2,8 @@ import classnames from 'classnames';
 
 // @ts-ignore
 import spinnerSVG from '../../images/icons/spinner--spokes.svg';
-import { registerIcon, SvgIcon } from './SvgIcon';
+import { Icon } from './Icon';
+import { registerIcon } from './SvgIcon';
 
 // Register the spinner icon for use
 const spinnerIcon = registerIcon('spinner', spinnerSVG);
@@ -22,6 +23,9 @@ const spinnerIcon = registerIcon('spinner', spinnerSVG);
 export function Spinner({ classes = '', size = 'medium' }) {
   const baseClass = `Hyp-Spinner--${size}`;
   return (
-    <SvgIcon name={spinnerIcon} className={classnames(baseClass, classes)} />
+    <Icon
+      name={spinnerIcon}
+      containerClasses={classnames(baseClass, classes)}
+    />
   );
 }
--- a/src/components/Spinner.js
+++ b/src/components/Spinner.js
@@ -16,6 +16,12 @@ const spinnerIcon = registerIcon('spinner', spinnerSVG);
  */
 
 /**
+ * @typedef FullScreenSpinnerProps
+ * @prop {string} [classes] - Additional CSS classes to apply
+ * @prop {string} [containerClasses] - CSS classes to apply to wrapping element.
+ */
+
+/**
  * Loading indicator.
  *
  * @param {SpinnerProps} props
@@ -27,5 +33,21 @@ export function Spinner({ classes = '', size = 'medium' }) {
       name={spinnerIcon}
       containerClasses={classnames(baseClass, classes)}
     />
+  );
+}
+
+/**
+ * Full-screen loading indicator.
+ *
+ * @param {FullScreenSpinnerProps} props
+ */
+export function FullScreenSpinner({ classes = '', containerClasses = '' }) {
+  return (
+    <div className={classnames('Hyp-FullScreenSpinner', containerClasses)}>
+      <Spinner
+        classes={classnames('Hyp-FullScreenSpinner__spinner', classes)}
+        size="large"
+      />
+    </div>
   );
 }
--- a/src/components/test/Spinner-test.js
+++ b/src/components/test/Spinner-test.js
@@ -9,25 +9,25 @@ describe('Spinner', () => {
 
   it('renders', () => {
     const wrapper = createSpinner();
-    assert.isTrue(wrapper.find('SvgIcon.Hyp-Spinner--medium').exists());
+    assert.isTrue(wrapper.find('Icon .Hyp-Spinner--medium').exists());
   });
 
   it('uses the registered `hyp-spinner` icon', () => {
     const wrapper = createSpinner();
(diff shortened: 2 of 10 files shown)


src/components/Spinner.js
```diff
@@ -2,7 +2,8 @@ import classnames from 'classnames';
 
 // @ts-ignore
 import spinnerSVG from '../../images/icons/spinner--spokes.svg';
-import { registerIcon, SvgIcon } from './SvgIcon';
+import { Icon } from './Icon';
+import { registerIcon } from './SvgIcon';
```
lyzadanger, line 6: Moving `registerIcon` into `Icon` is on my short list and is tracked in https://github.com/hypothesis/frontend-shared/issues/224

styles/mixins/patterns/_spinners.scss
```diff
@@ -15,6 +17,18 @@
...
+  font-size: $-size;
+}
+
+/**
+ * Style a container with an svg.&__spinner as a full-screen loading spinner,
+ * by applying an overlay to the container and fixed-centered positioning to
+ * the contained svg icon
+ */
+@mixin full-screen-spinner() {
+  @include layout.overlay($variant: 'light');
```
lyzadanger, line 29: The implementation of a full-screen spinner on LMS uses a light background, and I do think it looks nicer than a heavy, dark one.

src/components/test/Spinner-test.js
```diff
@@ -9,25 +9,73 @@ describe('Spinner', () => {
...
+    );
+  });
+
+  it('applies container classes', () => {
+    const wrapper = createSpinner({ containerClasses: 'foo bar' });
+
+    assert.deepEqual(
+      [
+        ...wrapper
+          .find(`div.Hyp-FullScreenSpinner.foo.bar`)
```
robertknight, line 73: The `find` selector seems overly specific here, since you're already checking the exact classes on the matching element.

src/pattern-library/components/patterns/SpinnerComponents.js
```diff
@@ -25,6 +30,22 @@ export default function SpinnerComponents() {
           </Library.Demo>
         </Library.Example>
       </Library.Pattern>
+
+      <Library.Pattern title="Full-Screen Spinner">
+        <Library.Example>
+          <p>
+            A component that renders a full-screen spinner over an overlay. Note
+            that this page has to be reloaded to clear the full-screen spinner
+            after showing it.
```
robertknight, line 39: It would be nice to have a way of dismissing the spinner without a full page reload. Maybe something like:

```diff
diff --git a/src/pattern-library/components/patterns/SpinnerComponents.js b/src/pattern-library/components/patterns/SpinnerComponents.js
index ab244bb..17f9f84 100644
--- a/src/pattern-library/components/patterns/SpinnerComponents.js
+++ b/src/pattern-library/components/patterns/SpinnerComponents.js
@@ -7,6 +7,7 @@ export default function SpinnerComponents() {
   const [fullScreenSpinnerVisible, setFullScreenSpinnerVisible] = useState(
     false
   );
+  const hideSpinner = () => setFullScreenSpinnerVisible(false);
   return (
     <Library.Page title="Spinner">
       <p>
@@ -42,7 +43,13 @@ export default function SpinnerComponents() {
             <LabeledButton onClick={() => setFullScreenSpinnerVisible(true)}>
               Show Full-Screen Spinner
             </LabeledButton>
-            {fullScreenSpinnerVisible && <FullScreenSpinner />}
+            {fullScreenSpinnerVisible && (
+              // Suppress a11y errors
+              // eslint-disable-next-line
+              <div onClick={hideSpinner}>
+                <FullScreenSpinner />
+              </div>
+            )}
           </Library.Demo>
         </Library.Example>
       </Library.Pattern>
```

(If there is an easy fix for the a11y complaint great, but otherwise I think we can accept it as a known issue for this developer-facing tool).

src/pattern-library/components/patterns/SpinnerPatterns.js
```diff
@@ -21,23 +26,23 @@ export default function SpinnerPatterns() {
             At its default size, the spinner is <code>2em</code> square.
           </p>
           <Library.Demo withSource>
-            <SvgIcon name="hyp-spinner" className="hyp-spinner" />
+            <Icon name="hyp-spinner" containerClasses="hyp-spinner" />
```
robertknight, line 29: The notes on the "Hypothesis UI playground" homepage say that the Patterns are built "using HTML and CSS" whereas components "are Preact components that are built using underlying patterns. I note that we are using `Icon` here, which I guess is OK assuming in a non-JS context we'd have some other way to render the same SVG icon (like the Jinja2 helper that h has). I'm also ignoring the usage of components which are just helpful for the demo page and not required to actually use the pattern.

robertknight (approved):
LGTM with some minor comments.

src/pattern-library/components/patterns/SpinnerPatterns.js
lyzadanger: You're right! This is non-normalized. There's a bit of work here because the offending bit is actually that we don't have a "pattern" established for `icon`, just a component. I'm going to track this for follow-up work.

src/pattern-library/components/patterns/SpinnerComponents.js
lyzadanger: @robertknight I've pushed a commit that uses `use-element-should-close` hook to make it so that you can close the spinner with `ESC` in both pattern library examples.

PR #239, opened 2021-11-15
Revert "Add types for SVG "modules""

lyzadanger:
Reverts hypothesis/frontend-shared#230

See also https://github.com/hypothesis/lms/pull/3374


robertknight (approved):
LGTM. To summarize why this is needed: it caused typechecking errors in downstream projects because the `src/types/svg.d.ts` files were not included in the package. We're planning to fix this in future by compiling SVG => JS as part of the package build, rather than relying on downstream projects to do it.

PR #241, opened 2021-11-17
PoC: Use tailwind to help us automate our utilities layer, simplify structure, speed prototyping

lyzadanger:
This spike represents about 2 1/2 hours of exploration into using [`tailwindcss`](https://tailwindcss.com/) to speed up, optimize and simplify the generation and management of our utility styles.

## Why

Encapsulation of patterns and applying those patterns with components for export is working well in our approach. Using a utility-first approach when prototyping, both here and in other apps, is also working well in practice.

To this point, this package has been mapping tokens/variables in SASS to a specific set of utility classes, that are then exposed to applications using this package (prefixed `.hyp-u-`). This approach works all right, it's not dreadful, but in practice I've seen a few scaling challenges arise:

* Utility classes have to be "thought of" in advance and generated by hand in various SASS mixin modules. I often find myself building out in another application that uses this package and realizing there isn't an available utility class to do what I want.
* Utility classes have to be maintained and built. This is a bit of a chore. This includes naming and testing and whatnot.
* Utility classes need to be documented and, frankly, aren't.
* There's a conflation of true utility classes, e.g., `.hyp-u-bg-color--grey-3`, which sets a specific background color, a single rule, with atomic patterns, e.g. `.hyp-u-border`, which is an encapsulation of a common border color, width and style.
* We don't have a mechanism for pruning unused utility classes.

Thus, I spend quite a bit of time gardening and weeding utility classes and mechanics, versus focusing on isolating and expressing design patterns.

### This spike

This spike demonstrates using the `tailwindcss` library to help us out on the utilities layer so that we can spend our time solving Hypothesis-design-specific problems and less on housekeeping. It rewrites one atomic pattern and one molecular pattern using `tailwindcss` syntax (`hyp-u-border`, `hyp-link`), and shows how the usage of auto-generated color utility classes in concert with standard tailwind utility classes can be used to quickly rebuild the Color Foundations page in the pattern library.

### Tailwind in brief

`tailwindcss` is a `postcss` plugin. You feed it a config (`tailwind.config.js` here, and yes, configs can extend other configs, which would be useful in downstream projects) and it generates every conceivable utility class from this config. This removes the need to build, document and name things, because the [`tailwind` docs](https://tailwindcss.com/docs) are quite thorough. Later, we can use [the `purge` feature](https://tailwindcss.com/docs/optimizing-for-production) to effectively tree-shake our CSS classes and tighten out the build for production.

`tailwindcss` prefers to act as its own, solo preprocessor but it will play well with SASS and that's how I have it set up here. This allows us to lean on our SASS infrastructure either during transition or indefinitely. 

### Risks

I've satisfied myself with respect to a few risks already:

* Tailwind can be added to our existing build without too much fuss. Tailwind directives (`@apply`, `@tailwind`) can be used in our SASS files without SASS pitching a fit[^1]. I'm able to provide a tailwind config from this project to the spike branch of `frontend-build`.
* Dropping in Tailwind utilities into our current SASS pile here and adding classes to the utilities and components ["layers"](https://tailwindcss.com/docs/functions-and-directives#layer) works. I'm also able to use the `@apply` directive to reuse other classes that are in turn composed of `@apply` directives. I wanted to make sure such a thing was possible. I'm also able to mix and match tailwind with SASS `@include`s.
* The utility classes _work_ and the little prototyping work on `ColorFoundations` was pleasurable and quick. This is actually quite important!

Some other risks I'm aware of but I believe are surmountable:

* Tuning build correctly to prune unused CSS without over-pruning.
* Getting some ES module-related funkiness sorted out. I'm getting this warning but I believe it can be sorted out with a little help from ES-module-expert Rob (Typescript seems happy enough, though, after I made sure to add the appropriate types dependency in `frontend-build`).

Screen Shot 2021-11-17 at 8 21 08 AM

Human-factors risks:

* The learning curve associated with `tailwindcss`. This mostly consists of recognizing some of its shorthand in naming, as it aims to keep class names very short. Once you understand the basic abbreviations ("p" for padding, e.g.) it's fairly readable. 
* Finding the _right place_ to start documenting our applied-design philosophy, which Rob pointed out has evolved considerably. I'm happy to write words down. But _where_? And who gets consulted/provides approval?

## Try it out

To see this in action:

* Check out the [tailwind-spike](https://github.com/hypothesis/frontend-build/tree/tailwind-spike) branch of `frontend-build`
* In the `frontend-build` directory, run `yalc publish`
* Check out this branch of `frontend-shared`
* In the `frontend-shared` branch, run `yalc add @hypothesis/frontend-build`
* Run `make dev`

You can visit http://localhost:4001/foundations-colors to see an example of some stuff built with `tailwindcss` utility classes. You can visit http://localhost:4001/patterns-links to see an example of a pattern (`.hyp-link`) reimplemented using `tailwindcss` instead of a mixin.

image

[^1]: My editor is still complaining about style linting on these directives but that is an editor-level problem that is solvable. An aside that we could define CSS/SASS linting in general in our projects if we so desired.

### DIFF
--- a/src/pattern-library/components/patterns/ColorFoundations.js
+++ b/src/pattern-library/components/patterns/ColorFoundations.js
@@ -1,74 +1,43 @@
+import tailwindConfig from '../../../../tailwind.config';
+
 import Library from '../Library';
 
 // TODO:
 // - color descriptions and guidance
-// - hex and other metadata about colors
 // - foreground color examples
-// - more examples of overriding with `!` classes
 // - valid contrast combinations
 
-const brand = ['brand'];
+const colorConfig = tailwindConfig.theme.extend.colors;
 
-const greys = [
-  'white',
-  'grey-0',
-  'grey-1',
-  'grey-2',
-  'grey-3',
-  'grey-4',
-  'grey-5',
-  'grey-6',
-  'grey-7',
-  'grey-8',
-  'grey-9',
-];
-
-const states = ['success', 'notice', 'error'];
-
-function ColorSwatch({ colorName }) {
+function ColorSwatch({ colorKey }) {
   return (
     <div>
-      <div
-        className={`hyp-u-bg-color--${colorName}`}
-        style="width:16rem;height:2rem;margin-right:1em"
-      />
+      <div className={`bg-${colorKey} h-12 mr-4`} />
       <p>
-        <i>{colorName}</i>
+        <i>
+          {colorKey}: {colorConfig[colorKey]}
+        </i>
       </p>
     </div>
   );
 }
 
-const brandExamples = brand.map(colorName => (
-  <ColorSwatch key={colorName} colorName={colorName} />
-));
-
-const greyExamples = greys.map(colorName => (
-  <ColorSwatch key={colorName} colorName={colorName} />
-));
-
-const stateExamples = states.map(colorName => (
-  <ColorSwatch key={colorName} colorName={colorName} />
+const colorSwatches = Object.keys(colorConfig).map(colorKey => (
+  <ColorSwatch key={colorKey} colorKey={colorKey} />
 ));
 
 export default function ColorFoundations() {
   return (
     <Library.Page title="Colors">
-      <Library.Pattern title="Brand red">
-        <div className="hyp-u-layout-row" style="flex-wrap:wrap">
-          {brandExamples}
+      <Library.Pattern title="Color swatches">
+        <div className="grid md:grid-cols-2 lg:grid-cols-3 gap-4">
+          {colorSwatches}
         </div>
       </Library.Pattern>
 
-      <Library.Pattern title="Greys">
-        <div className="hyp-u-layout-row" style="flex-wrap:wrap">
-          {greyExamples}
-        </div>
-      </Library.Pattern>
-
-      <Library.Pattern title="State indicators">
-        <div className="hyp-u-layout-row" style="flex-wrap:wrap">
-          {stateExamples}
+      <Library.Pattern title="Atomic example">
+        <div className="p-4 hyp-u-border">
+          Bordered with <code>.hyp-u-border</code>
         </div>
       </Library.Pattern>
     </Library.Page>
--- a/src/pattern-library/components/patterns/ColorFoundations.js
+++ b/src/pattern-library/components/patterns/ColorFoundations.js
@@ -7,15 +7,20 @@ import Library from '../Library';
 // - foreground color examples
 // - valid contrast combinations
 
-const colorConfig = tailwindConfig.theme.extend.colors;
+const colorConfig = tailwindConfig.theme.extend.colors ?? {};
 
 function ColorSwatch({ colorKey }) {
+  const colorValue =
+    typeof colorConfig[colorKey] === 'string'
+      ? `: ${colorConfig[colorKey]}`
+      : '';
   return (
     <div>
       <div className={`bg-${colorKey} h-12 mr-4`} />
       <p>
         <i>
-          {colorKey}: {colorConfig[colorKey]}
+          {colorKey}
+          {colorValue}
         </i>
       </p>
     </div>
--- a/styles/patterns/index.scss
+++ b/styles/patterns/index.scss
@@ -1,7 +1,16 @@
 @use 'containers';
 @use 'forms';
-@use 'links';
 @use 'panels';
 @use 'spinners';
 @use 'tables';
 @use 'thumbnails';
+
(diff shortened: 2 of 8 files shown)


src/pattern-library/components/patterns/ColorFoundations.js
```diff
@@ -1,74 +1,48 @@
...
-        <div className="hyp-u-layout-row" style="flex-wrap:wrap">
-          {brandExamples}
-        </div>
-      </Library.Pattern>
-
-      <Library.Pattern title="Greys">
-        <div className="hyp-u-layout-row" style="flex-wrap:wrap">
-          {greyExamples}
+      <Library.Pattern title="Color swatches">
+        <div className="grid md:grid-cols-2 lg:grid-cols-3 gap-4">
```
lyzadanger, line 38: This concise bit of utility usage, boom, gives us a three-column grid on wide screens, 2 on mid, 1 on narrow, with appropriate gaps, without me needing to author and test widely. Yep, we can define the breakpoints (and/or rename them) in our config.

styles/patterns/index.scss
```diff
@@ -1,7 +1,16 @@
...
 @use 'spinners';
 @use 'tables';
 @use 'thumbnails';
+
+@use '../mixins/focus';
+
+@layer components {
+  .hyp-link {
+    @include focus.outline-on-keyboard-focus;
+    @apply inline-block text-brand no-underline rounded-sm;
```
lyzadanger, line 13: `rounded-sm` applies a border radius. Once again, we can define what those radii are in our config.

styles/util/index.scss
```diff
@@ -3,3 +3,11 @@
 @use 'colors';
 @use 'focus';
 @use 'layout';
+
+@tailwind utilities;
+
+@layer utilities {
+  .hyp-u-border {
+    @apply border border-solid border-grey-3;
+  }
```
lyzadanger, line 12: This is an example of an atomic pattern, which can get confused with true utility classes. Here it is implemented by composing tailwind utilities.

tailwind.config.js
```diff
@@ -0,0 +1,25 @@
+export default {
+  theme: {
+    extend: {
```
lyzadanger, line 3: In this spike, I used `extend` to extend tailwind's default configuration, which already includes like a billion colors. We'd likely want to _not_ extend but instead define the entire theme to prevent style bloat.

src/pattern-library/components/patterns/ColorFoundations.js
```diff
@@ -1,74 +1,48 @@
...
-  'grey-4',
-  'grey-5',
-  'grey-6',
-  'grey-7',
-  'grey-8',
-  'grey-9',
-];
-
-const states = ['success', 'notice', 'error'];
+const colorConfig = tailwindConfig.theme.extend.colors ?? {};
```
lyzadanger, line 10: Because tailwind config is just a POJO, we can get at it without having to repeat ourselves as we do with SASS variables.

src/pattern-library/components/patterns/ColorFoundations.js
lyzadanger: I should also point out that we can use an option in tailwind to prefix all tailwind classes if we want.

src/pattern-library/components/patterns/ColorFoundations.js
```diff
@@ -1,74 +1,48 @@
...
+    typeof colorConfig[colorKey] === 'string'
+      ? `: ${colorConfig[colorKey]}`
+      : '';
   return (
     <div>
-      <div
-        className={`hyp-u-bg-color--${colorName}`}
-        style="width:16rem;height:2rem;margin-right:1em"
-      />
+      <div className={`bg-${colorKey} h-12 mr-4`} />
```
lyzadanger, line 19: In case it's not clear, it's tailwind magically providing `bg-<color>` classes here. It also gives you `text-<color>`, `border-<color>`, etc.

styles/patterns/index.scss
```diff
@@ -1,7 +1,16 @@
...
+    @apply hover:no-underline hover:text-brand-dark;
```
lyzadanger, line 14: All of these `@apply` directives could totally be on one line; I split them up for readability

styles/util/index.scss
```diff
@@ -3,3 +3,11 @@
 @use 'colors';
 @use 'focus';
 @use 'layout';
+
+@tailwind utilities;
```
lyzadanger, line 7: `@import "tailwindcss/utilities"` also works.

PR #242, opened 2021-11-23
Add tailwind dependency and provide tailwind preset

lyzadanger:
~~Depends on https://github.com/hypothesis/frontend-build/pull/8~~ 

_Updated_: I have elected not to export the Tailwind preset from the package's entry point; instead consuming applications can import it directly under `lib` (e.g. `import tailwindPreset from '@hypothesis/frontend-shared/lib/tailwind.preset.js'`). This makes Tailwind an optional dependency for consuming applications at this point unless they import that preset. I expect that Tailwind will be a hard dependency for consuming applications quite soon, however, based on how this package provides source SASS. In summary, this is not a breaking change and does not have a dependency on `frontend-build`'s support of Tailwind.

This PR attempts to provide a Tailwind preset for consumption by downstream apps. This can be used and extended in local Tailwind configurations (you can see that happening here in the local Tailwind configuration: it uses the preset).

### DIFF
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -1,0 +1,29 @@
+import colors from 'tailwindcss/colors.js';
+
+export default {
+  theme: {
+    colors: {
+      transparent: 'transparent',
+      current: 'currentColor',
+      white: colors.white,
+      black: colors.black,
+      'grey-0': '#fafafa',
+      'grey-1': '#f2f2f2',
+      'grey-2': '#ececec',
+      'grey-3': '#dbdbdb',
+      'grey-4': '#a6a6a6',
+      'grey-5': '#9c9c9c',
+      'grey-6': '#737373',
+      'grey-7': '#595959',
+      'grey-8': '#3f3f3f',
+      'grey-9': '#202020',
+      brand: {
+        dark: '#84141e',
+        DEFAULT: '#bd1c2b',
+      },
+      success: '#00a36d',
+      notice: '#fbc168',
+      error: '#d93c3f',
+    },
+  },
+};
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -1,0 +1,9 @@
+import tailwindPreset from './src/tailwind.preset.js';
+
+export default {
+  presets: [tailwindPreset],
+  mode: 'jit',
+  purge: {
+    content: ['./src/**/*.js', './templates/**/*.mustache'],
+  },
+};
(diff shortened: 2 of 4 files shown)


src/tailwind.preset.js
```diff
@@ -0,0 +1,29 @@
+import colors from 'tailwindcss/colors.js';
```
lyzadanger, line 1: I am completely unsure about extensions here and whether this preset should be `.js`, `.mjs`...I basically monkey-fingered combinations until I found something that worked and didn't complain.

robertknight: This is correct. `tailwindcss` is still a CommonJS package, and `colors.js` is therefore treated as a CommonJS module. The exports of CommonJS modules are converted to default exports when imported by an ESM package. In other words `tailwindcss/colors.js` contains:

```js
module.exports = { colorA: value, ... }
```

When importing this from an ES module, Node treats this as if the module contained:

```js
export default { colorA: value, ... }
````

File: src/tailwind.preset.js [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+  theme: {
+    colors: {
+      transparent: 'transparent',
+      current: 'currentColor',
+      white: colors.white,
+      black: colors.black,
```
robertknight, line 9: The values of `colors.white` and `colors.black` here are just the strings `#000` and `#fff`. Was there a reason to need the import?

tailwind.config.js
```diff
@@ -0,0 +1,9 @@
+import tailwindPreset from './src/tailwind.preset.js';
+
+export default {
+  presets: [tailwindPreset],
+  mode: 'jit',
+  purge: {
+    content: ['./src/**/*.js', './templates/**/*.mustache'],
```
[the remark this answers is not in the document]
robertknight: From some local testing in the client, limiting the scanned files to just components saves ~200ms or so off the `build-css` gulp task, out of a total of ~2s with Tailwind. In _this_ project however the difference will be much smaller.

robertknight (approved):

tailwind.config.js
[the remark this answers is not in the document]
lyzadanger: Might want to be more choosy about the `mustache`-related purge globs in the client, good idea.

src/tailwind.preset.js
lyzadanger: I'll look into this. I had the sense it might give us access some variants and opacity-related classes based on white and black, but if that's not the case, I may directly use the hex values. Still a bit of fiddling to come on getting these color settings just right.

PR #244, opened 2022-01-05
Add dependabot configuration

lyzadanger:
Let's get dependabot actually going here...


esanzgar (approved):
It looks good.

I know you assigned to Robert, but it was simple enough for me to approve it.

PR #245, opened 2022-01-05
Upgrade tailwind and add touch-size minimum to preset

lyzadanger:
Upgrade this package's local `tailwindcss` (used for building the CSS for the local pattern library) and add a new spacing value to the preset: `touch-minimum`, which is a dimension (`44px`) we use frequently in our apps.

### DIFF
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -25,5 +25,8 @@ export default {
       notice: '#fbc168',
       error: '#d93c3f',
     },
+    spacing: {
+      'touch-minimum': '44px', // Equivalent to spacing 11; minimum touch-target size
+    },
   },
 };
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "rollup-plugin-string": "^3.0.0",
     "sass": "^1.43.2",
     "sinon": "^9.0.0",
-    "tailwindcss": "^2.2.19",
+    "tailwindcss": "^3.0.11",
     "typescript": "^4.1.5",
     "yalc": "^1.0.0-pre.50"
   },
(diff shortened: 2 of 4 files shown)


src/tailwind.preset.js
```diff
@@ -25,5 +25,8 @@ export default {
       notice: '#fbc168',
       error: '#d93c3f',
     },
+    spacing: {
+      'touch-minimum': '44px', // Equivalent to spacing 11; minimum touch-target size
```
robertknight, line 29: Can you explain a bit more about how this is or will be used? I couldn't find a reference in the Tailwind docs.

robertknight (approved):
LGTM. I had a query about `touch-minimum`.

src/tailwind.preset.js
lyzadanger: @robertknight This is extending the `spacing` theme configuration values to include a custom value: `touch-minimum`. This makes a full set of tailwind utility classes based on this token available (e.g. `w-touch-minimum` or `m-touch-minimum`).

It's an abstracting token so that we can use `w-touch-minimum` (or `h-touch-minimum`) instead of the equivalent `w-11` (Tailwind's default spacing scale is `1 unit = 4px` or more accurately `0.25rem`) to provide more context. The set of default values for spacing an be seen on [Tailwind's documentation for customization of spacing units](https://tailwindcss.com/docs/customizing-spacing). To put it another way: `touch-minimum` is our own invention; we're extending Tailwind's defaults with this Hypothesis-custom token.

I had already added this locally to LMS' tailwind configuration and [made use of it in the `StudentSelector` component](https://github.com/hypothesis/lms/blob/9c1e56cc3696506641a02dbb654bcb4f5cc33b17/lms/static/scripts/frontend_apps/components/StudentSelector.js#L82). This is me extracting it to the common preset used by all apps.

Is this still unclear in any way? I'm happy to make it more clear—I want this to make sense!

robertknight: This makes sense, thanks!

PR #258, opened 2022-01-06
Extend tailwind preset with more goodies and support

lyzadanger:
This PR fleshes out the shared tailwind preset (design token/value definitions):

* Extends instead of replaces theme values to make sure all tailwind utilities are available
* Adds a default `borderColor` such that using Tailwind's `border` utility class alone without modification will have the same net effect as the current `hyp-u-border` (applies correct border width and color)
* Adds a variant to allow for styling the clean theme
* Renames state-related colors in the same manner as https://github.com/hypothesis/lms/pull/3554

These changes are _additive_ and don't cause any visible changes. Just getting ducks in a row. The next set of PRs will show these in action.

### DIFF
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -1,3 +1,5 @@
+import plugin from 'tailwindcss/plugin.js';
+
 import colors from 'tailwindcss/colors.js';
 
 export default {
@@ -34,4 +36,13 @@ export default {
       },
     },
   },
+  plugins: [
+    plugin(({ addVariant }) => {
+      // Add a custom variant such that the `theme-clean:` modifier is available
+      // for all tailwind utility classes. e.g. `.theme-clean:bg-white` would
+      // only apply (set the element's background color to white) if a parent
+      // element had the `.theme-clean` class.
+      addVariant('theme-clean', '.theme-clean &');
+    }),
+  ],
 };
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -3,6 +3,9 @@ import colors from 'tailwindcss/colors.js';
 export default {
   theme: {
     extend: {
+      borderColor: {
+        DEFAULT: '#dbdbdb',
+      },
       colors: {
         transparent: 'transparent',
         current: 'currentColor',
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -23,13 +23,19 @@ export default {
         'grey-7': '#595959',
         'grey-8': '#3f3f3f',
         'grey-9': '#202020',
+        green: {
+          success: '#00a36d',
+        },
+        yellow: {
+          notice: '#fbc168',
+        },
+        red: {
+          error: '#d93c3f',
+        },
         brand: {
           dark: '#84141e',
           DEFAULT: '#bd1c2b',
         },
-        success: '#00a36d',
-        notice: '#fbc168',
-        error: '#d93c3f',
       },
       spacing: {
         'touch-minimum': '44px', // Equivalent to spacing 11; minimum touch-target size
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -2,31 +2,33 @@ import colors from 'tailwindcss/colors.js';
 
 export default {
   theme: {
-    colors: {
-      transparent: 'transparent',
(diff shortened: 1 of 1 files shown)


src/tailwind.preset.js
```diff
@@ -1,32 +1,54 @@
...
+        'grey-2': '#ececec',
+        'grey-3': '#dbdbdb',
+        'grey-4': '#a6a6a6',
+        'grey-5': '#9c9c9c',
+        'grey-6': '#737373',
+        'grey-7': '#595959',
+        'grey-8': '#3f3f3f',
+        'grey-9': '#202020',
+        green: {
+          success: '#00a36d',
```
robertknight, line 27: What is the effect of nesting here?

src/tailwind.preset.js
```diff
@@ -1,32 +1,54 @@
...
-      'touch-minimum': '44px', // Equivalent to spacing 11; minimum touch-target size
     },
   },
+  plugins: [
+    plugin(({ addVariant }) => {
+      // Add a custom variant such that the `theme-clean:` modifier is available
+      // for all tailwind utility classes. e.g. `.theme-clean:bg-white` would
+      // only apply (set the element's background color to white) if a parent
+      // element had the `.theme-clean` class.
+      addVariant('theme-clean', '.theme-clean &');
```
robertknight, line 51: The selector here refers to the `theme-clean` class that is set on the root element by the `HypothesisApp` component in the client? Does the clean theme need to be part of this project or could it be part of the client's configuration?

robertknight (commented):
I had a query about which project the clean theme variant should live in and a general question about nested colors. Otherwise LGTM.

src/tailwind.preset.js
lyzadanger: I'll answer this in a specific way, but let me know if you were after an explanation of _why_ one would nest here.

Tailwind has a [nested object syntax](https://v1.tailwindcss.com/docs/customizing-colors#nested-object-syntax) it supports for colors. Paraphrasing the docs, the keys will "combine with parent keys" to form such classnames as `bg-green-success` or `bg-brand-dark`. Note that the special `DEFAULT` key is handled specially. Such that we end up with `bg-brand` and `bg-brand-dark` from this preset.

Does this help?

src/tailwind.preset.js
lyzadanger: > Does the clean theme need to be part of this project [...]

Are you asking in terms of mechanics (making sure Tailwind doesn't purge needed things) or strategy (it's weird to have some support for theming in this repo vs. some in the `client`)?

The first question is easy: we're fine mechanically to split it up. If we were to, say, use a class on a component in this project, let's call it `theme-clean:border-0` (turn off borders when the modifier `theme-clean` applies), Tailwind will see `theme-clean:border-0` and make sure it provides that utility class in the resulting stylesheet. That will look roughly like (this might be slightly off in a character or two):

```
.theme-clean .theme-clean\:border-0 {
  border-width: 0; // This is from Tailwind's `border-0` implementation
}
```

Tailwind doesn't care about where the outer `.theme-clean` comes from; adding the variant as is done here translates `theme-clean:border-0` into a selector based on the second argument passed to `addVariant`. 

As to the second question about strategy for the clean theme, there are many ways to shape that balloon into an animal. We could opt to do all of the `theme-clean` adaptations within within the client itself only, in fact we might want to. We could use the same `addVariant` there and extend the classes applied to shared components. So, instead of, say, a shared `Card` component here in this project adding clean-theme styling, we could use `Card` in the client like so:

```
<Card classes="theme-clean:border-0"> ... </Card>
```

In fact, that's likely a better approach. _But I don't want to break anything during the transition_, so propose we start here.

src/tailwind.preset.js
robertknight: Yes, thanks.

src/tailwind.preset.js
robertknight: Hi Lyza - I was thinking more about the strategy question. In other words, should the frontend-shared package know about the clean theme.

> In fact, that's likely a better approach. But I don't want to break anything during the transition, so propose we start here.

OK. I just wanted to make sure that we were considering where this logically belongs.

robertknight (approved):

PR #293, opened 2022-01-26
Enable and use Tailwind in pattern library 

lyzadanger:
This PR enables Tailwind layers and refactors pattern library components to use Tailwind layers and utility styles.

It refactors some of the "Foundations" patterns to account for Tailwind.

I'd like to release the package after getting this landed. It's been tested against the `client` and `lms`. 

After these changes, applications using pattern-library styles need to activate Tailwind layers first, as the pattern-library CSS adds rules to those layers. There's a [PR on LMS that readies that project for these changes](https://github.com/hypothesis/lms/pull/3603). The `client` is already ready.

There is more to do in terms of extracting patterns and DRYing out the Tailwind classes, but we'll get there.


src/pattern-library/components/Library.js
```diff
@@ -62,9 +63,9 @@ function Page({ children, title }) {
  */
 function Pattern({ children, title }) {
   return (
-    <section className="LibraryPattern">
-      <h2 className="Library__heading2">{title}</h2>
-      {children}
+    <section className="space-y-4">
```
lyzadanger, line 66: Every time you see `space-y-4`, think: we need a `Block` component :)

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -50,23 +50,26 @@ export default function PlaygroundApp({
   }
 
   return (
-    <main className="PlaygroundApp">
-      <div className="PlaygroundApp__sidebar">
-        <div className="PlaygroundApp__sidebar-home">
+    <main className="grid grid-cols-[20rem_1fr] w-screen h-screen">
```
lyzadanger, line 53: If you're new to the syntax, square brackets allow the use of arbitrary values in certain Tailwind utility classes. This obviates the need to enshrine this grid layout in the configuration, as it is a one-off.

src/pattern-library/components/patterns/ColorFoundations.js
```diff
@@ -1,75 +1,62 @@
...
+const brandExamples = (
+  <>
+    <ColorSwatch colorClass="bg-brand" colorName="brand" />
+    <ColorSwatch colorClass="bg-brand-dark" colorName="brand-dark" />
+  </>
+);
+
+const greyExamples = (
+  <>
+    <ColorSwatch colorClass="bg-white" colorName="white" />
```
lyzadanger, line 24: A reminder that this can't be looped/scripted as the utility styles need to be present as strings in the source, or else tailwind's purge won't see them and they won't be included. So each classname does need to be included here. I can think of other ways to do this — the string just literally needs to be present in this source, even in comments — but this seemed the least mysterious, if not so DRY.

src/pattern-library/components/patterns/DialogComponents.js
```diff
@@ -188,9 +189,9 @@ function LongModalExample() {
         onCancel={() => setDialogOpen(false)}
         title="Modal with overflowing content"
       >
-        <div className="hyp-scrollbox">
+        <Scrollbox classes="px-4 py-2">
```
lyzadanger, line 192: Part of the move toward removing the patterns layer, but sort of distracting here, admittedly.

src/pattern-library/components/patterns/IconFoundations.js
```diff
@@ -15,13 +15,13 @@ export default function IconFoundations() {
   return (
     <Library.Page title="Icons">
       <Library.Pattern title="Hypothesis icon set">
-        <div className="LibraryGrid">
+        <div className="grid grid-cols-8 gap-6">
           {Object.keys(icons).map(iconName => (
             <Card
               key={iconName}
-              classes="hyp-u-vertical-spacing hyp-u-layout-column--center IconTile"
+              classes="flex flex-col items-center bg-grey-0 border-[#9c9cab] space-y-3 rounded-lg"
```
lyzadanger, line 22: The border-color here is shortly to be enshrined as an actual color in our palette; it just hasn't happened yet. I'm going to make a note to myself to clean this up ASAP.

src/pattern-library/components/patterns/LayoutFoundations.js
```diff
@@ -1,159 +1,15 @@
...
-      )}
-    </div>
-  );
-}
-
 export default function LayoutFoundations() {
   const [showExample1, setShowExample1] = useState(false);
   const [showExample2, setShowExample2] = useState(false);
   const [showExample3, setShowExample3] = useState(false);
-  const scaleUnits = [
```
lyzadanger, line 62: We no longer need to build and maintain our own scaled spacing system, thank you tailwind. See Tailwind docs instead :).

src/pattern-library/components/Library.js
```diff
@@ -48,9 +49,9 @@ import { jsxToString } from '../util/jsx-to-string';
  */
 function Page({ children, title }) {
   return (
-    <section className="LibraryPage">
-      <h1 className="Library__heading1">{title}</h1>
-      {children}
+    <section className="max-w-6xl space-y-4 p-4">
+      <h1>{title}</h1>
```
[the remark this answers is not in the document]
esanzgar: Ok, the style is in the `@layer base` in the `library.scss`. Some how I was expecting this `@layer` on a 'css' file not a 'scss' one.

esanzgar (approved):
Great job! Apologies for the late review.

esanzgar:
I have noticed a few issues with the responsiveness to different layouts. For example the flex layout of the Icons page doesn't seem to work. Also the sidebar menu use to move to the top, but not longer does that.

I guess this are small details we can iron later, isn't?

lyzadanger:
> I have noticed a few issues with the responsiveness to different layouts. For example the flex layout of the Icons page doesn't seem to work. Also the sidebar menu use to move to the top, but not longer does that.

> I guess this are small details we can iron later, isn't?

Yes, this is expected, if lame.

src/pattern-library/components/patterns/DialogComponents.js
lyzadanger: My goal in this pass is remove dependencies in the Foundations and Components patterns on any Patterns, uh, patterns. It's arguable both ways whether to do this now or not. But it doesn't hurt. Patterns are going away very soon.

src/pattern-library/components/patterns/LayoutFoundations.js
```diff
@@ -174,20 +30,20 @@ export default function LayoutFoundations() {
                   showExample1 ? 'visibility:visible' : 'visibility:hidden'
                 }`}
               >
-                <div className="hyp-card">
+                <Card classes="w-[450px]">
```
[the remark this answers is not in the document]
lyzadanger: I'm preparing to remove the patterns layer imminently. Not sure what it buys us to leave dependencies in place where they don't need to be!

src/pattern-library/components/Library.js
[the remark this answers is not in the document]
lyzadanger: Ahhh! Yes, we are still using SASS as part of our build process....

PR #303, opened 2022-01-31
Add text colors to tailwind preset

lyzadanger:
This itsy-bitsy PR adds some text colors to the shared TW preset (these are copied in from the local `client` TW configuration).


esanzgar (approved):
👍🏼

PR #306, opened 2022-01-31
Reformat using latest Prettier version

esanzgar:


lyzadanger (approved):
Thank you! Let's get CI green again.

PR #307, opened 2022-01-31
Swap Tailwind `.js` files for `.mjs` extension

esanzgar:
This PR copies the extension format used in the client.


robertknight:
What was the motivation for this PR, just consistency with the client or something else?

The plain `.js` extension for the Tailwind config in this repository is intentional, and other repositories should in future follow the same pattern. The client repository has to use `.mjs` for some files so that Node will treat them as ES modules, even though the package as a whole is still CommonJS. The plan for the future is to convert all frontend packages to ES modules so that all of the JavaScript code, whether it runs in Node or the browser, uses the same module system.

See https://github.com/hypothesis/frontend-shared/commit/9acbc98f13d86451b1a7520e464a72078c7dae4e for the commit that changed this package from being CommonJS to ESM.

esanzgar:
Thanks, I understand now that imported JS files have the extension `.mjs` _only if_ the package is CommonJS. `fronten-shared` package as a whole is of type `module`.

All clear now.

PR #310, opened 2022-02-02
Add a few slate greys to the preset colors

lyzadanger:
A few years ago, Jared introduced some proposed slatey blue-greys to our proposed color palette: https://hypothesis-pattern-library.netlify.app/helpers (the "base" colors here).

There are just a few component patterns that have started making use of these colors (most specifically: `Table`). I'd like to get them sorted out and in our defined palette.

This PR defines slate colors for "values" 1, 3, 5 and 7. These slates are still nascent and only used in a few places.

Pattern library after these changes:

image

### DIFF
--- a/src/pattern-library/components/patterns/ColorFoundations.js
+++ b/src/pattern-library/components/patterns/ColorFoundations.js
@@ -36,6 +36,14 @@ const greyExamples = (
   </>
 );
 
+const slateExamples = (
+  <>
+    <ColorSwatch colorClass="bg-slate-1" colorName="slate-1" />
+    <ColorSwatch colorClass="bg-slate-3" colorName="slate-3" />
+    <ColorSwatch colorClass="bg-slate-7" colorName="slate-7" />
+  </>
+);
+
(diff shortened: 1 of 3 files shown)


src/pattern-library/components/patterns/ColorFoundations.js
```diff
@@ -55,6 +64,14 @@ export default function ColorFoundations() {
         <div className="flex flex-row flex-wrap gap-4">{greyExamples}</div>
       </Library.Pattern>
 
+      <Library.Pattern title="Slates">
+        <p>
+          These slightly blue greys may be used sparingly to help with
+          differentiation and clarity within interfaces.
```
robertknight, line 70: Is there specific functional guidance about when to use slate greys (eg. what kinds of elements etc.) or are you still figuring that out?

robertknight (approved):
LGTM. If in future we get a stronger sense of where the proposed slate greys should be used over regular greys, it would be good to make the docs more specific.

PR #339, opened 2022-02-25
Update project READMES with info about Tailwind conversion

lyzadanger:
Update and greatly simplify README for CSS/styling to ease confusion during transition to Tailwind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ $ npm install @hypothesis/frontend-shared --save
 
 ### SASS
 
+Your project must have `sass` and `tailwindcss` dependencies installed.
+
 To add styles for all shared components to your project's SASS:
 
 ```scss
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -16,6 +16,8 @@ $ npm install @hypothesis/frontend-shared --save
 
 ### SASS
 
+Your project must have `sass` and `tailwind` dependencies installed.
```
robertknight, line 19: The Tailwind npm package is called `tailwindcss`.

robertknight (approved):
LGTM except that the npm package name for Tailwind is `tailwindcss`. There is an npm package called `tailwind` but it is something entirely different.

README.md
lyzadanger: Thanks for the pointer! Updated.

PR #382, opened 2022-04-13
Forward HTML attributes for div-based presentational components

lyzadanger:
I had hoped to defer updates to this package's components until "Phase IV" of the current "applied design updates" (we're in "Phase III", in which _application_ components are being converted to Tailwind, etc.), but I have a need to use the `Card` component from this package in a place where it wouldn't be be usable without HTML attribute forwarding (specifically, the `ThreadCard` component in the client's sidebar).

This PR attempts to:

- Add attribute forwarding for `div`-based "container" components in this package
- Prefer naming `elementRef` to `containerRef` for ref forwarding
- Adjust typing and a few bits and bobs to reflect more recent conventions (i.e. "what we've learned lately")

Out of scope for now, in order to keep this quick and allow for more "learning" before we cast any of this in stone:

- DRYing out these components. I acknowledge there is a fair amount of boilerplate and repeated props. I am OK with this for now as it is contained within a single component module.
- Likewise, generating these components with another mechanism as we're exploring in the `client` right now...
- Anything to do with Tailwind whatsoever. These components still use SASS styling. We'll get to that later!

### DIFF
--- a/src/components/containers.js
+++ b/src/components/containers.js
@@ -1,31 +1,46 @@
-/**
- * Components for laying out content within a consistently-styled container
- */
-
 import classnames from 'classnames';
 
 import { downcastRef } from '../util/typing';
 
 /**
+ * Components for laying out content within a consistently-styled container
+ */
+
+/**
  * @typedef {import('preact').ComponentChildren} Children
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ * @typedef {import('preact').Ref<HTMLElement>} ElementRef
+ */
+
+/**
+ * @typedef CommonPresentationalProps
+ * @prop {Children} [children]
+ * @prop {string|string[]} [classes] - Optional extra CSS classes to append to
+ *   the component's default classes
+ * @prop {ElementRef} [containerRef] - Deprecated; use
+ *   `elementRef` instead.
+ * @prop {ElementRef} [elementRef]
  *
- * @typedef ContainerProps
- * @prop {Children} children
- * @prop {string} [classes] - Additional CSS classes to apply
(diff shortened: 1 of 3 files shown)


src/components/containers.js
```diff
@@ -1,31 +1,46 @@
...
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ * @typedef {import('preact').Ref<HTMLElement>} ElementRef
+ */
+
+/**
+ * @typedef CommonPresentationalProps
+ * @prop {Children} [children]
+ * @prop {string|string[]} [classes] - Optional extra CSS classes to append to
+ *   the component's default classes
+ * @prop {ElementRef} [containerRef] - Deprecated; use
```
robertknight, line 20: I assume the plan is to remove this in a semver-major release once downstream consumers have been migrated?

src/components/containers.js
```diff
@@ -1,31 +1,46 @@
...
+ */
+
 /**
  * @typedef {import('preact').ComponentChildren} Children
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ * @typedef {import('preact').Ref<HTMLElement>} ElementRef
+ */
+
+/**
+ * @typedef CommonPresentationalProps
```
robertknight, line 16: "Presentational" feels a bit verbose but I don't have an immediate suggestion for an alternative.

src/components/containers.js
lyzadanger: Exactly! I was thinking a major release with the Tailwind conversion and removing this then. I was going to mention that in the next Changelog (i.e. "`containerRef` will be removed in vWhatever").

PR #442, opened 2022-06-07
Remove manual CHANGELOG in preference of auto-generated GitHub notes

lyzadanger:
The release process is simplified if we rely on GitHub's auto-generated release notes instead of running a local `gulp` task to generate suggested CHANGELOG content.

CHANGELOG.md is already redundant to existing release notes, so removing it should not be a loss.


docs/releases.md
```diff
@@ -14,61 +14,21 @@ We use [Semantic Versioning](https://semver.org/#semantic-versioning-200).
...
-## 4. Merge and release
+## 2. Merge and release on GitHub
 
 1. Go through the PR/merge cycle for the version-update branch
 2. Create a [new GitHub release](https://github.com/hypothesis/frontend-shared/releases/new/)
 
    - Title: use the tag name
-   - Message/Description: Paste in changelog changes
+   - Click the `Auto-generate release notes` button to generate release notes.
+     Edit as desired.
```
robertknight, line 26: I'd like to adjust this to emphasize here that:

1. Proper release notes are required, and not an optional step.
2. It is important to manually refine the release notes to highlight breaking changes. The format advice from https://keepachangelog.com/en/1.0.0/ is still useful even if we are not keeping the changelog as a file in the source tree

robertknight (commented):
I think I'm on board with the general proposal to move change information from a CHANGELOG.md file to GitHub release notes, in order to simplify the release workflow. However I would like us to emphasize the importance of adding release notes and ensuring breaking changes are highlighted. You know this already obviously, but we've seen with previous employees that not everyone is used to these expectations around documentation.

robertknight (approved):

PR #443, opened 2022-06-10
Upgrade tailwindcss to v3.1.2

lyzadanger:
This version of tailwindcss introduces some additional typing but some
affordances were needed in `tsconfig.json` and the typing of the
shared tailwind preset to accommodate this.

### DIFF
--- a/src/tailwind.preset.js
+++ b/src/tailwind.preset.js
@@ -2,7 +2,7 @@ import plugin from 'tailwindcss/plugin.js';
 
 import colors from 'tailwindcss/colors.js';
 
-export default {
+export default /** @type {Partial<import('tailwindcss').Config>} */ ({
   theme: {
     extend: {
       borderColor: {
@@ -67,4 +67,4 @@ export default {
       addVariant('theme-clean', '.theme-clean &');
     }),
   ],
-};
+});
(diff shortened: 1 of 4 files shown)


src/tailwind.preset.js
```diff
@@ -2,7 +2,7 @@ import plugin from 'tailwindcss/plugin.js';
 
 import colors from 'tailwindcss/colors.js';
 
-export default {
+export default /** @type {Partial<import('tailwindcss').Config>} */ ({
```
lyzadanger, line 5: I think it's early days for `tailwindcss`'s type definitions and I was unable to find an appropriate type for a _preset_ , which is what this is, not a full `Config`. Without the `Partial` this will gripe about a missing `content` property, which is indeed required for configs, but not for presets.

Alternately, might be able to type this using `Omit`, but I'd rather just do this for now and wait for `tailwindcss` to add a `Preset` type or similar.

lyzadanger:
🤦🏻‍♀️ failing in CI. I'll look next week.

lyzadanger:
Ah, not my mistake for once: https://github.com/tailwindlabs/tailwindcss/issues/8586

lyzadanger:
Converting to draft until there is a `tailwindcss` release that fixes this.

Issue #312, opened 2022-02-04
Provide shared `focus-visible` styling


lyzadanger:
Problem: When we updated application UI styling to remove any dependencies on `.hyp-u-*` utility classes or mixins from this package, we were forced to add temporary, duplicative styling rules for our preferred `focus-visible` styling to each application, as that design pattern cannot be achieved with simple tailwind utility classes.

Solution: Implement a convention- and tailwind-friendly replacement for this mixin and `.hyp-u-` style so that applications may remove their custom temporary local implementations and use the common one provided by this package.

PR #449, opened 2022-06-14
Provide `.focus-visible-ring` utility

lyzadanger:
After the conversion of application UI component styling to updated conventions, there were just a few UI-styling needs that couldn't be fulfilled with a combination of tailwind utility styles or using shared components. This is one of them: our custom focus-ring styling (applied to elements via `:focus-visible` or polyfilled equivalent).

This PR adds a utility class to tailwind: `.focus-visible-ring`, adds some default styling configuration for focus rings and a simple pattern-library page to demonstrate the class in action.

This should be an exact replacement of the current focus-ring styling we use.

Fixes https://github.com/hypothesis/frontend-shared/issues/312

### DIFF
--- a/src/pattern-library/components/patterns/UtilityFoundations.js
+++ b/src/pattern-library/components/patterns/UtilityFoundations.js
@@ -1,0 +1,26 @@
+import Library from '../Library';
+
+export default function UtilityFoundations() {
+  return (
+    <Library.Page title="Utilities">
+      <Library.Pattern title="Focus rings">
+        <Library.Example title="Custom focus-visible ring">
+          <p>
+            The <code>.focus-visible-ring</code> utility class customizes an
+            interactive element&apos;s focus ring such that it is only visible
+            when the element has <code>:focus-visible</code> or the polyfilled
+            equivalent for browsers that do not support{' '}
+            <code>:focus-visible</code>.
+          </p>
+          <Library.Demo
+            title="input and button with .focus-visible-ring"
+            withSource
+          >
+            <input type="text" className="border focus-visible-ring" />
+            <button className="focus-visible-ring">Button</button>
+          </Library.Demo>
+        </Library.Example>
+      </Library.Pattern>
+    </Library.Page>
+  );
+}
--- a/src/tailwind.focus-visible-ring.js
+++ b/src/tailwind.focus-visible-ring.js
@@ -1,0 +1,62 @@
+import plugin from 'tailwindcss/plugin.js';
+
+/**
+ * Provide the `.focus-visible-ring` utility class.
+ *
+ * Add a utility class that will render a focus ring on a focused element
+ * only when it has keyboard focus, or otherwise has `:focus-visible`. Do not
+ * show a focus ring if the element has focus but is not `:focus-visible`.
+ *
+ * This plugin should only be necessary until such a time that we feel browser
+ * support for :focus-visible is acceptable. At that time, what this plugin
+ * does should be achievable with a few standard tailwind utility classes.
+ *
+ * This styling requires the browser to support the :focus-visible
+ * pseudo-selector [1] or for the JS polyfill [2] to be loaded.
+ *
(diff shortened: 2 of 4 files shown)


src/pattern-library/components/patterns/UtilityFoundations.js
```diff
@@ -0,0 +1,26 @@
+import Library from '../Library';
```
lyzadanger, line 1: Please don't worry too much about where this pattern-library page shows up or its nuances; I envision a re-org of pattern-library pages in the next sizeable chunk of work in this package.

src/tailwind.focus-visible-ring.js
```diff
@@ -0,0 +1,62 @@
...
+ *    Set ring shadow offset
+ * -  <button className="ring-grey-3 focus-visible-ring">Click click</button>
+ *    Set ring color
+ *
+ * Ring width is this theme's `ringWidth.DEFAULT`
+ */
+export default plugin(({ addUtilities, theme }) => {
+  const ringWidth = theme('ringWidth.DEFAULT');
+  // Based on TW ring/shadow rules https://tailwindcss.com/docs/ring-width
+  const focusRing = `var(--tw-ring-inset) 0 0 0 calc(${ringWidth} + var(--tw-ring-offset-width)) var(--tw-ring-color)`;
```
lyzadanger, line 36: I'm not gonna lie: This line is fiddly and involved. But: It will make the generated styles work with related tailwind utility classes (see example usage above) and does link to some related docs.

src/tailwind.preset.js
```diff
@@ -53,12 +55,23 @@ export default {
           light: '#737373',
         },
       },
+      ringColor: {
+        DEFAULT: '#59a7e8',
+      },
+      ringOpacity: {
+        DEFAULT: '1.0',
+      },
```
lyzadanger, line 63: Even though we use a `box-shadow` to style our focus rings, tailwind still involves the `ringOpacity` value when calculating `--tw-ring-color` and the default value is, annoyingly, `0.5`.

robertknight: It would be useful to capture this in a code comment.

src/tailwind.preset.js
```diff
@@ -53,12 +55,23 @@ export default {
           light: '#737373',
         },
       },
+      ringColor: {
+        DEFAULT: '#59a7e8',
```
robertknight, line 59: Might be worth linking to the original PR from Kyle that introduced the custom ring styling to document the accessibility reasons why we customized it.

src/tailwind.focus-visible-ring.js
```diff
@@ -0,0 +1,62 @@
...
+ * -  <button className="ring-inset focus-visible-ring">No, click me</button>
+ *    Same as above, but explicitly inset
+ * -  <button className="ring-offset-4 focus-visible-ring">Click!</button>
+ *    Set ring shadow offset
+ * -  <button className="ring-grey-3 focus-visible-ring">Click click</button>
+ *    Set ring color
+ *
+ * Ring width is this theme's `ringWidth.DEFAULT`
+ */
+export default plugin(({ addUtilities, theme }) => {
```
robertknight, line 33: We might want to consider applying our "no default exports" policy here for consistency with the rest of our code, even if that differs from Tailwind plugin conventions. I'll let you make a call on that.

robertknight (approved):
LGTM with some minor comments.

src/tailwind.preset.js
lyzadanger: Here's a PR in which the `focus-visible` polyfill was added to the client: https://github.com/hypothesis/client/pull/1479

lyzadanger: OK, I've timed out and can't find anything useful to add at the moment. The plugin explains what it does fairly clearly; we can add a "why" as a simple follow-up if we determine what we'd like to say :)

Issue #453, opened 2022-06-15
Stop shipping unused assets


lyzadanger:
Problem: We've stopped using several deprecated things and the entire collection of utility classes in anything using this package, but this package is still shipping them. That's confusing and could lead to bloat.

Solution: Identify and remove explicitly unused assets.

Outcome: This package no longer provides unused assets as listed below, and we make a major version release (as this is breaking).

Not in scope:

* Removing the entire "patterns" layer, patterns classes and patterns pages in the pattern library. Some of these patterns are still used by components (the "old", SASS versions).

-----

Known unused assets:

* All `.hyp-u-`-prefixed utility classes
* All SASS mixins
* `SvgIcon` (still used internally but not externally)

Issue #421, opened 2022-05-17
Pull common configuration from `lms` and `client` upstream to this package


lyzadanger:
Design configuration has been extracted during application component conversions into tailwind configurations in the `client` and `lms`, but those configurations are partially duplicative and it's hard to see which things are common to all applications.

Solution: Pull up common design configuration into this package's `tailwind.preset.js`.

Out of scope:

* Improvements or analyses of these values

Known tasks:

* [ ] `zIndex`: See `client`'s tailwind config

PR #466, opened 2022-06-21
Breaking: Remove all `.hyp-u-` utility classes from package

lyzadanger:
This PR removes all `.hyp-u-` prefixed classes from the package's CSS API, as these should be superseded by Tailwind-managed utility classes. None of the apps now are using any `.hyp-u-` classes, so while this is technically a breaking change (and will be represented as such in the release cycle), it won't break any real-life applications.

The mixins driving these utility classes need to remain in place as many are being used by component styles.

Part of https://github.com/hypothesis/frontend-shared/issues/453


src/pattern-library/components/patterns/ContainerPatterns.js
```diff
@@ -182,7 +182,7 @@ export default function ContainerPatterns() {
                 Show example
               </LabeledButton>
               <div
-                className="hyp-u-overlay"
+                className="fixed top-0 left-0 bottom-0 right-0 bg-black/50 z-5"
```
lyzadanger, line 185: Yes, this is wordy. This pattern may bear abstraction in future — an `Overlay` component? something? — but the `.hyp-u-overlay` class is not part of the CSS API and this styling is currently only used in pattern-library examples. So I have opted, for now, for the "unrolled" representation using TW utility classes to avoid any premature abstraction.

src/tailwind.preset.js
```diff
@@ -68,6 +68,14 @@ export default /** @type {Partial<import('tailwindcss').Config>} */ ({
...
         'touch-minimum': '44px', // Equivalent to spacing 11; minimum touch-target size
       },
+      zIndex: {
+        1: '1',
+        2: '2',
+        3: '3',
+        4: '4',
+        5: '5',
+        max: '2147483647',
+      },
```
lyzadanger, line 78: Bringing more configuration values up from applications is tracked in a separate issue (https://github.com/hypothesis/frontend-shared/issues/421) but I needed a `z-index` value available for styling an overlay using Tailwind within these changes.

robertknight (approved):
LGTM

PR #485, opened 2022-07-08
Prepare pattern library for component updates

lyzadanger:
This PR makes some quick and non-breaking updates to the pattern library to prepare for planned component re-implementations.

It includes these changes:

* Navigation and routing: adds support for tertiary navigation (allows for component grouping) and placeholder navigation
* Layout, spacing, styling: Make component documentation easier to read and compose in a non-fussy way
* Landing-page: Update content for clarity

These changes have been tested against pattern-library usage in the `client` and `lms` to verify backwards compatibility. My hope is that we can move quickly with little improvements to make this coming work more convenient, and then revisit the pattern library with a more critical eye when it's time to do so.

To test:

Run `make dev` and visit http://localhost:4001/


src/pattern-library/components/Library.js
```diff
@@ -3,12 +3,13 @@ import { toChildArray } from 'preact';
 import { useState } from 'preact/hooks';
 
 import { Frame } from '../../components/containers';
-import { LabeledButton } from '../../components/buttons';
 
 import { jsxToString } from '../util/jsx-to-string';
 
 /**
  * @typedef LibraryBaseProps
+ * @prop {import("preact").ComponentChildren|string} [intro] - Optional
```
lyzadanger, line 11: This prop addition is the only change to any of the exported components' APIs here. The rest of the changes in this module are styling and extracting internal subcomponents; an in-depth review is not required unless you're into it.

src/pattern-library/components/LibraryHome.js
```diff
@@ -5,35 +5,42 @@ export default function LibraryHome() {
   return (
     <Library.Page title="Pattern Library">
       <p>
-        This pattern library demonstrates foundations, patterns and components
-        that can be used in Hypothesis front-end applications.
+        This pattern library demonstrates UI toolkit components and how to use
```
lyzadanger, line 8: Content for this landing page will continue to become more meaningful through the component-update project.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -39,49 +41,129 @@ export default function PlaygroundApp({
...
+
+            <NavHeader>Legacy</NavHeader>
+            <NavSection title="Legacy Patterns">
+              <NavList routes={getRoutes('patterns')} />
+            </NavSection>
+            <NavSection title="Legacy Components">
+              <NavList routes={getRoutes('components')} />
+            </NavSection>
+
+            {extraRoutes.length > 0 && (
```
lyzadanger, line 157: Supports custom "pages" added by applications (`client`, `lms`).

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -2,11 +2,13 @@ import classnames from 'classnames';
 
 import { Link, SvgIcon } from '../../';
 
-import { getRoutes } from '../routes';
+import { componentGroups, getRoutes } from '../routes';
```
lyzadanger, line 5: Changes in this module support a third level of navigation (via `componentGroups`) and extract out some local navigation-related components.

src/pattern-library/components/patterns/ButtonComponents.js
```diff
@@ -4,30 +4,32 @@ import Library from '../Library';
...
-        In addition, a button may be in one of four states, which are
-        represented in their variant examples, left to right:
-      </p>
-      <ul className="hyp-list">
-        <li>Default</li>
-        <li>Pressed</li>
-        <li>Expanded</li>
-        <li>Disabled</li>
-      </ul>
+      <Library.Pattern title="Button variants">
```
lyzadanger, line 7: No substantive content changes here, just markup.

src/pattern-library/components/patterns/IconFoundations.js
```diff
@@ -15,7 +15,7 @@ export default function IconFoundations() {
   return (
     <Library.Page title="Icons">
       <Library.Pattern title="Hypothesis icon set">
-        <div className="grid grid-cols-8 gap-6">
+        <div className="grid grid-cols-3 md:grid-cols-6 gap-6">
```
lyzadanger, line 18: Quick fix to make sure the icons fit in the viewport

src/pattern-library/router.js
```diff
@@ -7,7 +7,7 @@ function routeFromCurrentURL(baseURL) {
 export function useRoute(baseURL, routes) {
   const [route, setRoute] = useState(() => routeFromCurrentURL(baseURL));
   const routeData = useMemo(
-    () => routes.find(r => route.match(r.route)),
+    () => routes.find(r => r.route && route.match(r.route)),
```
lyzadanger, line 10: Needed for placeholder nav entries, which don't have routes.

src/pattern-library/util/jsx-to-string.js
```diff
@@ -8,11 +8,11 @@ function escapeQuotes(str) {
...
-  if (typeof type === 'string') {
-    return type;
-  } else {
-    return type.displayName || type.name;
-  }
+  const name = typeof type === 'string' ? type : type.displayName ?? type.name;
+  // Handle (display)name conflicts if there are two components with the same
+  // name. e.g. if there are two components named `Foo`, the second of those
+  // encountered will have a name of `Foo$1`. Strip the `$1` in this case.
+  return name.replace(/\$[0-9]+$/, '');
```
lyzadanger, line 15: This is real quick-and-dirty, but, it works, and is only needed until we don't have duplicate components with the same names in the package.

src/pattern-library/util/jsx-to-string.js
```diff
@@ -66,8 +66,16 @@ export function jsxToString(vnode) {
           return value ? name : '';
         }
 
-        const valueStr =
-          typeof value === 'string' ? `"${escapeQuotes(value)}"` : `{${value}}`;
+        let valueStr;
```
lyzadanger, line 69: Ran into some challenges with `Symbol` props and `vNode`/component props when spiking.

robertknight:
I noticed in testing this that Preact devtools don't currently work with the pattern library. That's because there is no `import 'preact/debug'` statement in the entry file. See examples in the client's entry points. In the client we have some code in the Rollup config to exclude this import in production. For the pattern library we could do the same thing, although we could also afford to skip this optimization.

src/pattern-library/components/Library.js
```diff
@@ -151,40 +196,40 @@ function Demo({ children, withSource = false, style = {}, title }) {
...
+        </div>
+        <div className="flex flex-row items-center justify-end gap-x-4">
+          {withSource && (
+            <>
+              <DemoButton
+                onClick={() => setVisibleTab('demo')}
+                pressed={visibleTab === 'demo'}
+              >
+                Demo
+              </DemoButton>
```
robertknight, line 212: If `withSource` is false, we don't render the "Demo" OR "Source" buttons. I assume this is because there is only one tab in that case, so it is redundant. Previously a "Demo" tab title was rendered in both cases.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -39,49 +41,129 @@ export default function PlaygroundApp({
...
+   *   @param {PlaygroundRoute['route']} props.route
+   */
+  function NavLink({ title, route }) {
+    return (
+      <li className="-ml-[2px]">
+        {route && (
+          <Link
+            classes={classnames('pl-4 w-full border-l-2 hover:border-l-brand', {
+              'border-l-2 border-brand font-semibold':
+                activeRoute?.route === route,
```
robertknight, line 70: The nullability of `activeRoute` made me wonder what the pattern library's 404 page looks like (the case when there is no matching route). It's a very low priority task, but it could use some polish.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -39,49 +41,129 @@ export default function PlaygroundApp({
...
+              'border-transparent': activeRoute?.route !== route,
+            })}
+            href={`${route}`}
```
robertknight, line 73:
```suggestion
            href={route}
```

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -39,49 +41,129 @@ export default function PlaygroundApp({
...
+  }
 
-  if (extraRoutes.length) {
-    routeGroups.push({ title: 'Application Patterns', routes: extraRoutes });
+  /**
+   * A single navigation link
+   *
+   * @param {object} props
+   *   @param {string} props.title
+   *   @param {PlaygroundRoute['route']} props.route
```
robertknight, line 61: If you're going to tie the `route` field to the `PlaygroundRoute` type, you could just define it as `@param {PlaygroundRoute} props.route` and then get the title value from the route itself.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -39,49 +41,129 @@ export default function PlaygroundApp({
...
+      <div className="md:grid md:grid-cols-[2fr_5fr]">
+        <div className="md:h-screen md:sticky md:top-0 overflow-scroll">
+          <div className="md:sticky md:top-0 h-16 px-6 flex items-center bg-slate-0 border-b">
+            <Link
+              href={baseURL}
+              classes="grow flex gap-x-2 text-lg"
+              onClick={e => navigate(e, '/')}
+            >
+              <SvgIcon name="logo" />
+              Component Library
```
robertknight, line 129: Should this be a heading?

src/pattern-library/components/patterns/ButtonComponents.js
```diff
@@ -4,30 +4,32 @@ import Library from '../Library';
...
+          <li>Default</li>
+          <li>
+            Primary: For indicating that a button represents a primary action
+          </li>
+          <li>Dark: For rendering a button on a grey background</li>
+          <li>
+            Light: This variant should only be used for non-critical buttons on
+            white backgrounds (low contrast).
+          </li>
+        </ul>
```
robertknight, line 21: The layout of the demos, below this text, of the variants looks wrong / not what is intended:

Button variants

src/pattern-library/components/patterns/ButtonComponents.js
```diff
@@ -4,30 +4,32 @@ import Library from '../Library';
...
+        </p>
+        <ul className="list-disc list-inside">
+          <li>Default</li>
+          <li>
+            Primary: For indicating that a button represents a primary action
+          </li>
+          <li>Dark: For rendering a button on a grey background</li>
+          <li>
+            Light: This variant should only be used for non-critical buttons on
+            white backgrounds (low contrast).
```
robertknight, line 19: I presume "low contrast" means, "reduced contrast compared to primary action" rather than "low contrast" in absolute terms (which could present an a11y issue).

robertknight (approved):
LGTM. I added some minor notes and questions as I looked through.

src/pattern-library/components/Library.js
```diff
@@ -99,26 +119,51 @@ function Example({ children, title, variant = 'split' }) {
...
+ * @typedef DemoButtonProps
+ * @prop {import("preact").ComponentChildren} [children]
+ * @prop {() => void} [onClick]
+ * @prop {boolean} pressed
+ */
+
+/**
+ *
+ * @param {DemoButtonProps} props
+ * @returns
```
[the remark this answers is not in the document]
lyzadanger: D'oh. I've gotta research if I can tame my editor's integration; it auto-populates a bunch of tags for me and sometimes I don't spot the lingering `@returns`

src/pattern-library/components/Library.js
lyzadanger: Correct. This is meant to be an enhancement.

src/pattern-library/components/PlaygroundApp.js
lyzadanger: Noted.

src/pattern-library/components/PlaygroundApp.js
lyzadanger: Actually, this is necessary (effectively a `toString()`) to avoid a type gripe because `route` could ostensibly be a RexExp: 

image

src/pattern-library/components/patterns/ButtonComponents.js
lyzadanger: It's saying don't use it on anything but white because it _will_ be an a11y issue. This is gonna get revisited when I update the button components...

lyzadanger:
I've got `preact/debug` as a TODO.

Issue #455, opened 2022-06-16
Epic: Re-implement components


lyzadanger:
Re-implement existing components or otherwise provide equivalent functionality with updated conventions and API.

This issue serves as a burn-down list.

* [x] `LabeledButton` -> name change to `Button`
* [x] `IconButton`
* [x] `LinkButton`
* [x] ~~`Frame`~~ Use `<Card variant="flat" />` instead
* [x] `Card`
* [x] ~~`Actions`~~ -> will not directly re-implement
* [x] `Scrollbox`
* [ ] `Dialog`
* [x] `Icon`
* [x] `Modal`
* [ ] `ConfirmModal`
* [x] `Checkbox`
* [x] ~~`LabeledCheckbox`~~ (Supported by updated `Checkbox`)
* [x] `TextInput` --> `Input`
* [x] `TextInputWithButton` --> `InputGroup`
* [x] `Link`
* [x] `Panel`
* [x] `Spinner`
* [x] `FullScreenSpinner`
* [x] `Table`
* [x] `Thumbnail`

Notes/tasks:

### Card

* Current styles for `Card` include an `.is-focused` class. This should be changed to an appropriate prop.
* Current styles for `Card` handle for `theme-clean` affordance. This should be done in the `client` instead.
* Default padding for `Card` may be better at `p-3` instead of `p-4`.

### Buttons

* Our shared button component styling asserts cursor styling (pointer). If this is necessary, update the disabled styling to not use a cursor, as it is misleading.
* There are no built-in tailwind variants for aria- attributes, which is how we currently style "active" (pressed, expanded) buttons. We'll need an approach for styling here.

lyzadanger:
At this time, we are not providing an updated `Dialog` or `ConfirmModal` component. No applications were using `Dialog` directly, so it won't be missed. `ConfirmModal` is easily achieved through usage of `Modal`, and has already been adjusted as such in the `client`.

PR #491, opened 2022-07-13
Add re-implemented `Link` components

lyzadanger:
This PR re-implements the `Link` component and adds a `LinkUnstyled` component. This is the first component to undergo reimplementation as part of #455 

The first several commits take care of some pattern-library housekeeping and styling. These should only require a light review. Because this work is text- and markup-intensive — a large part of it is documentation — the diff looks large, but the actual functional changes are smallish and can be found easily by navigating commit by commit.

55072aa4253be63d62f1877adf51be61e8e17fa8 is the main implementation commit and 52f3796aa219e8fa991f3469ea7c5a931119fdef adds pattern-library documentation for `Link` and `LinkUnstyled`.

_Updated_: I pushed an additional commit to deprecate the legacy `Link` and remove the `Link` pattern. Patterns aren't part of the API since v5.0 (i.e. we're not shipping the styles that would support them), so removal is non-breaking.

`Link` is a simple component, but it is a good candidate for an unstyled component (`LinkUnstyled`), and a base component (`LinkBase`, not part of the package API). _This is as complicated as component "inheritance" should get in the current model_.

### See it in play

Check out this branch, run `make dev` and visit http://localhost:4001/navigation-link

You should see something like:

image

### TODO

* [x] Test against `client` and see if new components serve well where `Link` legacy is being used. (I want to do this in parallel with getting this out for review).

### Out of scope, but tracked

* DRYing out tests and extracting reusable tests
* DRYing out anything else (boilerplate, e.g.)
* Documentation for how to use components and the package in a broader sense than just this component, or any other package-level documentation


src/components/navigation/Link.js
```diff
@@ -0,0 +1,56 @@
...
+  elementRef,
+
+  underline = 'none',
+  color = 'brand',
+
+  ...htmlAttributes
+}) {
+  return (
+    <LinkBase
+      {...htmlAttributes}
```
lyzadanger, line 33: Note that HTML attributes are applied in components consistently before the attributes that the component controls. In the past we've been inconsistent about this.

src/components/navigation/Link.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+export default function Link({
+  children,
+  classes,
+  elementRef,
+
+  underline = 'none',
+  color = 'brand',
+
+  ...htmlAttributes
```
lyzadanger, line 29: Because it can become a world of hurt to have rest props that may include properties from multiple sources, I've established the convention that only HTML attributes should be rest props, and the other props expanded out, even if it's just to essentially forward them on. I've also ordered props by their source and put line breaks between sources. I find this easier to reason about...

src/components/navigation/LinkBase.js
```diff
@@ -0,0 +1,30 @@
...
+ * @typedef {import('../../types').BaseProps} BaseProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLAnchorElement>} HTMLAnchorAttributes
+ */
+
+/**
+ * Base component for Link components. Applies common attributes.
+ *
+ * @param {BaseProps & HTMLAnchorAttributes} props
+ */
+export default function LinkBase({
```
lyzadanger, line 13: So, `LinkBase` is responsible for attributes (or any other common functionality) and the components that use it are handling styling.

src/components/navigation/LinkUnstyled.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Render a link (`<a>`) with common attributes but no styling
+ *
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLAnchorElement>} HTMLAnchorAttributes
+ *
+ * @param {CommonProps & HTMLAnchorAttributes} props
+ */
+export default function LinkUnstyled({
```
lyzadanger, line 15: Basically a wrapper around `LinkBase` but disallowing `className` (so, shaped like the presentational component it is).

src/components/navigation/test/Link-test.js
```diff
@@ -0,0 +1,77 @@
...
+describe('Link', () => {
+  const createComponent = (props = {}) => {
+    return mount(
+      <Link href="http://www.example.com" {...props}>
+        This is content inside of a Link
+      </Link>
+    );
+  };
+
+  // Tests common to presentational components
```
lyzadanger, line 15: All tests in this body of work are unrolled and wet, but I have noted groups of tests for presumable extraction.

lyzadanger (commented):
Apparently a main body comment for a review is no longer optional. :)

src/pattern-library/components/LibraryNext.js
```diff
@@ -0,0 +1,87 @@
...
+ * Render import "usage" of a given `componentName`
+ *
+ * @param {object} props
+ *   @prop {string} componentName
+ *   @prop {'next'|'legacy'} [generation]
+ */
+function Usage({ componentName, generation = 'next' }) {
+  const importPath =
+    generation === 'next'
+      ? '@hypothesis/frontend-shared/lib/next'
```
lyzadanger, line 70: Derp. I had proposed `@hypothesis/frontend-shared/next` but that was before I remembered how our current build works. I'm not convinced it's worth the trouble to do the re-futzing necessary to make that path work; the extra `/lib/` for this interim entrypoint doesn't bother me too much.

src/components/navigation/Link.js
robertknight: > I've established the convention that only HTML attributes should be rest props, and the other props expanded out, even if it's just to essentially forward them on.

Makes sense.

src/next.js
```diff
@@ -0,0 +1 @@
+export * from './components/navigation';
```
robertknight, line 1: I tend to avoid `export *` on the basis that it is useful to be able to look at a file and see exactly what it exports. It also makes it easier to see when new things are added/removed from the exported interface when looking through the history.

robertknight (commented):
I had an initial look through this and added some comments. I think the overall approach makes sense.

src/components/navigation/Link.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+/**
+ * Styled component for a link (`<a>` element).
+ *
+ * @param {CommonProps & LinkProps & HTMLAnchorAttributes} props
+ */
+export default function Link({
+  children,
+  classes,
```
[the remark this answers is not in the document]
lyzadanger: Please see https://github.com/hypothesis/frontend-shared/pull/492 which should I hope answer this question thoroughly.

src/next.js
lyzadanger: Ah, yes, I totally see that perspective! Happy to expand this, and to remove the somewhat crufty `src/components/navigation/index` module as it doesn't serve a purpose, really. It was hanging around from when I thought I'd be grouping components more "aggressively."

Note I'm proposing still using `export *` for _icons_ because there are so many of them and it's ergonomic to auto-generate a local `index` in the icons directory (see #494)

src/pattern-library/components/LibraryNext.js
```diff
@@ -0,0 +1,87 @@
...
+      <code>
+        {jsxToString(`
+import { ${componentName} } from '${importPath}';
+`)}
+      </code>
+    </div>
+  );
+}
+
+export default {
```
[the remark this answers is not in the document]
lyzadanger: Yeah, this has been on my long list for a little while...

src/components/navigation/test/Link-test.js
```diff
@@ -0,0 +1,77 @@
...
+  it('applies extra classes', () => {
+    const wrapper = createComponent({ classes: 'foo bar' });
+    const allClasses = [...wrapper.find('a').getDOMNode().classList.values()];
+    assert.equal(allClasses[allClasses.length - 2], 'foo');
+    assert.equal(allClasses[allClasses.length - 1], 'bar');
```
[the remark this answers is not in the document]
lyzadanger: Gotcha. I'd like to tighten up tests when I extract common tests. I shall do that here.

robertknight:
I think we need to provide stronger guidance around when underlines must be used in links vs when they are optional. My suggestion would be that they are optional when the link has sufficient other non-color cues. Color can serve as an emphasis of link-ness, but can't be the only cue. Consider the screenshot below:

Link underline hints

Here I think it is acceptable that annotation timestamps, "About this version ->" and Help Topics don't have underlines, since they have sufficient other cues:

1. Arrow glyphs ("About this version")
2. Font size/weight ("Help Topics")
3. Convention (In forum posts, the timestamp is usually a link) 
4. Position (isolated pieces of text not surrounded by non-link text) 

The text "visible only to you" and "join link" links on the other hand really do need underlines because of their position in the middle of non-linkable sentences of text.

Applying these principles to the examples added here, all the links in the text "Content that contains a link to somewhere ..." do need underlines. The standalone links don't.

src/pattern-library/components/LibraryNext.js
```diff
@@ -0,0 +1,87 @@
...
+      {status === 'added' && <span>Added:</span>}
+    </span>
+  );
+}
+
+/**
+ * Wrapper around a list of changelog items
+ *
+ * @param {object} props
+ *   @param {import('preact').ComponentChildren}  props.children
```
robertknight, line 38:
```suggestion
 *   @param {import('preact').ComponentChildren} props.children
```

src/pattern-library/components/LibraryNext.js
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @param {object} props
+ *   @param {ChangeStatus} props.status
+ *   @param {import('preact').ComponentChildren} props.children
+ */
+function ChangelogItem({ status, children }) {
+  return (
+    <li className="space-x-2">
+      <StatusChip status={status} />
+      <span>{children}</span>
```
robertknight, line 55: Is the `<span>` needed here?

src/pattern-library/components/LibraryNext.js
robertknight: OK.

robertknight (commented):
I think this is good to go except for the issue with underline-less links in the middle of text. For links that use underline only on hover, I suggest we find some examples that are more representative of use cases where that would be acceptable (eg. timestamp or isolated piece of text that has other click-ability cues).

src/components/navigation/LinkBase.js
```diff
@@ -0,0 +1,30 @@
...
+  children,
+  className,
+  elementRef,
+
+  ...htmlAttributes
+}) {
+  return (
+    <a
+      {...htmlAttributes}
```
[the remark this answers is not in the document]
lyzadanger: Right: it's not the rule that the HTML attributes rest props will _always_ come first, but it will probably be the case frequently. I think it should be the rough default case, as it avoids ambiguity of precedence for any component-controlled attributes. And exceptions would be documented or commented. Does that sound about right?

src/components/navigation/test/Link-test.js
```diff
@@ -0,0 +1,75 @@
...
+    const wrapperOuterEl = createComponent({ 'data-testid': 'foo-container' })
+      .childAt(0)
+      .getDOMNode();
+
+    assert.isTrue(wrapperOuterEl.hasAttribute('data-testid'));
+    assert.equal(wrapperOuterEl.getAttribute('data-testid'), 'foo-container');
+  });
+
+  // Tests specific to this component
+  it('styles link color', () => {
```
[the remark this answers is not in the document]
lyzadanger: Expect test cleanup in a separate PR within the next several days. I think these are acceptable as-is to land.

robertknight (approved):
LGTM. On a pedantic note in the client "Log in" and "Sign up" links are actually buttons, but these are reasonable examples of things that could be links in other contexts.

lyzadanger:
> LGTM. On a pedantic note in the client "Log in" and "Sign up" links are actually buttons, but these are reasonable examples of things that could be links in other contexts.

DARN IT! I was hoping you wouldn't be so observant. But here's the thing: it's still a relevant pattern because those components are actually `LinkButton`s — there should be harmony between `Link` and `LinkButton`, IMO.

PR #492, opened 2022-07-14
Pattern library: Add package usage docs and customization guide

lyzadanger:
This PR adds some pattern-library content for how to install/configure the package and how to customize components. It also adds a placeholder for a future page to describe the common API of components.

Depends on #491 (for pattern-library components added in that PR).

### Check it out...

Check out this branch and run `make dev` and visit http://localhost:4001/getting-started or http://localhost:4001/customizing-components

Screen shots:

Screen Shot 2022-07-14 at 3 50 21 PM

image

### DIFF
--- a/src/pattern-library/components/LibraryNext.js
+++ b/src/pattern-library/components/LibraryNext.js
@@ -58,6 +58,37 @@ function ChangelogItem({ status, children }) {
 }
 
 /**
+ * Render provided `content` as a "code block" example.
+ *
+ * @param {object} props
+ *   @param {string} props.content - Code content
+ *   @param {'sm'|'md'|'lg'} [props.size]
+ *   @param {string} [props.title] - Caption (e.g. filename, description) of
+ *     code block
+ */
+function Code({ content, size, title }) {
+  return (
+    <figure className="space-y-2">
+      <div
+        className={classnames(
+          'unstyled-text bg-slate-7 text-color-text-inverted p-4 rounded-md',
+          { 'text-sm': size === 'sm' }
+        )}
+      >
+        <code className="text-color-text-inverted">
+          <pre className="whitespace-pre-wrap">{jsxToString(content)}</pre>
+        </code>
+      </div>
+      {title && (
+        <figcaption className="flex justify-end">
+          <span className="italic">{title}</span>
+        </figcaption>
+      )}
+    </figure>
+  );
+}
+
+/**
  * Render import "usage" of a given `componentName`
  *
  * @param {object} props
@@ -70,18 +101,16 @@ function Usage({ componentName, generation = 'next' }) {
       ? '@hypothesis/frontend-shared/lib/next'
       : '@hypothesis/frontend-shared';
   return (
-    <div className="unstyled-text bg-slate-7 text-color-text-inverted p-4 rounded-md">
-      <code>
-        {jsxToString(`
-import { ${componentName} } from '${importPath}';
-`)}
-      </code>
-    </div>
+    <Code
+      content={`import { ${componentName} } from '${importPath}';
+`}
+    />
   );
 }
 
 export default {
   Changelog,
   ChangelogItem,
+  Code,
   Usage,
 };
--- a/styles/library.scss
+++ b/styles/library.scss
@@ -11,7 +11,11 @@
     }
 
     :where(ul):not(:where([class~='unstyled-text'] *)) {
-      @apply list-disc list-outside pl-4 space-y-2;
+      @apply list-disc list-outside ml-8 space-y-2;
+    }
+
+    :where(ol):not(:where([class~='unstyled-text'] *)) {
+      @apply list-decimal list-outside ml-8 space-y-2;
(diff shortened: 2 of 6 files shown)


robertknight:
While reviewing this I noticed that the contrast for body text felt a little weak to me. Plugging the body text color (rgb(89, 89, 105) == #595969) values into https://webaim.org/resources/contrastchecker/ I get a ratio of 6.86:1, which is indeed below the recommended range for normal text (although still WCAG AA legal).

styles/library.scss
```diff
@@ -11,7 +11,11 @@
     }
 
     :where(ul):not(:where([class~='unstyled-text'] *)) {
-      @apply list-disc list-outside pl-4 space-y-2;
+      @apply list-disc list-outside ml-8 space-y-2;
+    }
+
+    :where(ol):not(:where([class~='unstyled-text'] *)) {
```
robertknight, line 17: Is `:where` being used here for its specificity behavior?

src/pattern-library/components/LibraryNext.js
```diff
@@ -57,6 +57,35 @@ function ChangelogItem({ status, children }) {
...
+ */
+function Code({ content, size, title }) {
+  return (
+    <div className="space-y-2">
+      <div
+        className={classnames(
+          'unstyled-text bg-slate-7 text-color-text-inverted p-4 rounded-md',
+          { 'text-sm': size === 'sm' }
+        )}
+      >
```
robertknight, line 75: What is the reason for the wrapper `<div>` as opposed to styling the `<code>` itself?

robertknight (approved):
I had some suggestions around the `Code` component, otherwise LGTM.

src/pattern-library/components/LibraryNext.js
```diff
@@ -57,6 +57,35 @@ function ChangelogItem({ status, children }) {
   );
 }
 
+/**
+ *
+ * @param {object} props
+ *   @param {string} props.content
+ *   @param {'sm'|'md'|'lg'} [props.size]
```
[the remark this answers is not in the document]
lyzadanger: Good feedback. I'll rethink.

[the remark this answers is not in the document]
lyzadanger: @robertknight 

Interesting food for thought here. In terms of "how the other guys do it", it's split (this survey was specific to `Button` components so is not exhaustive for sizing across the entire library):

* MUI: `small`, `medium`, `large`
* Mantine: `sm`, `md`, `lg`
* Ant: `large`, `small` (oddly, no value for the default size. Weird)
* NextUI: `sm`, `md`, `lg`, `xl`
* Reactstrap: `sm`, `lg`
* Chakra: `sm`, `md`, `lg`, `xl`

I almost just went and switched to `small`, `medium` and `large`, especially since MUI does it and they are most closely aligned to our overall approach. But I was swayed by two factors:

* Ergonomics: It's just a bit faster to use (standardized and well-recognized) abbreviations
* More relevantly, systematized-ness (oy, not a word): The abbreviations are easier to reason about if we ever got beyond `sm`, `md`, `lg`. It's conventionally obvious that the thing after `lg` would be `xl`. But it's less conventionally obvious what the thing after `large` is. I don't know whether we'd ever need sizes beyond the these three in real life, but in the end I have a preference for `sm`, `md`, `lg` because it's more systematized.

styles/library.scss
lyzadanger: Yep.

src/pattern-library/components/LibraryNext.js
```diff
@@ -57,6 +57,35 @@ function ChangelogItem({ status, children }) {
...
+        <code>
+          <pre className="whitespace-pre-wrap">{jsxToString(content)}</pre>
+        </code>
+      </div>
+      {title && (
+        <div className="flex justify-end">
+          <span className="italic">{title}</span>
+        </div>
+      )}
+    </div>
```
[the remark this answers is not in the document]
lyzadanger: Great idea!

src/pattern-library/components/LibraryNext.js
lyzadanger: `code` is an inline element, so it felt more appropriate to have a block container around it here.

lyzadanger:
> While reviewing this I noticed that the contrast for body text felt a little weak to me. Plugging the body text color (rgb(89, 89, 105) == #595969) values into https://webaim.org/resources/contrastchecker/ I get a ratio of 6.86:1, which is indeed below the recommended range for normal text (although still WCAG AA legal).

Well, aren't _you_ eagle-eyed 🦅 ?

It was SO CLOSE. 1:6.87. I tweaked it by a tiny amount to get to 1:7.07. I find the current level of contrast rather soothing, but if you thought it looked low-contrast when it was so, so close to being AAA-OK...hmmm...I am rather fond of this text color, however, so I'm going to leave it as-is unless it's a high-intensity opinion that we should change it...

PR #506, opened 2022-07-22
Add Spinner (simple) component

lyzadanger:
This PR adds a reimplemented simple `Spinner` component. A "simple" component is one that does not offer any common props (cf. presentational or composite components).

Depends on #505 

For documentation about component "categories" and their props API:  See #508 

### Testing

Check out this branch, run `make dev` and visit http://localhost:4001/feedback-spinner in a browser.

image

### Component checklist

- [x] component tests
- [x] exports in package
- [x] remove any associated patterns: Removed pattern library page for these patterns, but SASS is a dependency for the legacy Thumbnail component and pattern


src/components/feedback/Spinner.js
```diff
@@ -0,0 +1,33 @@
+import classnames from 'classnames';
+
+import { SpinnerSpokesIcon } from '../icons';
+
+/**
+ *
+ * @typedef SpinnerProps
+ * @prop {'sm'|'md'|'lg'} [size='sm']
```
lyzadanger, line 8: This is called out in the pattern-library documentation changelog, but I've changed default sizing from medium to small because in most of our usage of spinners that's the size we're using.

src/components/test/common-tests.js
```diff
@@ -79,3 +83,17 @@ export function testBaseComponent(Component) {
...
 }
+
+/**
+ * Common tests for simple design components
+ *
+ * @param {FunctionComponent} Component
+ */
+export function testSimpleComponent(Component) {
+  describe('Common simple component functionality', () => {
+    it('renders', () => {
```
lyzadanger, line 94: "Look, I'm alive" test, as discussed.

src/components/feedback/Spinner.js
```diff
@@ -0,0 +1,33 @@
+import classnames from 'classnames';
+
+import { SpinnerSpokesIcon } from '../icons';
+
+/**
+ *
+ * @typedef SpinnerProps
```
robertknight, line 7:
```suggestion
/**
 * @typedef SpinnerProps
```

src/pattern-library/components/patterns/feedback/Spinner.js
```diff
@@ -0,0 +1,86 @@
...
+
+export default function SpinnerPage() {
+  return (
+    <Library.Page
+      title="Spinner"
+      intro={
+        <p>
+          <code>Spinner</code> is a simple component to render a spinner,
+          representing loading or other progress. To customize, use the{' '}
+          <code>SpinnerIcon</code> component instead.
```
robertknight, line 13: There isn't a single component called `SpinnerIcon`, but rather two separate `Spinner<Variant>Icon` components.

src/pattern-library/components/patterns/feedback/Spinner.js
```diff
@@ -0,0 +1,86 @@
...
+      <Library.Pattern title="Size">
+        <Library.Demo title="size: 'sm' (1em) (default)" withSource>
+          <Spinner size="sm" />
+        </Library.Demo>
+        <Library.Demo title="size: 'md' (2em)" withSource>
+          <Spinner size="md" />
+        </Library.Demo>
+        <Library.Demo title="size: 'lg' (4em)" withSource>
+          <Spinner size="lg" />
+        </Library.Demo>
```
robertknight, line 82: Are these all based on existing sizes that we actually use, or just chosen to fit a particular scale?

src/tailwind.preset.js
```diff
@@ -87,6 +87,8 @@ export default /** @type {Partial<import('tailwindcss').Config>} */ ({
       },
       spacing: {
         em: '1em',
+        '2em': '2em',
+        '4em': '4em',
```
robertknight, line 91: Does Tailwind not have a way to specify em-multiples like this without declaring all the multiples that you use?

robertknight (changes requested):
Found a small documentation issue with a reference to a component that doesn't exist. Also had a query about the em-scale sizes.

src/tailwind.preset.js
lyzadanger: Tailwind's default sizing scale is `rem`-based. Most of our sizing uses that and doesn't require us to add values to the configuration. We need `em-`based sizing in just a few places, and I feel that being explicit about that in the preset values is actually helpful. It's likely this is the end of the line for em-unit sizing that we'll need.

There certainly are other options for doing this if it bugs yah.

src/pattern-library/components/patterns/feedback/Spinner.js
lyzadanger: Good catch, thanks.

src/pattern-library/components/patterns/feedback/Spinner.js
lyzadanger: Yes. These are identical to the legacy `Spinner` and the different sizes are used — at least I'm certain that `sm` and `lg` are. `lg` is used for the full-screen spinner, e.g., which hasn't yet been updated.

robertknight (approved):

PR #507, opened 2022-07-22
Add `Scroll` (presentational) components and `ScrollBox` (composite) component

lyzadanger:
This PR adds new components: `Scroll`, `ScrollContainer`, `ScrollContent`, and `ScrollBox`. `Scroll`, `ScrollContainer` and `ScrollContent` are presentational components, which are composed by the composite component `ScrollBox`, which is meant to replace the legacy `Scrollbox` (note casing difference) component.

_Updated_: Pushed a couple additional commits to fix a bug with route matching and to put the `ScrollBox` link before `Scroll` in the nav for now. There's still some more thinking about how to arrange navigation best, but I'm kind of falling into a rathole with it so am going to put aside for now.

Depends on #505

### Testing

Check out this branch, run `make dev` and visit http://localhost:4001/data-scroll and http://localhost:4001/data-scrollbox

image

image

### Checklist

- [x] component tests
- [x] exports in package

### DIFF
--- a/src/components/data/Scroll.js
+++ b/src/components/data/Scroll.js
@@ -30,8 +30,8 @@ export default function Scroll({
       {...htmlAttributes}
       ref={downcastRef(elementRef)}
       className={classnames(
-        // Prevent vertical flex or grid blowout: keep this contained vertically
-        // to parent by setting a min width
+        // Prevent overflow by overriding `min-height: auto`.
+        // See https://stackoverflow.com/a/66689926/434243.
         'min-h-0',
         'h-full w-full overflow-auto',
         { 'scroll-shadows': variant === 'raised' },
--- a/src/components/data/Scroll.js
+++ b/src/components/data/Scroll.js
@@ -1,0 +1,44 @@
+import classnames from 'classnames';
+
+import { downcastRef } from '../../util/typing';
+
+/**
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLElement>} HTMLAttributes
+ *
+ * @typedef ScrollProps
+ * @prop {'raised'|'flat'} [variant='raised'] - Render with scroll-hinting
+ *   shadows ('raised') or without ('flat')
+ */
+
+/**
+ * Render a fluid container that scrolls on overflow.
+ *
+ * @param {CommonProps & ScrollProps & HTMLAttributes} props
+ */
+export default function Scroll({
+  children,
+  classes,
+  elementRef,
+
+  variant = 'raised',
+
+  ...htmlAttributes
+}) {
+  return (
+    <div
+      {...htmlAttributes}
+      ref={downcastRef(elementRef)}
+      className={classnames(
+        // Prevent vertical flex or grid blowout: keep this contained vertically
+        // to parent by setting a min width
+        'min-h-0',
+        'h-full w-full overflow-auto',
+        { 'scroll-shadows': variant === 'raised' },
+        classes
+      )}
+    >
+      {children}
+    </div>
+  );
+}
--- a/src/components/data/ScrollContainer.js
+++ b/src/components/data/ScrollContainer.js
@@ -1,0 +1,41 @@
+import classnames from 'classnames';
+
+import { downcastRef } from '../../util/typing';
+
+/**
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ *
+ * @typedef ScrollContainerProps
+ * @prop {boolean} [borderless=false] - Remove border around container
+ */
+
+/**
+ * Constrain children (which may include both scrollable and non-scrolling
+ * content) to the dimensions of the immediate parent.
+ *
+ * @param {CommonProps & ScrollContainerProps & HTMLDivAttributes} props
+ */
+export default function ScrollContainer({
+  children,
+  classes,
+  elementRef,
+
+  borderless = false,
+
+  ...htmlAttributes
+}) {
+  return (
+    <div
+      {...htmlAttributes}
+      ref={downcastRef(elementRef)}
+      className={classnames(
+        'flex flex-col min-h-0 h-full w-full',
+        { border: !borderless },
+        classes
+      )}
+    >
+      {children}
+    </div>
+  );
+}
--- a/src/components/data/ScrollContainer.js
+++ b/src/components/data/ScrollContainer.js
@@ -30,7 +30,10 @@ export default function ScrollContainer({
       {...htmlAttributes}
       ref={downcastRef(elementRef)}
       className={classnames(
-        'flex flex-col min-h-0 h-full w-full',
+        'flex flex-colh-full w-full',
+        // Prevent overflow by overriding `min-height: auto`.
+        // See https://stackoverflow.com/a/66689926/434243.
+        'min-h-0',
         { border: !borderless },
         classes
       )}
--- a/src/components/data/ScrollContent.js
+++ b/src/components/data/ScrollContent.js
@@ -1,0 +1,30 @@
+import classnames from 'classnames';
+
+import { downcastRef } from '../../util/typing';
+
+/**
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ */
+
+/**
+ * Apply consistent padding and spacing to content within a Scroll
+ *
+ * @param {CommonProps & HTMLDivAttributes} props
+ */
+export default function ScrollContent({
+  children,
+  classes,
+  elementRef,
+  ...htmlAttributes
+}) {
+  return (
+    <div
+      {...htmlAttributes}
+      ref={downcastRef(elementRef)}
+      className={classnames('px-3 py-2', classes)}
+    >
+      {children}
+    </div>
+  );
+}
--- a/src/components/data/test/ScrollBox-test.js
+++ b/src/components/data/test/ScrollBox-test.js
@@ -1,0 +1,21 @@
+import { mount } from 'enzyme';
+
+import { testCompositeComponent } from '../../test/common-tests';
+
+import ScrollBox from '../ScrollBox';
+
+const createComponent = (Component, props = {}) => {
+  return mount(<Component {...props}>This is child content</Component>);
+};
+
+describe('ScrollBox', () => {
+  testCompositeComponent(ScrollBox);
+
+  it('disables container border when `borderless` is set', () => {
+    const withBorder = createComponent(ScrollBox);
+    const withoutBorder = createComponent(ScrollBox, { borderless: true });
+
+    assert.isFalse(withBorder.find('ScrollContainer').props().borderless);
+    assert.isTrue(withoutBorder.find('ScrollContainer').props().borderless);
+  });
+});
--- a/src/pattern-library/components/patterns/data/Scroll.js
+++ b/src/pattern-library/components/patterns/data/Scroll.js
@@ -1,0 +1,188 @@
+import { Scroll, ScrollContainer, ScrollContent } from '../../../../next';
+import Library from '../../Library';
+import Next from '../../LibraryNext';
+
+import { SampleListElements } from '../samples';
+
+export default function ScrollPage() {
+  return (
+    <Library.Page
+      title="Scroll"
+      intro={
+        <>
+          <p>
+            The <code>Scroll</code> family of presentational components provide
+            support for customized scrollable content in cases that{' '}
+            <code>ScrollBox</code> does not satisfy a use case.
+          </p>
+          <ul>
+            <li>
+              <code>ScrollContainer</code>: Contains its content — a mix of
+              scrollable and (optionally) non-scrollable content — to the
+              dimensions of a parent element
+            </li>
+            <li>
+              <code>Scroll</code>: A layout container that scrolls on overflow,
+              with shadows for scroll-hinting
+            </li>
+            <li>
+              <code>ScrollContent</code>: Applies consistent spacing and padding
+              to content in a <code>Scroll</code>
+            </li>
+          </ul>
+        </>
+      }
+    >
+      <Library.Pattern title="Status">
+        <p>
+          <code>Scroll</code>, <code>ScrollContainer</code>, and{' '}
+          <code>ScrollContent</code> are new components modeled after aspects of
+          the legacy <code>Scrollbox</code> component.
+        </p>
+      </Library.Pattern>
+
+      <Library.Pattern title="Usage">
+        <Next.Usage
+          componentName="Scroll, ScrollContainer, ScrollContent"
+          size="sm"
+        />
+        <Library.Example>
+          <p>
+            Sizing constraints are dictated by the immediate parent container of
+            the outermost <code>Scroll*</code> component. Here and in following
+            examples, the parent container sets width (<code>320px</code>) and
+            height (<code>200px</code>).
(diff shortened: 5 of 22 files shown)


src/tailwind.scroll-shadows.js
```diff
@@ -0,0 +1,62 @@
+import plugin from 'tailwindcss/plugin.js';
+
+/**
```
lyzadanger, line 3: Almost everything here is taken directly from [the SASS implementation](https://github.com/hypothesis/frontend-shared/blob/d5181845da9b4e5985ac90e09c25f86cdcc7ac12/styles/mixins/patterns/_containers.scss#L154-L229), but restructured for Tailwind.

src/components/data/Scroll.js
```diff
@@ -0,0 +1,42 @@
...
+import { downcastRef } from '../../util/typing';
+
+/**
+ *
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ *
+ * @typedef ScrollProps
+ * @prop {'raised'|'flat'} [variant='raised'] - Render with scroll-hinting
+ *   shadows ('raised') or without ('flat')
```
robertknight, line 12: Using the flat variant without a scroll-ability hint may be confusing unless there is some other affordance to indicate scrollability (assuming an OS that doesn't always render scrollbars). Do you know what the use cases we currently have for `flat` are?

src/pattern-library/components/patterns/data/Scroll.js
```diff
@@ -0,0 +1,187 @@
+import { Scroll, ScrollContainer, ScrollContent } from '../../../../next';
+import Library from '../../Library';
+import Next from '../../LibraryNext';
+
+import { SampleListElements } from '../samples';
+
+export default function ScrollPage() {
+  return (
+    <Library.Page
+      title="Scroll components"
```
robertknight, line 10: The `<title>` for this page reads "Scroll" after the entry in the navigation list rather than "Scroll components".

src/pattern-library/components/patterns/data/Scroll.js
```diff
@@ -0,0 +1,187 @@
...
+      intro={
+        <>
+          <p>
+            The <code>Scroll</code> family of presentational components provide
+            support for contained, scrollable content:
```
robertknight, line 15: It would be good to mention `ScrollBox` here as the recommended combination of these lower-level components.

src/pattern-library/components/patterns/data/Scroll.js
```diff
@@ -0,0 +1,187 @@
...
+            </li>
+          </ul>
+        </>
+      }
+    >
+      <Library.Pattern title="Status">
+        <p>
+          <code>Scroll</code>, <code>ScrollContainer</code>, and{' '}
+          <code>ScrollContent</code> are new components modeled after aspects of
+          the legacy <code>Scrollbox</code> component.
```
robertknight, line 39: Out of interest, can you expand on the rationale for the UI / API design changes from the old `Scrollbox` component to the three new ones?

src/pattern-library/router.js
```diff
@@ -6,13 +6,22 @@ function routeFromCurrentURL(baseURL) {
...
+  // Data associated with the currently-applied route
+  const routeData = useMemo(() => {
+    return routes.find(r => {
+      if (!r.route) {
+        return false;
+      }
+      if (typeof r.route === 'string') {
+        return r.route === route;
+      }
+      return r.route && route.match(r.route);
```
robertknight, line 19: Was this change to prevent the `/data-scroll` route from matching the `/data-scrollbox` route?

src/tailwind.preset.js
```diff
@@ -3,6 +3,7 @@ import plugin from 'tailwindcss/plugin.js';
 import colors from 'tailwindcss/colors.js';
 
 import focusVisibleRing from './tailwind.focus-visible-ring.js';
+import scrollShadows from './tailwind.scroll-shadows.js';
```
robertknight, line 6: Are we required to make these default exports here? If we could follow our usual convention of avoiding default exports except for Preact component modules with a single export, I think that would be preferable.

src/tailwind.scroll-shadows.js
```diff
@@ -0,0 +1,62 @@
...
+ *     that they are always positioned at the top and the bottom of the
+ *     _scrollbox_ frame. When these positions align with the positions of the
+ *     shadow covers--at the top and the bottom of the overflowing content--
+ *     they will be obscured by those shadow covers.
+ *
+ * See https://lea.verou.me/2012/04/background-attachment-local/
+ *
+ * Safari's behavior is different because of a bug with
+ * `background-attachment: local`.
+ * See https://bugs.webkit.org/show_bug.cgi?id=219324
```
robertknight, line 28: This bug is now marked as fixed. Reading through the comments it looks like this issue affects Safari 14 through to Safari 15.2 (Safari 15.3 was released after the fix landed, though I don't know for sure that it includes this change). Everything seems to work correctly in my testing in Safari 15.4. Do you want to check and confirm?

If so then we would note that only certain Safari versions are affected.

src/tailwind.scroll-shadows.js
```diff
@@ -0,0 +1,62 @@
+import plugin from 'tailwindcss/plugin.js';
+
+/**
+ * Utility class to show scroll-hint shadows at the top and bottom of a
+ * scrollable frame element if:
+ *  - The content height exceeds the frame height: i.e. can be scrolled, and
+ *  - The content is scrollable in the shadow's direction (up or down)
+ *
+ * Shadows are not visible once the frame has been scrolled all the way in the
```
robertknight, line 9: This may be copied and pasted from SASS, but such nice documentation ❤️

src/types.js
```diff
@@ -11,6 +11,13 @@
  *   outermost element.
  */
 
+/**
+ * Props common to components that are opinionated compositions of other
+ * components.
```
robertknight, line 16: Is this only for components which are "simple" combinations of other components? I expect we might in future have composite components which are somewhat more complex than say `ScrollBox`.

robertknight (commented):
I've had a look through and left some comments, questions and small corrections.

The main feedback I had was already given in Slack, which was about improving the discoverability of ScrollBox and its relationship with the sub-components.

src/components/data/test/ScrollBox-test.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { testCompositeComponent } from '../../test/common-tests';
+
+import ScrollBox from '../ScrollBox';
+
+const createComponent = (Component, props = {}) => {
+  return mount(<Component {...props}>This is child content</Component>);
+};
+
+describe('Scroll', () => {
```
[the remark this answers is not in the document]
lyzadanger: These c-n-p idiocies should be alleviated in future as I'm using a tool for scaffolding instead of copying and pasting.

src/pattern-library/components/patterns/data/Scroll.js
```diff
@@ -0,0 +1,187 @@
...
+        <>
+          <p>
+            The <code>Scroll</code> family of presentational components provide
+            support for contained, scrollable content:
+          </p>
+          <ul>
+            <li>
+              <code>ScrollContainer</code>: Contains its content — a mix of
+              scrollable and (optionally) non-scrollable content — to the
+              dimensions of a parent element
```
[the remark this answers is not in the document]
lyzadanger: `ScrollContainer` does two things:

* Sets a `flex` columnar layout that will prevent the vertical size of the content from escaping the bounds of the parent container. i.e. height of the children, combined, will be the height of the available space in the parent. `Scroll` without this has no constraints; it's just an element that has overflow set, so a `ScrollContainer` could have use without mixed scrollable/non-scrollable content, as well.
* Puts a border on it, unless told not to. This is minor, just implementation of a design pattern

src/components/data/ScrollContent.js
```diff
@@ -0,0 +1,31 @@
...
+  children,
+  classes,
+  elementRef,
+  ...htmlAttributes
+}) {
+  return (
+    <div
+      {...htmlAttributes}
+      ref={downcastRef(elementRef)}
+      className={classnames('px-3 py-2', classes)}
```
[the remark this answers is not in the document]
lyzadanger: This follows the composition pattern that is also used in `Card` (`CardContent`), and follows a preference for composition over configuration (props). This cribs from Material UI in part. As with `Card`, a `Scroll` without `ScrollContent` is useful if the author wanted full-bleed content inside the `Scroll` or other control over the layout of the contained content.

There is a use case, e.g., in LMS dialogs, where we want a contained mix of scrollable and non-scrollable content, but without padding around it, or borders. The breakdown of these components is both consistent with how I foresee other components being structured (again, e.g., `Card`) and flexible enough to allow for these use cases.

src/components/data/Scroll.js
```diff
@@ -0,0 +1,42 @@
...
+  variant = 'raised',
+
+  ...htmlAttributes
+}) {
+  return (
+    <div
+      {...htmlAttributes}
+      ref={downcastRef(elementRef)}
+      className={classnames(
+        'min-h-0 h-full w-full overflow-auto',
```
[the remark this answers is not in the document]
lyzadanger: This is to prevent a vertical container blowout if the `Scroll` is the immediate child of a `flex` container (which it likely is, if it's used with `ScrollContainer`), consistent with how we handle this situation for horizontal `flex` children. 

https://stackoverflow.com/questions/36230944/prevent-flex-items-from-overflowing-a-container , but in vertical because it's a columnar layout.

src/components/data/Scroll.js
lyzadanger: Fair question. I can't cite a current use case. But I could foresee potential use cases:

* A case in which we are unable to, for some reason, set a background color on the scrollable content, or it needs to be non-white. This would break the shadowing.
* A case in which the alignment of scrollable content was such that the shadows were mis-aligned.

However, arguably, that could be handled by not using this component in those cases. Hmmmm...

src/pattern-library/components/patterns/data/Scroll.js
lyzadanger: I touched on this in another comment, but the pattern I'm aiming for is composition of presentational components within a composite component, versus a single, and sometimes magical or inflexible, composite component. This gives the author the opportunity to compose things differently if there is a need for flexibility. Compare this to the current situation, in which occasionally it's like, "oh, crap, I need there to be one tick less padding on an element two levels deep in this composite component" or "the font size of a subheading absolutely must be `14px` in this context", which, believe me, comes up more often than you'd think (especially with our existing applied design in the client; it's like the English language, more inconsistencies than consistencies :) ). 

A theme I'm hearing is that we should be more clear about putting the components that are in the "standard/default" set more in the forefront, and backgrounding the lower-level presentational components that make up these default composite components a bit, as, in a worst case, they could sort of be implementation details of the composite components. This feels more like a documentation issue than anything.

src/tailwind.scroll-shadows.js
lyzadanger: Thanks! This CSS is frankly complicated and non-obvious, so comments were needed.

src/tailwind.scroll-shadows.js
lyzadanger: Ah, still broken in 15.2. I need to perform a system update to update my Safari to test in 15.4. I can do that later today, but don't want to take the time to do the update at this moment.

To wit, I see only bottom shadows (and they never go away) in Safari 15.2.

src/types.js
lyzadanger: Let's see how this pans out when we implement `Table`, which is coming up soon. That's probably on the more complex side of any _common_ component we'd have.

src/components/data/Scroll.js
lyzadanger: After some reflection, I think I'd like to leave this "shadow kill switch" in place for the time being, and leave myself a note to revisit it at the end of this current project. One reason is that I know that `Table` will make use of scroll components, which may involve a use case in which shadows are problematic. 

I'm sort of like an airline pilot, a gut dislike of having any automated system that can't be disabled in an emergency :P

I added a task on the project board to revisit this.

src/pattern-library/components/patterns/data/Scroll.js
```diff
@@ -0,0 +1,187 @@
+import { Scroll, ScrollContainer, ScrollContent } from '../../../../next';
+import Library from '../../Library';
+import Next from '../../LibraryNext';
+
+import { SampleListElements } from '../samples';
+
+export default function ScrollPage() {
```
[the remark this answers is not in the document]
lyzadanger: Ah, interesting. Would you prefer that pattern-library page modules be named `{{componentName}}Page.js`?

I'm open to that as a follow-up.

src/components/data/Scroll.js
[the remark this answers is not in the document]
lyzadanger: p.s. I can add a comment.

src/components/data/ScrollContainer.js
```diff
@@ -0,0 +1,42 @@
+import classnames from 'classnames';
+
+import { downcastRef } from '../../util/typing';
+
+/**
+ *
+ * @typedef {import('../../types').PresentationalProps} CommonProps
```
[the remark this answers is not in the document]
lyzadanger: This dumb gap should also be fixed in future as my scaffolding templates are fixed for it.

src/pattern-library/components/patterns/data/Scroll.js
lyzadanger: Changed to `Scroll`

src/tailwind.scroll-shadows.js
lyzadanger: Comment updated.

lyzadanger:
Thanks for taking the time and giving feedback, @robertknight !

I've taken action on:

* Small nits (idiot typos), typing adjustments
* Updating comments for Safari `background-attachment` and reasoning for `min-h-0` on an element
* Clarifying documentation to promote `ScrollBox` as the preferred approach

There are a number of follow-up items, which I've added as tasks to this project's [project board](https://github.com/orgs/hypothesis/projects/62/views/1). These items should be small tasks/PRs:

* Restructure tailwind plugins to not use default exports (and I might put tailwind plugins in a directory)
* Add `data-component` attribute to all components for debugging ergonomics
* Rename pattern-library page modules to match their component names

A larger, but important general follow-up is to revise documentation organization and navigation to make component hierarchies clearer (promote standard components while de-emphasizing the options for customization). This needs a bit of a think, and there are a few limitations with the current router implementation. This is also tracked.

src/components/data/Scroll.js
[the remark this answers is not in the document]
robertknight: > p.s. I can add a comment.

Please do. The explanation from https://stackoverflow.com/a/66689926/434243 is subtle:

> min-width in a flex context: While the default min-width value is 0 (zero), for flex items it is auto. This can make block elements take up much more space than desired, resulting in overflow.

src/components/data/Scroll.js
```diff
@@ -0,0 +1,44 @@
...
+
+  ...htmlAttributes
+}) {
+  return (
+    <div
+      {...htmlAttributes}
+      ref={downcastRef(elementRef)}
+      className={classnames(
+        // Prevent vertical flex or grid blowout: keep this contained vertically
+        // to parent by setting a min width
```
robertknight, line 34:
```suggestion
        // to parent by setting a min height
```

I would also be fine with delegating most of the details to the Stack Overflow answer you mentioned (eg. ``// Prevent overflow by overriding `min-height: auto`. See https://stackoverflow.com/a/66689926/434243.``)

robertknight (approved):
LGTM. Thanks for the detailed responses.

PR #508, opened 2022-07-22
Add documentation for Using Components

lyzadanger:
This PR adds some basic documentation for how to use the different categories of components provided by the package and their APIs.

Run `make dev` and visit http://localhost:4001/using-components

image

Depends on #505

### DIFF
--- a/src/pattern-library/components/patterns/UsingComponents.js
+++ b/src/pattern-library/components/patterns/UsingComponents.js
@@ -1,0 +1,89 @@
+import Library from '../Library';
+
+import Next from '../LibraryNext';
+
+export default function UsingComponentsPage() {
+  return (
+    <Library.Page
+      title="Using components"
+      intro={
+        <>
+          <p>This guide applies to updated ({'"next"'}) components only.</p>
+        </>
+      }
+    >
+      <Library.Pattern title="Component categories">
+        <ul>
+          <li>
+            <strong>Simple components</strong>: Opinionated, simple components
+            that apply design patterns. Example: <code>Spinner</code>.
+          </li>
+          <li>
+            <strong>Presentational components</strong>: Composable components
+            with customization flexibility. Example: <code>Button</code>,{' '}
+            <code>Link</code>.
+          </li>
+          <li>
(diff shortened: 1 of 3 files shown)


src/pattern-library/components/patterns/UsingComponents.js
```diff
@@ -0,0 +1,89 @@
...
+ * @prop {import('preact').ComponentChildren} [children]
+ * @prop {string|string[]} [classes] - Optional extra CSS classes to append to the
+ *   component's default classes
+ * @prop {never} [className] - Use variants, props, unstyled component (when
+ *   available) or classes instead
+ * @prop {import('preact').Ref<HTMLElement>} [elementRef] - Ref for component's
+ *   outermost element.
+ */
+`}
+        />
```
lyzadanger, line 75: There are about fifteen hundred more elegant ways to pull this off but...it works for the moment.

I had initially written out prose descriptions of the common props but found that nothing was as insightful as the typedefs themselves.

src/pattern-library/components/patterns/UsingComponents.js
```diff
@@ -0,0 +1,89 @@
+import Library from '../Library';
+
+import Next from '../LibraryNext';
+
+export default function UsingComponentsPage() {
+  return (
+    <Library.Page
+      title="Using components"
+      intro={
+        <>
```
robertknight, line 10: The fragment is currently redundant here.

src/pattern-library/components/patterns/UsingComponents.js
robertknight: One way we could address this is with convenient links from the component library web pages to the relevant part of the source code, either on GitHub (easiest solution) or rendered by the dev server itself (more work, especially if we wanted to serve this as a static site).

robertknight (approved):
LGTM with just a couple of minor comments.

lyzadanger:
@robertknight : I was a little trigger-happy with merge on this PR, apologies; I would like to combine your minor feedback with a documentation-related request in another PR as a small follow-up.

PR #514, opened 2022-07-26
Add re-implemented `Card` presentational components and `Panel` composite component

lyzadanger:
This PR adds updated `Card` and `Panel` components, using conventions for presentational and composite components.

A known thing is that we need to improve the navigation in the pattern library, as in the `Scroll` PR (#507). I will be working on that as a follow-up.

Check out this branch, run `make dev` and visit http://localhost:4001/layout-card and http://localhost:4001/layout-panel

image

image

Depends on #507 

### Component checklist

- [x] component tests
- [x] exports in package


src/components/layout/Card.js
```diff
@@ -0,0 +1,57 @@
+import classnames from 'classnames';
+
+import { downcastRef } from '../../util/typing';
+
+/**
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
```
robertknight, line 7: I'd suggest using `HTMLElement` here instead of `HTMLDivElement`. Same with other elements.

src/components/layout/CardHeader.js
```diff
@@ -0,0 +1,54 @@
...
+        classes
+      )}
+      ref={downcastRef(elementRef)}
+      data-component="CardHeader"
+    >
+      {title && <CardTitle>{title}</CardTitle>}
+      {children}
+      {onClose && (
+        <IconButton onClick={onClose} title="Close">
+          <CancelIcon />
```
robertknight, line 49: Is it intentional that the `icon` prop is not being used to render the icon here?

src/components/layout/CardTitle.js
```diff
@@ -0,0 +1,33 @@
...
+ * @typedef {import('preact').JSX.HTMLAttributes<HTMLDivElement>} HTMLDivAttributes
+ *
+ */
+
+/**
+ * Style a title for a Card
+ *
+ * @param {CommonProps & HTMLDivAttributes} props
+ */
+export default function CardTitle({
```
robertknight, line 16: I wonder if we should be using a semantic heading of some sort for card titles?

src/components/layout/test/Panel-test.js
```diff
@@ -0,0 +1,37 @@
...
+import { EditIcon } from '../../icons';
+
+const createComponent = (Component, props = {}) => {
+  return mount(<Component {...props}>This is child content</Component>);
+};
+
+describe('Panel', () => {
+  testCompositeComponent(Panel, 'Panel');
+
+  // This is an auto-generated test. Expand or delete.
```
robertknight, line 15: Does anything need to be done with this TODO?

src/pattern-library/routes.js
```diff
@@ -225,8 +221,18 @@ const routes = [
...
+    group: 'layout',
+    component: CardPage,
+    route: '/layout-card',
+  },
+  {
+    title: 'Panel',
+    group: 'layout',
+    component: PanelPage,
+    route: '/layout-panel',
+  },
```
robertknight, line 235: As with `ScrollBox` and `Scroll`, I think it would be good to put the high-level component first, or in some other way make the composite component, that is a recommended composition of the lower-level components, more discoverable.

robertknight (approved):
I left some feedback on a few things. The overall picture looks good though.

src/pattern-library/components/patterns/layout/Card.js
```diff
@@ -0,0 +1,349 @@
...
+            </li>
+            <li>
+              <code>CardActions</code> lay out actions (buttons) in a{' '}
+              <code>Card</code>
+            </li>
+          </ul>
+        </p>
+      }
+    >
+      <Library.Pattern title="Status">
```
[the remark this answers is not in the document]
lyzadanger: Already on a TODO list for the next week or so!

src/pattern-library/components/patterns/layout/Card.js
```diff
@@ -0,0 +1,349 @@
...
+            The <code>size</code> prop (<em>default</em> <code>md</code>)
+            adjusts relative padding and spacing in <code>CardContent</code>.
+          </p>
+
+          <Library.Demo title="size" withSource>
+            <Card>
+              <CardContent size="sm">
+                <div>
+                  <code>CardContent</code> component with{' '}
+                  <code>size=&quot;sm&quot;</code>.
```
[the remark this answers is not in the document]
lyzadanger: Until then, I should really write this as:

```
<code>size={"'sm'"}</code>
```

for readability.

src/pattern-library/components/patterns/layout/Card.js
```diff
@@ -0,0 +1,349 @@
...
+                </div>
+                <div>
```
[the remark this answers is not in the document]
lyzadanger: OK, I'll change this!

src/components/layout/test/Panel-test.js
lyzadanger: This is part of the auto-generated test file from the `plop` scaffolding I'm using...hmmm...in this case I did the "expanding" as indicated but neglected to remove the comment...hmmmm...I'll think about how not to overlook this.

src/components/layout/CardHeader.js
lyzadanger: Ah! My explanation is that I actually implemented the `Card` components before we sorted out that pattern for handling icons. I want to do this as a follow-up because I think...we may be able to get rid of the `CardTitle` component, potentially.

src/components/layout/CardTitle.js
lyzadanger: Good idea. I'll address this when I address adding `icon` to `CardHeader` in a follow-up.

src/pattern-library/routes.js
lyzadanger: I've swapped this navigation ordering for now. `Card` vs. `Panel` is different than `ScrollBox` in that using `Card` directly is a common pattern, whereas using the "lower-level" `Scroll` components directly would be more uncommon.

src/pattern-library/components/patterns/layout/Card.js
```diff
@@ -0,0 +1,349 @@
...
+import Next from '../../LibraryNext';
+
+export default function CardPage() {
+  return (
+    <Library.Page
+      title="Card"
+      intro={
+        <p>
+          The <code>Card</code> family of presentational components provide
+          support for laying out content in a card-like interface:
```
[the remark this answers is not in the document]
lyzadanger: Captured for follow-up!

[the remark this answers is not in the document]
lyzadanger: > I notice the leading "C" of "Card" in the <code> blocks gets clipped slightly in Safari. 

Are you sure? When I look at this in Safari it looks like the monospace `C` glyph has a very flat vertical (not much curve), but I'm not spotting actual clipping.

PR #515, opened 2022-07-26
Add JSX syntax highlighting

robertknight:
Add syntax highlighting to JSX code snippets. I looked at [Prism](https://prismjs.com) and [highlight.js](https://highlightjs.org). Highlight.js won as it has built-in types and better support for being imported as an ES module (ie. plays nicer with Rollup).

**Before:**

No syntax highlighting

**After:**

JSX syntax highlighting


lyzadanger (approved):
Very cool, ship it.

PR #516, opened 2022-07-27
Add data- attributes to components for better debugging

lyzadanger:
This PR adds `data-component` and `data-composite-component` attributes to all components' outer elements and adds tests to ensure we keep these up-to-date and consistent.

The `data-composite-component` attribute is necessary for composite components, whose outermost container is another component. Thus for `ScrollBox` the outer `div` is both a `ScrollBox` and a `ScrollContainer`:

`<div data-composite-component="ScrollBox" class="..." data-component="ScrollContainer">`

I added an optional `componentName` parameter to common tests to account for component-name collisions during this time of re-implementation.

### DIFF
--- a/src/components/data/Scroll.js
+++ b/src/components/data/Scroll.js
@@ -37,6 +37,7 @@ export default function Scroll({
         { 'scroll-shadows': variant === 'raised' },
         classes
       )}
+      data-component="Scroll"
     >
       {children}
     </div>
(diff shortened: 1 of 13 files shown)


robertknight (commented):
Somewhere we should indicate that the `data-component` and `data-composite-component` attributes exist for the benefit of developers looking at the DOM structure in devtools, and potentially other debugging use cases, and aren't intended to be used (at least, not currently) for other purposes. Currently this is mentioned in the tests and this PR description. Perhaps worth adding a note in the documentation too?

robertknight (approved):

PR #518, opened 2022-07-28
Improve rendering of JSX props whose values are JSX expressions

robertknight:
Call `jsxToString` recursively to render these prop values as readable JSX strings, rather than just stringifying them, which produces the string `[object Object]`. I also fixed rendering of fragments in this function. Previously they were rendered as `<p>...</p>` instead of `<>...</>` because the minified `Fragment` function happened to have a `name` property value of `p`.

**Before:**

JSX attr before

**After:**

JSX attr after v2

Note that in the "after" screenshot the text inside the JSX element has some incorrect highlighting. From various discussions in the Highlight.js repository I believe this is due to its relatively crude regex-based processing and the way JSX support is implemented, although I'd need to look into that deeper to get a better understanding.

### DIFF
--- a/src/pattern-library/util/jsx-to-string.js
+++ b/src/pattern-library/util/jsx-to-string.js
@@ -1,6 +1,7 @@
 import hljs from 'highlight.js/lib/core';
 import hljsXMLLang from 'highlight.js/lib/languages/xml';
 import hljsJavascriptLang from 'highlight.js/lib/languages/javascript';
+import { Fragment } from 'preact';
 
 /**
  * Escape `str` for use in a "-quoted string.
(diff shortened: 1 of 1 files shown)


robertknight:
We don't currently have any tests for code in the pattern library. Although type checking provides some safety net, I think it will probably make sense to add some in future, for utilities like this.

src/pattern-library/util/jsx-to-string.js
```diff
@@ -31,6 +32,17 @@ function indentLines(str, indent) {
   return lines.map(line => indentStr + line).join('\n');
 }
 
+/**
+ * Test if an element looks like a JSX element.
+ *
+ * @param {any} value
+ * @return {value is import('preact').VNode<any>}
```
lyzadanger, line 39: I always forget that such a thing is possible :)

lyzadanger (approved):
Thank you for doing this! It was on my longlist.

PR #521, opened 2022-07-29
Intercept in-app anchor links and trigger client-side navigation

robertknight:
Previously clicking on a link from within the content of a page to another page
would trigger a full page reload. Only the links in the navigation bar used
client-side navigation. Update the router to intercept all clicks on "in-app"
URLs and handle them client-side.

In the process add missing types in `router.js`, and remove the custom `onClick`
handlers for links in the navbar, which are no longer required.

We discussed that we might want to replace the custom router with a 3rd-party
solution at some point, but we will need to evaluate options.

**Testing:**

1. Click around the links in the navigation bar, including the Hypothesis logo. They should work as before.
2. Go to the Spinner page (http://localhost:4001/feedback-spinner) and click on the "SpinnerSpokesIcon" link near the top. It should do a client-side navigation and scroll to the appropriate location.

### DIFF
--- a/src/pattern-library/components/PlaygroundApp.js
+++ b/src/pattern-library/components/PlaygroundApp.js
@@ -32,14 +32,15 @@ export default function PlaygroundApp({
 }) {
   const routes = getRoutes();
   const allRoutes = routes.concat(extraRoutes);
-  const [activeRoute, navigate] = useRoute(baseURL, allRoutes);
-  const content = activeRoute ? (
-    <activeRoute.component />
-  ) : (
-    <Library.Page title=":( Sorry">
-      <h1 className="text-2xl">Page not found</h1>
-    </Library.Page>
-  );
+  const [activeRoute] = useRoute(baseURL, allRoutes);
+  const content =
+    activeRoute && activeRoute.component ? (
+      <activeRoute.component />
+    ) : (
+      <Library.Page title=":( Sorry">
+        <h1 className="text-2xl">Page not found</h1>
+      </Library.Page>
+    );
 
   /**
    * Header for a primary section of nav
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/PlaygroundApp.js
```diff
@@ -32,14 +32,15 @@ export default function PlaygroundApp({
...
-  const content = activeRoute ? (
-    <activeRoute.component />
-  ) : (
-    <Library.Page title=":( Sorry">
-      <h1 className="text-2xl">Page not found</h1>
-    </Library.Page>
-  );
+  const [activeRoute] = useRoute(baseURL, allRoutes);
+  const content =
+    activeRoute && activeRoute.component ? (
```
robertknight, line 37: Some entries in `allRoutes` don't have components, as they are navigation placeholders.

src/pattern-library/components/PlaygroundApp.js
```diff
@@ -70,7 +71,6 @@ export default function PlaygroundApp({
               'border-transparent': activeRoute?.route !== route.route,
             })}
             href={route.route.toString()}
-            onClick={e => navigate(e, route.route)}
```
robertknight, line 73: This is no longer needed, as clicks on all `<a>` elements are now intercepted and handled client-side, depending on the URL.

lyzadanger:
Note: for testing, apply the patch linked from the description of #520 (this will add the links referenced in the testing instructions).

src/pattern-library/components/PlaygroundApp.js
lyzadanger: The data structure for `route`s is an accretion over time. What they really are could probably more accurately be described as "pages", which may or may not have a route associated with them. Might make sense to re-assess how they're structured at some point but it's not a burning thing (IMHO) at this time.

lyzadanger (approved):
This is super! Thanks!

PR #539, opened 2022-08-01
Differentiate export names from function names in components

lyzadanger:
To alleviate collisions in the tests in projects using these shared
components — for now, at least — differentiate the reference naming of
default component exports to ensure unique default export names.

### DIFF
--- a/src/components/data/Scroll.js
+++ b/src/components/data/Scroll.js
@@ -16,7 +16,7 @@ import { downcastRef } from '../../util/typing';
  *
  * @param {CommonProps & ScrollProps & HTMLAttributes} props
  */
-export default function Scroll({
+const ScrollNext = function Scroll({
   children,
   classes,
   elementRef,
@@ -42,4 +42,6 @@ export default function Scroll({
       {children}
     </div>
   );
-}
+};
+
+export default ScrollNext;
--- a/src/components/test/common-tests.js
+++ b/src/components/test/common-tests.js
@@ -96,7 +96,7 @@ export function testPresentationalComponent(Component, componentName) {
       assert.isTrue(wrapperOuterEl.hasAttribute('data-component'));
       assert.equal(
         wrapperOuterEl.getAttribute('data-component'),
-        componentName ?? Component.name
+        componentName ?? Component.displayName ?? Component.name
       );
     });
   });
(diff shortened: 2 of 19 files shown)


src/components/data/Scroll.js
```diff
@@ -16,7 +16,7 @@ import { downcastRef } from '../../util/typing';
  *
  * @param {CommonProps & ScrollProps & HTMLAttributes} props
  */
-export default function Scroll({
+const ScrollNext = function Scroll({
```
lyzadanger, line 19: Differentiating between the export name (which must be unique at this time to prevent `rollup` collisions) and `function.name` (which is used by tests in this package, and sets the display name — handy for app-side tests).

robertknight:
To expand on the issue with collisions, for the benefit of someone else reading this:

1. We have a lot of `wrapper.find("ComponentName")`-type expressions in our tests, where `ComponentName` matches the component's `displayName` or `name` properties.
2. If multiple top-level symbols in a bundle have the same name, Rollup will modify the names to avoid a conflict (eg. `function Link` will be renamed to `function Link$1`) and update references. This has the side effect of altering the function's `.name` property, which in turn breaks the expressions from (1).
3. This PR fixes the problem by temporarily assigning the component functions to variables rather than exporting them directly. It is these variables that will be renamed by Rollup if needed, instead of the functions themselves, although non-conflicting names have been chosen for them. As a result, the old/new components of the same name retain their `name` property, and the tests mentioned in (1) don't break.

src/components/test/common-tests.js
```diff
@@ -96,7 +96,7 @@ export function testPresentationalComponent(Component, componentName) {
       assert.isTrue(wrapperOuterEl.hasAttribute('data-component'));
       assert.equal(
         wrapperOuterEl.getAttribute('data-component'),
-        componentName ?? Component.name
+        componentName ?? Component.displayName ?? Component.name
```
robertknight, line 99: Is this still needed, given that the component's `name` properties haven't changed?

robertknight (approved):
LGTM. Do you want to create a reminder issue to undo this once the migration is complete?

src/components/test/common-tests.js
lyzadanger: Well-spotted! Technically _no_, but I believe it's the right precedence overall.

lyzadanger:
> LGTM. Do you want to create a reminder issue to undo this once the migration is complete?

Was planning on it, yep!

PR #558, opened 2022-08-13
Updated DataTable and Table components

lyzadanger:
This is a draft PR proposing a structure for updated table-related components. The resulting set of functionality is reasonably consistent with the legacy `Table` component, but how we get there is restructured.

Highlights:

* Clearer separation of presentational/styling aspects (presentational components `Table`, `TableHead`, `TableBody`, `TableRow`, `TableCell`, hewing closely to their respective HTML elements) vs. data/behavior (new composite `DataTable` component)
* Introduction of lightweight contexts in a few places to give presentational components more composition flexibility
* Integration of `useArrowKeyNavigation` hook for interactive tables

Don't panic! The 1500+-line diff is misleading. Changes that require code review are about 500 lines (the rest of it: 650 lines of pattern-library page documentation and sample fixture data; 300 lines of existing hook and util code copied over from the `client` as-is).

Places where feedback would be helpful:

* The use of contexts in several places. I cribbed the idea from a couple of other libraries I reviewed (I spent a while looking at how the Other Guys structure table-related components, because they can be Very Complicated). We don't use contexts a whole lot usually, but they seemed helpful in these cases. There are a lot of bits and pieces that make up tables. Drilling props down gets fairly hairy. Thoughts?
* `Ref` typing hell/advice. Components in this collection take optional `elementRef` props. A particular use case kept popping up in this work: I need a ref for the component's outer element: if `elementRef` is provided, use that, otherwise generate a usable, properly-typed Ref. In these changes, you'll see a simple `useElementRef` hook that "does what I need", but I'm 95% certain it's not the right way to solve this problem. I'm all ears. By the end of the week I was pulling individual hairs out of my head in frustration at Ref object typing. I'm sure I can figure it out further if needed, but I was busy, and it's not fun and it makes me feel like a moron.


src/components/data/DataTable.js
```diff
@@ -0,0 +1,151 @@
...
+import TableBody from './TableBody';
+import TableRow from './TableRow';
+import TableCell from './TableCell';
+/**
+ * @typedef {import('../../types').CompositeProps} CommonProps
+ * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLElement>, 'size'|'rows'|'role'|'loading'>} HTMLAttributes
+ * @typedef {import('./TableContext').TableInfo} TableInfo
+ * @typedef {import('preact').ComponentChildren} Children
+ *
+ * @typedef Column
```
lyzadanger, line 17: In these newer Table components, terminology is tighter:

* `columns` vs. legacy `tableHeaders`
* `rows` vs. `items`
* An `item` in new components is an individual cell, not a row

src/components/data/DataTable.js
```diff
@@ -0,0 +1,151 @@
...
+
+  ...htmlAttributes
+}) {
+  const tableRef = useElementRef(elementRef);
+
+  useArrowKeyNavigation(tableRef, {
+    selector: 'tbody tr',
+    horizontal: true,
+    vertical: true,
+  });
```
lyzadanger, line 67: At this time, because we can't use hooks conditionally, _all_ `DataTable`s get keyboard navigation. We may want to be more selective. Any thoughts about that welcome.

src/components/data/Scroll.js
```diff
@@ -25,22 +29,30 @@ const ScrollNext = function Scroll({
...
-        'min-h-0',
-        'h-full w-full overflow-auto',
-        { 'scroll-shadows': variant === 'raised' },
-        classes
-      )}
-      data-component="Scroll"
-    >
-      {children}
-    </div>
+    <ScrollContext.Provider value={scrollContext}>
```
lyzadanger, line 39: See `TableRow` for "why" on this context

src/components/data/ScrollContext.js
```diff
@@ -0,0 +1,10 @@
+import { createContext } from 'preact';
+
+/**
+ * @typedef ScrollInfo
+ * @prop {import('preact').RefObject<HTMLElement>} scrollRef
+ */
+
+const ScrollContext = createContext(/** @type {ScrollInfo} */ ({}));
+
+export default ScrollContext;
```
lyzadanger, line 10: In these changes, contexts are default exports. This matches patterns I saw in other libraries, but I'm totally flexible.

src/components/data/TableCell.js
```diff
@@ -0,0 +1,53 @@
...
+const TableCellNext = function TableCell({
+  children,
+  classes,
+  elementRef,
+
+  ...htmlAttributes
+}) {
+  const sectionContext = useContext(TableSectionContext);
+  const isHeadCell = sectionContext && sectionContext.section === 'head';
+  const Cell = isHeadCell ? 'th' : 'td';
```
lyzadanger, line 31: Yay context!

src/components/data/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+  const tableContext = useContext(TableContext);
+
+  const isHeadRow = sectionContext?.section === 'head';
+
+  // Ensure that a selected row is visible when there is a scrolling context
+  useEffect(() => {
+    if (!selected || !scrollContext || !tableContext) {
+      return;
+    }
+    const scrollEl = scrollContext.scrollRef.current;
```
lyzadanger, line 46: In the legacy `Table` implementation, the `Table` component itself renders its own `ScrollBox` and then has to serve as a mediator between that `ScrollBox` and table rows in what feels like a slightly uncomfortable way.

Here, the row can make some decisions about itself based on whether it's in a scroll context, and is able to own its own logic/destiny.

src/components/data/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+    const tableHead = tableContext.tableRef.current?.querySelector('thead');
+
+    if (selected && scrollEl) {
+      // Ensure the row is visible within the scroll content area
```
lyzadanger, line 50: This comment should be expanded to explain when this can happen: a row is selected on initial render that is outside of the initial scroll viewport (i.e. below it).

src/components/data/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+      const scrollOffset = rowRef.current.offsetTop - scrollEl.scrollTop;
+      if (scrollOffset > scrollEl.clientHeight) {
+        rowRef.current.scrollIntoView();
+      }
+      // Ensure the row is not obscured by a sticky header
+      if (tableHead && tableContext.stickyHeader) {
```
lyzadanger, line 56: Once again, we don't need to drill down a prop here about whether there's a sticky header to account for, and we don't need to make any guesses about its dimensions, as we have access to a relevant ref.

src/components/data/DataTable.js
```diff
@@ -0,0 +1,151 @@
...
+const DataTableNext = function DataTable({
+  children,
+  elementRef,
+
+  columns,
+  rows,
+  title,
+  selectedRow,
+  loading = false,
+  renderItem,
```
lyzadanger, line 54: Ostensibly, we might add `renderRow` later if we find it useful. `renderItem` covers most use cases, but if there were a need to, say, have a cell span multiple columns in some situation, e.g., `renderRow` could be useful.

src/components/data/DataTable.js
```diff
@@ -0,0 +1,151 @@
...
+  onSelectRow,
```
lyzadanger, line 55: Similarly, we could have support for an `onSelectItem`/`onConfirmItem` later on—that's moving into true grid territory so would involve broader thinking.

src/hooks/keyboard-navigation.js
```diff
@@ -0,0 +1,198 @@
+import { useEffect } from 'preact/hooks';
```
lyzadanger, line 1: Directly copied from `client` with no changes.

src/hooks/use-element-ref.js
```diff
@@ -0,0 +1,14 @@
+import { useRef } from 'preact/hooks';
```
lyzadanger, line 1: As mentioned in the PR description, I doubt this hook is doing it right.

src/util/listener-collection.js
```diff
@@ -0,0 +1,81 @@
+/**
```
lyzadanger, line 1: Copied from `client` with no changes

lyzadanger (commented):
Notes.

robertknight:
> Ref typing hell/advice. Components in this collection take optional elementRef props. A particular use case kept popping up in this work: I need a ref for the component's outer element: if elementRef is provided, use that, otherwise generate a usable, properly-typed Ref. 

My initial understanding is that you are trying to support two uses of a ref:

1. A standard public API that allows consumers to get a reference to a component's root element. This is normally a `Ref<T>` which can be either a `RefObject<T>` or a `RefCallback<T>`.
2. Internal uses of a ref within a component, which require a ref object

One solution would be to have two separate refs which may be of different types (public ref may be an object or function, internal ref is always an object). A React/Preact element only accepts a single `ref` argument. We could perhaps create a utility similar to [this package](https://github.com/gregberge/react-merge-refs) (it is only ~15 LOC, so easy to adapt).

robertknight:
An alternative solution for refs would be to disallow function refs entirely, for certain components. We use them only rarely, but they do provide additional flexibility for consumers as they provide a way to get called when an element is mounted and removed. For that reason I would probably try to find a way to accommodate them.

src/components/data/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const isHeadRow = sectionContext?.section === 'head';
+
+  // Ensure that a selected row is visible when there is a scrolling context
+  useEffect(() => {
+    if (!selected || !scrollContext || !tableContext) {
+      return;
+    }
+    const scrollEl = scrollContext.scrollRef.current;
+    const tableHead = tableContext.tableRef.current?.querySelector('thead');
```
robertknight, line 47: If we can assume the parent will be a table, we could get the `<table>` ancestor of the `<tr>` and then gets its `<thead>`, avoiding the need for this particular ref. The context is still needed if you want to communicate stickiness down.

src/components/data/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+
+    if (selected && scrollEl) {
+      // Ensure the row is visible within the scroll content area
+      const scrollOffset = rowRef.current.offsetTop - scrollEl.scrollTop;
+      if (scrollOffset > scrollEl.clientHeight) {
+        rowRef.current.scrollIntoView();
```
robertknight, line 53: Does `scrollIntoView` do anything if the element is already in view. In other words, is the conditional essential in that scenario?

src/components/data/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+      // Ensure the row is not obscured by a sticky header
+      if (tableHead && tableContext.stickyHeader) {
+        const headingHeight = tableHead.clientHeight;
+        const headingOffset = scrollOffset - headingHeight;
+        if (headingOffset < 0) {
+          scrollEl.scrollBy(0, headingOffset);
+        }
+      }
+    }
+  }, [selected, rowRef, scrollContext, tableContext]);
```
robertknight, line 64: An alternative design here would be to for the row to receive a callback (somehow, could be via context or props) that it can use to scroll itself into view. That might allow some of the concerns around adjusting the scroll location to be localized more within either the table or the scrolling widget.

lyzadanger:
@robertknight This is finally ready for prime time! Whew. Changes since you last reviewed:

* Components that need it use your `useSyncedRef` hook
* Auto-scrolling logic moved from `TableRow` to `DataTable`, which feels like the right place
* Tests

Might be easier to go commit-by-commit so the big diff from the documentation doesn't throw you off. I could also split this into two PRs — `Table` family of presentational components vs. `DataTable` — if that would make reviewing this easier for you.

robertknight (commented):
I've had a look through this. I probably would have suggested splitting out `Table` and friends vs `DataTable` into separate PRs for review, but I've looked through the whole PR now. Aside from minor comments, the main feedback I had is:

- I think another pass will be required to fully implement all the accessibility affordances that ought to be provided (eg. communicating selected row, confirmation of rows, communication of loading and blank states). Perhaps file a general issue for this?
- Given the complexity of this component, I suggest adding some more notes in the intro about how interactivity works at a conceptual level (selection vs confirmation, is the component controlled or uncontrolled, what props trigger interactivity). Again, it might be best to use a separate issue to track this.

src/components/data/TableRow.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+const TableRowNext = function TableRow({
+  children,
+  classes,
+  elementRef,
+
+  selected,
+
+  ...htmlAttributes
+}) {
```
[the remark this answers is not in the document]
lyzadanger: Do you know off-hand why this would be the case? We're using `aria-selected` and `role="grid"` on the `table` element. I can do some digging but let me know if you have info.

src/components/data/TableRow.js
```diff
@@ -0,0 +1,63 @@
...
+        {
+          // Low-opacity backgrounds allow any scroll shadows to be visible
+          'odd:bg-slate-9/[.03]': !isHeadRow && !selected,
+          'bg-slate-7 text-color-text-inverted': selected,
+          'hover:bg-slate-9/[.08]': tableContext?.interactive && !selected,
+        },
+        classes
+      )}
+      data-component="TableRow"
+      data-interactive={tableContext?.interactive}
```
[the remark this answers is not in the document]
lyzadanger: It was likely a test affordance. I'll go poke around.

lyzadanger:
@robertknight  I think I will go ahead and break this into multiple PRs. That way we can get the less-complex `Table` presentational elements landed and spend a little more time on `DataTable` to get it right. Unless you demur?

src/components/data/TableRow.js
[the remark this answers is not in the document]
robertknight: I don't know why this would happen, so please dig away👷‍♀️.

src/components/data/Table.js
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <TableContext.Provider value={tableContext}>
+      <table
+        {...htmlAttributes}
+        aria-label={title}
+        className={classnames(
+          'w-full h-full layout-fixed',
+          // `border-separate` is required to handle borders on sticky headers
+          'border-separate border-spacing-0',
+          // No top border: Top border is established by `TableHead`
```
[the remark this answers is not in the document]
lyzadanger: Thanks for spotting this.

src/components/data/TableHead.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <TableSectionContext.Provider value={sectionContext}>
+      <thead
+        {...htmlAttributes}
+        ref={downcastRef(elementRef)}
+        className={classnames(
+          // Top border is set here instead of on `Table` to prevent sticky
+          // headers from causing a border wobble on scroll
+          'bg-grey-2 border-t',
```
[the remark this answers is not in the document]
lyzadanger: There was a problem with borders, but the lack of a border-color class is a red herring. There's a default border color set in the shared tailwind preset:

```
      borderColor: {
        DEFAULT: '#dbdbdb',
      },
```

(Default width and style is `1px` and `solid` out of the box).

This is both an ergonomic affordance — in most contexts[^1] you can just use `border` and get a 1-pixel solid border in the default color — and a consistency lock (borders are always the same color unless you expressly change the border color).

[^1]: "Most contexts" because this doesn't work consistently after certain resets we apply within shadow DOMs. I've called it out with comments when that is the case, e.g. in the client's `AdderToolbar`:

```
        // Because `.AdderToolbar` rules reset `all:initial`, we cannot use
        // default border values from Tailwind and have to be explicit about
        // border attributes
        'border border-solid border-grey-3',
```

src/components/data/Table.js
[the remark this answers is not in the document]
lyzadanger: 😅 Whew. Figured this out after some wrestling.

robertknight:
@lyzadanger - Can this be closed now that the changes have been extracted into separate PRs?

lyzadanger:
Oh, yes, overlooked. Closing in favor of #608

PR #559, opened 2022-08-13
Updated Checkbox component

### LINKED ISSUE #527: Re-implement Checkbox component(s)
Part of #455

lyzadanger:
This PR adds an updated `Checkbox` component and deprecates the associated legacy components.

* Updated `Checkbox` uses two separate icons for styling (unchecked, checked) instead of doing hackery with CSS classes. This nicely decouples the icons entirely. (Aside: Looks like I didn't add an example, yet, to the pattern-library documentation showing the use of different icons for styling. I can totally add that.)
* There's more clarity/division between using `Checkbox` in a controlled vs. uncontrolled "configuration"
* As with other components, converging on HTML/DOM APIs where possible (i.e. use `onChange` event handlers instead of custom `onToggle`)
* There's no need for a separate `LabeledCheckbox` component anymore

Docs available at http://localhost:4001/input-checkbox

image

Fixes #527


src/components/input/test/Checkbox-test.js
```diff
@@ -0,0 +1,47 @@
...
+    act(() => {
+      // See https://github.com/preactjs/enzyme-adapter-preact-pure/issues/123#issuecomment-699474177
+      const el = checkbox.getDOMNode();
+      el.setAttribute('checked', true);
+      el.dispatchEvent(new Event('change'));
+    });
+    wrapper.update();
+
+    assert.calledOnce(onChange);
+    assert.equal(checkbox.getDOMNode().getAttribute('checked'), 'true');
```
lyzadanger, line 45: HALLLLPPPPPP

robertknight: Maybe something like this:

```js
it('calls provided `onChange` callback when input changes', () => {
  const onChange = sinon.stub();
  const wrapper = createComponent(Checkbox, {
    onChange,
  });
                                                                    
  const checkbox = wrapper.find('input').getDOMNode();
  const event = new Event('change');
  act(() => {
    checkbox.dispatchEvent(event);
  });
                                                                    
  assert.calledWith(onChange, event);
});
                                                                    
it('updates controlled checkbox state on "change" event', () => {
  const wrapper = createComponent(Checkbox);
  const checkbox = wrapper.find('input').getDOMNode();
                                                                    
  // Toggle the checkbox without updating the controlled state
  const toggle = () => (checkbox.checked = !checkbox.checked);
                                                                    
  // Toggle the checkbox and update the controlled state
  const simulateChange = () => {
    toggle();
    act(() => {
      checkbox.dispatchEvent(new Event('change'));
    });
  };
                                                                    
  simulateChange();
  toggle();
  wrapper.setProps({});
  assert.isTrue(checkbox.checked);
                                                                    
  simulateChange();
  toggle();
  wrapper.setProps({});
  assert.isFalse(checkbox.checked);
});
```

I split the two scenarios into two, and to test controlled state updates, relied on the fact there is a way to update the checkbox's actual state without updating `Checkbox`'s controlled state.

robertknight: Edit: I used "controlled" in the previous comment originally when I meant "uncontrolled".

lyzadanger: Thank you!

src/components/input/test/Checkbox-test.js
```diff
@@ -0,0 +1,67 @@
...
+  return mount(<Component {...props}>This is child content</Component>);
+};
+
+describe('Checkbox', () => {
+  testCompositeComponent(Checkbox, { elementSelector: 'input' });
+
+  it('shows an icon representing checkbox state', () => {
+    const wrapper = createComponent(Checkbox);
+
+    assert.isTrue(wrapper.find('CheckboxOutlineIcon').exists());
```
robertknight, line 18: You can also shorten this to `wrapper.exists(<selector>)`.

src/pattern-library/components/patterns/input/CheckboxPage.js
```diff
@@ -0,0 +1,194 @@
...
+      </Library.Pattern>
+
+      <Library.Pattern title="Props">
+        <Library.Example title="checked">
+          <p>
+            When the <code>checked</code> prop is present, <code>Checkbox</code>{' '}
+            will behave as a controlled component.
+          </p>
+          <Next.Code
+            content={`const [controlledChecked, setControlledChecked] = useState(false);
```
robertknight, line 93: I'd suggest removing the word "controlled" from the identifiers here, to make the code a bit less verbose.

src/pattern-library/components/patterns/input/CheckboxPage.js
```diff
@@ -0,0 +1,194 @@
...
+          </p>
+          <Library.Demo withSource>
+            <Checkbox icon={HideIcon} checkedIcon={ShowIcon} defaultChecked />
+          </Library.Demo>
+        </Library.Example>
+
+        <Library.Example title="onChange">
+          <p>
+            Any provided <code>onChange</code> function is invoked with a{' '}
+            <code>TargetedEvent</code> when there is a <code>change</code> event
```
robertknight, line 168: There is no `TargetedEvent` interface/class in the web platform. All `Event` objects have a target, whether they are delivered to a DOM element or a global object (eg. document, window). The `change` event just receives a plain `Event` instance. I think the docs could just reference MDN here (eg. "onChange adds a handler for the ["change" event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) on the `<input>`").

In general, I think we can keep the docs succinct by assuming knowledge of how things work in HTML and deferring to MDN where a prop or other feature is a wrapper for an HTML attribute, event etc.

robertknight (approved):
LGTM, with a correction about a `TargetedEvent` reference.

PR #565, opened 2022-08-16
Add new `AspectRatio`, re-implemented `Thumbnail` component

### LINKED ISSUE #531: Re-implement thumbnail component
Part of #455

lyzadanger:
This PR re-implements `Thumbnail` by way of adding a new `AspectRatio` component. 

* `AspectRatio` is a new simple component that uses the tried-and-true "bottom-padding hack" to constrain its first direct child to a particular aspect ratio.
* `Thumbnail` is a composite component based on the legacy component of the same name. It builds atop `AspectRatio` and provides styling as well as support for loading states and placeholder content.

Details can be explored in http://localhost:4001/data-aspectratio and http://localhost:4001/data-thumbnail

image

image

Fixes #531

Depends on #575


src/components/data/AspectRatio.js
```diff
@@ -0,0 +1,81 @@
+import classnames from 'classnames';
+import { cloneElement, toChildArray } from 'preact';
+
+/**
+ * @typedef {import('preact').ComponentChildren} Children
+ *
+ * @typedef AspectRatioProps
+ * @prop {Children} children
+ * @prop {'cover'|'contain'|'fill'|'none'} [objectFit="cover"]
```
robertknight, line 9: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit mentions a "scale-down" value for object-fit as well. Is that worth including?

src/components/data/AspectRatio.js
```diff
@@ -0,0 +1,81 @@
...
+  // - a "replaceable element", e.g. image or video (media), or
+  // - a block element, e.g. a div that contains placeholder content. In this
+  //   case, content within this node will be centered horizontally and
+  //   vertically.
+  const firstChildNode =
+    /** @type {import('preact').VNode<any>|undefined} */
+    (toChildArray(children).find(child => typeof child === 'object'));
+
+  const otherChildren = firstChildNode
+    ? toChildArray(children).filter(child => child !== firstChildNode)
```
robertknight, line 46: The `toChildArray` call could be hoisted to avoid computing it twice.

src/components/data/Thumbnail.js
```diff
@@ -0,0 +1,83 @@
...
+        'p-3': size === 'md' && !borderless, // default
+        'p-2': size === 'sm' && !borderless,
+        'p-4': size === 'lg' && !borderless,
+      })}
+      data-composite-component="Thumbnail"
+    >
+      <div className="bg-white h-full w-full flex items-center justify-center overflow-hidden">
+        <AspectRatio ratio={ratio}>
+          {loading ? (
+            <div>
```
robertknight, line 71: What is the purpose of the unstyled `<div>`? Is it to provide an element for `AspectRatio` to scale, without affecting the visual size of the spinner?

src/components/data/test/Thumbnail-test.js
```diff
@@ -0,0 +1,53 @@
...
+    // The "border" on Thumbnails is formed by padding classes. Default
+    // padding size ("md" size) is `p-3`. The absence of padding means there
+    // is no border.
+    assert.isTrue(
+      withBorder.find('[data-composite-component="Thumbnail"]').hasClass('p-3')
+    );
+    assert.isFalse(
+      withoutBorder
+        .find('[data-composite-component="Thumbnail"]')
+        .hasClass('p-3')
```
robertknight, line 27: This test would ideally check for element not having any padding, rather than not having a specific padding. If `borderless` caused the padding to be set to some other value for example, this would pass.

src/pattern-library/routes.js
```diff
@@ -6,6 +6,8 @@ import CustomizingComponentsPage from './components/patterns/CustomizingComponen
 import UsingComponentsPage from './components/patterns/UsingComponentsPage';
 import UtilitiesPage from './components/patterns/UtilitiesPage';
 
+import AspectRatioPage from './components/patterns/data/AspectRatioPage';
+import ThumbnailPage from './components/patterns/data/ThumbnailPage';
```
robertknight, line 10: Can you sort these imports.

src/pattern-library/components/patterns/data/AspectRatioPage.js
```diff
@@ -0,0 +1,171 @@
...
+                underline="always"
+              >
+                Object positioning
+              </Link>{' '}
+              within the content {"element's"} box can be handled by styling the
+              content element itself. By default, the browser will position
+              content <code>{'50% 50%'}</code> within the {"element's"} box. In
+              this example, the image is styled to apply{' '}
+              <code>{'object-position: top'}</code> instead.
+            </p>
```
robertknight, line 87: For this kind of example the behavior of the prop is best illustrated through comparison. A future improvement could add radio buttons to switch between different prop values.

robertknight (approved):
I left a few minor comments, but overall very nice 👍

src/components/data/Thumbnail.js
lyzadanger: `AspectRatio` extends styling on its first immediate child element to set it to have (as you queried about here) the right aspect-ratio, but also sets its as flex with items aligned and justified center (i.e. centered vertically and horizontally). Thus, this `div` is needed to position the loading icon correctly. This is explained, perhaps poorly, in AspectRatio pattern-library documentation when it talks about adding placeholder content. Any ideas for how to make this more clear?

src/components/data/AspectRatio.js
lyzadanger: Yeah, maybe I should be completist here; I was aware of `scale-down` but didn't feel that it would be a typical usage. But who am I to judge? I don't love the feeling of re-implementing `object-fit` via a prop here...I'll reflect. But yeah, I can add `scale-down`.

PR #567, opened 2022-08-17
Add keyboard-navigation hook and listener collection dependency

lyzadanger:
Add `use-keyboard-navigation` from the client, and its dependency, `listener-collection`. Also copy over tests and one test utility.

These were directly copied with no changes.

`useArrowKeyNavigation` will be used for `DataTable` components.

### DIFF
--- a/src/hooks/keyboard-navigation.js
+++ b/src/hooks/keyboard-navigation.js
@@ -1,0 +1,198 @@
+import { useEffect } from 'preact/hooks';
+
+import { ListenerCollection } from '../util/listener-collection';
+
+/**
+ * @param {HTMLElement & { disabled?: boolean }} element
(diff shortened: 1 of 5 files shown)


robertknight (approved):
LGTM. If we expose these hooks as part of the API of the package then we can make this package the canonical implementation of them, and remove the copies in the client, to avoid code duplication. In future there might be value in separating out some of these hooks/utilities into a separate package from the components.

lyzadanger:
> LGTM. If we expose these hooks as part of the API of the package then we can make this package the canonical implementation of them, and remove the copies in the client, to avoid code duplication. In future there might be value in separating out some of these hooks/utilities into a separate package from the components.

Agree on all counts.

PR #568, opened 2022-08-17
Add `useSyncedRef` hook

robertknight:
Add a hook that can be used to synchronize an external ref provided to a
component with an internal object ref used by the component itself.


lyzadanger (approved):
I'm happy to give this a whirl as-is. Thanks for your work here! Phew, dense stuff.

PR #569, opened 2022-08-17
Version 5.3 PR 1/n: Documentation updates from feedback

### LINKED ISSUE #532: Update documentation based on expansion and feedback
## Using/Working with Components

* [x] Fix redundant Fragment
* [x] Reduce strikethrough usage
* [x] Add documentation about commonly-used specific props like `variant`, `size`
* [x] Add documentation about `data-` attrs

## Customizing components

* [x] Emphasize "stop and think" section
* [x] Better explain intent, and that conflicts are possible if using `classes`

## Utilities

* [x] Add `input-group` variant (if relevant)
* [x] Add documentation on ARIA- variants
* [x] Add introduction

lyzadanger:
This is the first of a series of PRs representing changes merged to the `version-5.3` branch.

* The [first commit](49cf9c589de1e95cc63c8baa1381c1f9deeb0283) satisfies https://github.com/hypothesis/frontend-shared/issues/532, which is a laundry list of feedback items related to documentation in the pattern library. These are _content_ updates.
* The [second commit](2b5ab3f5e1828bcdfd83cafe528ef46b4c1962f1) does some markup and structural cleanup of pattern-library pages to ensure consistency of casing and sections. I don't anticipate this needs granular review.

Fixes #532

### DIFF
--- a/src/pattern-library/components/patterns/CustomizingComponentsPage.js
+++ b/src/pattern-library/components/patterns/CustomizingComponentsPage.js
@@ -14,51 +14,74 @@ export default function CustomizingComponentsPage() {
         </>
       }
     >
-      <Library.Pattern title="How to customize a component">
+      <Library.Pattern title="Step 1: Stop and reflect">
         <p>
-          These steps should be taken in order when design or layout
-          customization of a component is necessary.
+          Components in this library are designed to be flexible within the
+          bounds of established design patterns. Certain guard rails exist on
+          purpose. Reflect before you customize!
         </p>
-        <ol>
-          <li>
-            Use a <strong>different variant</strong> of the component. Check the{' '}
-            {"component's"} documentation in this library to see what variants
-            are available.
-          </li>
-          <li>
-            Use <strong>component styling props</strong>, e.g.{' '}
-            <code>color</code>, <code>underline</code> or <code>size</code>.
-            These differ per component and define styling aspects that are
-            intended to be adjusted.
-          </li>
-          <li>
-            Use an <strong>unstyled component</strong>, e.g.{' '}
-            <code>LinkUnstyled</code>. This is appropriate if you want to do
-            considerable customization of a component.
-          </li>
-          <li>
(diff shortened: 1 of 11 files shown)


src/pattern-library/components/patterns/CustomizingComponentsPage.js
```diff
@@ -14,51 +14,74 @@ export default function CustomizingComponentsPage() {
...
+                </strong>
+                . This allows authors to append arbitrary classes to the{' '}
+                {"component's"} outermost element but should be used with
+                consideration.
+              </p>
+              <p>
+                The intention is to allow for extension, not override. There is
+                no guarantee that provided <code>classes</code> {"won't"}{' '}
+                conflict with classes applied by the component.
+              </p>
```
robertknight, line 71: I would suggest to say something like, "in the event of a conflict between the component's own styling and the classes specified here, there is currently no guarantee of which will take precedence." This issue creates a hazard when updating the library, so it would be good for us to find a general solution to this in future. [tailwind-merge](https://github.com/dcastil/tailwind-merge) is an example of the kind of thing I was thinking, though the implementation of that is unfortunately quite complex.

src/pattern-library/components/patterns/UsingComponentsPage.js
```diff
@@ -72,16 +76,145 @@ export default function UsingComponentsPage() {
...
+          <Library.Demo
+            title="Card variants, affecting shadows and hover effects"
+            withSource
+          >
+            <Card variant="raised">
+              <CardContent>
+                This is content inside of a <code>Card</code> with the default,{' '}
+                <code>raised</code> <code>variant</code>.
+              </CardContent>
+            </Card>
```
robertknight, line 124: The JSX formatting in the pattern library currently always puts new elements (eg. `<code>` here) on a separate line. It would be good to tweak that in future to be more like Prettier's output.

robertknight (approved):

PR #571, opened 2022-08-17
Version 5.3 PR 3/n: Add `LinkButton` component

### LINKED ISSUE #526: Re-implement LinkButton
Part of #455

- [x] component tests
- [x] exports in package

lyzadanger:
This PR adds an updated `LinkButton` component. This component is modeled against established patterns; one thing to note is that I've grouped it with "navigation", as its theming is synchronized with `Link` and I want to emphasize that it should be used in navigation-like contexts.

Pattern-library docs: http://localhost:4001/input-linkbutton

Fixes #526

Depends on #570


src/pattern-library/components/patterns/navigation/LinkButtonPage.js
```diff
@@ -0,0 +1,197 @@
...
+          </Library.Demo>
+          <Library.Demo title="color: 'text'" withSource>
+            <p className="text-color-text">
+              <LinkButton color="text">Page notes</LinkButton>{' '}
+            </p>
+          </Library.Demo>
+        </Library.Example>
+
+        <Library.Example title="underline">
+          <Library.Demo title="underline:'none' (default)" withSource>
```
robertknight, line 85: I suggest that, as with `Link`, we emphasize the importance of underlines when the link has surrounding text or appears in some other context where its clickability is non-obvious.

robertknight (approved):
LGTM, but see note about emphasis of applying `underline` prop when appropriate.

PR #572, opened 2022-08-17
Version 5.3 PR 4/n: Add `Spinner`, `Overlay` and `SpinnerOverlay` components

### LINKED ISSUE #528: Re-implement replacement for `FullScreenSpinner`
Part of #455

lyzadanger:
This PR replaces `Spinner` and `FullScreenSpinner` with `Spinner` and `SpinnerOverlay`, as well as a supporting `Overlay` component. Previously, "overlay" was a design pattern expressed in CSS classes. Now it is a component. It will also get used for modals later.

Pattern-library documentation: http://localhost:4001/feedback-spinner

Depends on #571 

Fixes #528

image


src/pattern-library/components/patterns/feedback/SpinnerPage.js
```diff
@@ -80,6 +102,25 @@ export default function SpinnerPage() {
...
+      <Library.Pattern title="SpinnerOverlay">
+        <p>
+          This component encapsulates a design pattern for a full-page loading
+          spinner.
+        </p>
+        <Library.Example title="Usage">
+          <Next.Usage componentName="SpinnerOverlay" />
+          <Library.Demo withSource>
+            <Button onClick={toggleOverlayOpen}>Show overlay</Button>
+            {overlayOpen && <SpinnerOverlay onClick={toggleOverlayOpen} />}
```
robertknight, line 115: An unfortunate effect of the conditional prop is that the `<SpinnerOverlay/>` component doesn't appear in the output when you view the "Source" page. Some possible ways to resolve this:

- Change the way the spinner overlay is shown so that it doesn't use a conditional. The downside is that the demo would not match exactly how the component would normally be used.
- Show the Demo and Source views at the same time, so the user can see the change in output when they click the button. This would still make it impossible to copy the source though
- Other ideas?

Longer term we could potentially change how the source view work so that conditionals, comments etc. are preserved.

src/components/feedback/SpinnerOverlay.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLElement>, 'className'>} HTMLAttributes
+ */
+
+/**
+ * Render a full-screen spinner atop a light-colored overlay
+ *
+ * @param {HTMLAttributes} props
+ */
+const SpinnerOverlayNext = function SpinnerOverlay({ ...HTMLAttributes }) {
```
robertknight, line 13:
```suggestion
const SpinnerOverlayNext = function SpinnerOverlay({ ...htmlAttributes }) {
```

Per our naming conventions, acronyms are lower cased if the first word of an identifier.

src/components/layout/test/Overlay-test.js
```diff
@@ -0,0 +1,19 @@
...
+import Overlay from '../Overlay';
+
+const createComponent = (Component, props = {}) => {
+  return mount(<Component {...props}>This is child content</Component>);
+};
+
+describe('Overlay', () => {
+  testPresentationalComponent(Overlay);
+
+  it("Doesn't render if `open` is `false`", () => {
```
robertknight, line 14:
```suggestion
  it("doesn't render if `open` is `false`", () => {
```

src/pattern-library/components/patterns/layout/OverlayPage.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  function toggleLightOpen() {
+    setLightOpen(!lightOpen);
+  }
+
+  function toggleDarkOpen() {
+    setDarkOpen(!darkOpen);
+  }
+  return (
```
robertknight, line 24:
```suggestion
  }
  
  return (
```

src/pattern-library/components/patterns/layout/OverlayPage.js
```diff
@@ -0,0 +1,82 @@
...
+import { Overlay } from '../../../../next';
+import Library from '../../Library';
+import Next from '../../LibraryNext';
+
+export default function OverlayPage() {
+  const [defaultOpen, setDefaultOpen] = useState(false);
+  const [darkOpen, setDarkOpen] = useState(false);
+  const [lightOpen, setLightOpen] = useState(false);
+
+  function toggleDefaultOverlay() {
```
robertknight, line 13: I think it is more conventional for us to use arrow functions `const toggleDefaultOverlay = ...` rather than named functions for helpers inside components.

robertknight (commented):
I found an issue with the `SpinnerOverlay` example where `<SpinnerOverlay>` doesn't appear in the source view due to use of a conditional. Otherwise than this and some minor comments, this looks good to me.

src/pattern-library/components/patterns/feedback/SpinnerPage.js
lyzadanger: I can make `SpinnerOverlay`'s API align with `Overlay`'s by adding an `open` prop...

lyzadanger: Here's what I'm going to do for now:

image

There are two forthcoming improvements:

1. After some changes down the line a bit are merged, the source will move _under_ the demo example. At this point in the stack of diffs, there's some logic in `Library` components that enforces that all `Demo`s are rendered "at the end" of examples. That will not be the case for long, allowing interleaving of Demos and other content.
2. I am considering extending the `Demo` component to take optional `source` content that would override its auto-generated source.

src/components/feedback/SpinnerOverlay.js
lyzadanger: Good catch, thanks!

robertknight (approved):

PR #573, opened 2022-08-17
Version 5.3 PR 5/5: Add library `Section` component and reorganize routes

lyzadanger:
This housekeeping and organization PR:

* Introduces a `Section` component to pattern-library pages. Basically, we were one hierarchical level short to be able to format pages that document multiple components. `Section` provides a new top-level division.
* Update pages to use `Sections`. This makes for big diffs, but there are minimal content changes.
* Combine a couple of pages into a single page, e.g. `ScrollBox` and `Scroll` are now one page. This is more feasible with the additional sectioning component.
* Do a little cleanup of library component types and add optional `id` props for upcoming (I hope) fragment navigation (I doubt authors will want to set `id`s manually in most cases, but the manual override is there if it's needed).
* Re-order routes to be strictly alphabetical

There is nothing significant here that needs detailed review.


robertknight:
There are quite a lot of different font sizes in one screen here:

Font sizes

I wonder if it may look better if some of these were consolidated.

lyzadanger:
> There are quite a lot of different font sizes in one screen here:
> 
> Font sizes
> 
> I wonder if it may look better if some of these were consolidated.

Good feedback. I'll take a look and see if I can make this better.

src/pattern-library/components/patterns/data/IconsPage.js
```diff
@@ -11,100 +11,102 @@ export default function IconsPage() {
...
-        </Library.Example>
-      </Library.Pattern>
+        <Library.Pattern title="Sizing icons">
+          <p>
+            Icons are sized at <code>16×16</code> unless styled otherwise.
+          </p>
+          <p>
+            Icons are often sized relative to the surrounding text. The{' '}
+            <code>em</code> spacing value provided by this {"package's"}{' '}
+            tailwind preset allows sizing at <code>1em</code>.
```
robertknight, line 80:
```suggestion
            Tailwind preset allows sizing at <code>1em</code>.
```

src/pattern-library/components/Library.js
```diff
@@ -81,38 +85,69 @@ function PageIntro({ children }) {
...
- * @prop {import("preact").ComponentChildren} [children]
- * @prop {string} [title]
- * @prop {'split'|'wide'} [variant='split'] - Layout variant. Applies
- *   appropriate className.
- *   - Split (default) lays out in a row. Non-demo example content is rendered
- *     left, with demos right. Demos in this variant stack vertically.
- *   - Wide lays out in a full-width column. Non-example is rendered first,
- *     then a row to contain demos. Demos in this variant render next to each
- *     other in a single row.
+ * Page introductory text
```
robertknight, line 109: Should this say "_Section_ introductory text"?

src/pattern-library/components/Library.js
```diff
@@ -81,38 +85,69 @@ function PageIntro({ children }) {
...
+      <h3 className="text-2xl text-slate-7" id={id}>
+        {title}
+      </h3>
+      <div className="space-y-8 px-4">{children}</div>
+    </section>
+  );
+}
+
+/**
+ * Render information about a tertiary section on a page.
```
robertknight, line 143: This component seems more specific than just "a tertiary section on a page".

src/pattern-library/components/Library.js
```diff
@@ -168,25 +197,21 @@ function DemoButton({ children, onClick, pressed }) {
...
  * rendered Demo content.
  *
- * @param {DemoProps} props
+ * @param {object} props
+ *   @param {import("preact").ComponentChildren} [props.children]
+ *   @param {string} [props.classes] - Extra CSS classes for the demo content's
+ *     immediate parent container
+ *   @param {boolean} [props.withSource=false] - Should the demo also render the source?
+ *   When true, a "Source" tab will be rendered, which will display the JSX
+ *   source of the Demo's children
```
robertknight, line 212:
```suggestion
 *     When true, a "Source" tab will be rendered, which will display the JSX
 *     source of the Demo's children
```

robertknight (approved):
This is a big PR so I mainly looked at Library.js, routes.js and then reviewed the other page changes less closely with a skim through the code and the rendered output. I found a few minor issues/queries in Library.js and routes.js. Otherwise looks good.

src/pattern-library/routes.js
```diff
@@ -100,6 +97,71 @@ const routes = [
...
+
+  { title: 'Icons', group: 'data', component: IconsPage, route: '/data-icons' },
+  {
+    title: 'ScrollBox',
+    group: 'data',
+    component: ScrollBoxPage,
+    route: '/data-scrollbox',
+  },
+  { title: 'Table', group: 'data' },
+  { title: 'Dialog', group: 'feedback' },
```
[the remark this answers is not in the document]
lyzadanger: Putting dialog and modal components in the "Feedback" group is something I've seen in some other libraries — Material UI, Ant, e.g. — and it made intuitive sense to me at the time, probably more so for modal than dialog. Other libraries have a separate "Overlays" group (Chakra, Mantine). What do you think? This is still a placeholder so we can hash it out more in-depth when `Dialog` and `Modal` components are added? I'm open...

src/pattern-library/routes.js
```diff
@@ -100,6 +97,71 @@ const routes = [
...
+    title: 'Link',
+    group: 'navigation',
+    route: '/navigation-link',
+    component: LinkPage,
+  },
+  {
+    title: 'LinkButton',
+    group: 'navigation',
+    component: LinkButtonPage,
+    route: '/input-linkbutton',
```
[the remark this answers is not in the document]
lyzadanger: Yes, it should. Thanks. Missed that when I moved it between groups 👍🏻

src/pattern-library/components/Library.js
lyzadanger: (You're right. I expanded this)

PR #575, opened 2022-08-18
Make small enhancements to Code, Example components and format object values nicely

lyzadanger:
This PR pops off three small commits from other component branches so that those branches may contain purely component- and component-documentation-related changes.

* Add scrolling to `Code` — long code content will now scroll if the parent container has dimension constraints (used for showing the `rows` used in `Table` examples)
* Remove unneeded complexity from the rendering of `Example` children. `Demo`s may now be interleaved with other content in `Examples` (Used to interleave content and demos in `AspectRatio` and `Thumbnail` examples).
* Make prop values that are objects render a little more nicely than their stringified `[object Object]`. This is hamfisted and quick but it helps.  (Make props passed to `DataTable` more readable in demo Source tabs)

### DIFF
--- a/src/pattern-library/components/LibraryNext.js
+++ b/src/pattern-library/components/LibraryNext.js
@@ -8,6 +8,8 @@ import { useMemo } from 'preact/hooks';
 
 import { jsxToHTML } from '../util/jsx-to-string';
 
+import { Scroll, ScrollContainer } from '../../next';
+
 /**
  * Render a little label or pill next to changelog items
  *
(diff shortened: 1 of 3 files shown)


src/pattern-library/components/LibraryNext.js
```diff
@@ -72,25 +77,29 @@ function Code({ content, size, title }) {
   const codeMarkup = useMemo(() => jsxToHTML(content), [content]);
 
   return (
-    <figure className="space-y-2">
-      <div
-        className={classnames(
-          'unstyled-text bg-slate-7 text-color-text-inverted p-4 rounded-md',
-          { 'text-sm': size === 'sm' }
+    <figure className="space-y-2 min-h-0 h-full">
```
robertknight, line 80: Is the `min-h-0` here for overriding `min-height: auto` if this element happens to be positioned in a flex container?

lyzadanger: Yep yep yep

robertknight (approved):

PR #577, opened 2022-08-19
Update test CSS bundling and add a11y tests to common tests

lyzadanger:
When writing the somewhat-more-involved tests for `DataTable` this afternoon, I realized there were two test things that needed doing:

1. The test CSS bundle didn't include CSS for updated components (i.e. Tailwind). This became a problem when I was trying to test `Table` behavior within constrained `Scroll` components.
2. I had neglected to add automated a11y tests to the common tests

This PR rectifies both of these.

The addition of accessibility tests kicked up one issue in the `Input` component that I have remedied for now and updated the associated pattern-library documentation. There's a broader thing to look at here as we disseminate this new component: making sure we are correctly labeling our text inputs.

### DIFF
--- a/src/components/test/common-tests.js
+++ b/src/components/test/common-tests.js
@@ -9,6 +9,13 @@ const createComponent = (Component, props = {}) => {
   return mount(<Component {...props}>This is child content</Component>);
 };
 
+function primaryElement(wrapper, elementSelector) {
+  if (elementSelector) {
+    return wrapper.find(elementSelector).getDOMNode();
+  }
+  return wrapper.childAt(0).getDOMNode();
+}
+
 /**
  * Set of tests common to all presentational components
  *
@@ -16,6 +23,9 @@ const createComponent = (Component, props = {}) => {
  * @prop {string} [componentName] - Temporary affordance to prevent name
  *   collisions between updated components and legacy components with the same
  *   component function name.
+ * @prop {(component:FunctionComponent, props:object) => ReactWrapper} [createContent] -
+ *   A function that returns the rendered output to test or an Enzyme wrapper
+ *   created using Enzyme's `mount` function.
  * @prop {string} [elementSelector] - Selector to find the element to which any
  *   passed HTML attributes are applied, when it is not the outermost element
  *   rendered by the component.
@@ -25,14 +35,14 @@ const createComponent = (Component, props = {}) => {
  */
 export function testCompositeComponent(
   Component,
-  { componentName, elementSelector } = {}
+  { componentName, createContent = createComponent, elementSelector } = {}
 ) {
   const displayName = componentName ?? Component.displayName ?? Component.name;
 
   describe(`Common composite functionality for ${displayName}`, () => {
     it('applies `ref` via `elementRef`', () => {
       const elementRef = createRef();
-      createComponent(Component, { elementRef });
+      createContent(Component, { elementRef });
 
       assert.instanceOf(elementRef.current, Node);
     });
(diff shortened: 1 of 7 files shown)


src/components/test/common-tests.js
```diff
@@ -9,13 +11,23 @@ const createComponent = (Component, props = {}) => {
...
 /**
  * Set of tests common to all presentational components
  *
  * @typedef CommonTestOpts
  * @prop {string} [componentName] - Temporary affordance to prevent name
  *   collisions between updated components and legacy components with the same
  *   component function name.
+ * @prop {(component:FunctionComponent, props:object) => import('preact').VNode|ReactWrapper} [content] -
+ *   A function that returns the rendered output to test or an Enzyme wrapper
+ *   created using Enzyme's `mount` function.
```
robertknight, line 30: Various tests below assume that the result of `content` is always a wrapper, rather than a VNode. Also `content` sounds like an object rather than a function. Perhaps `createContent`?

src/components/test/common-tests.js
```diff
@@ -25,22 +37,23 @@ const createComponent = (Component, props = {}) => {
  */
 export function testCompositeComponent(
   Component,
-  { componentName, elementSelector } = {}
+  { componentName, content, elementSelector } = {}
 ) {
   const displayName = componentName ?? Component.displayName ?? Component.name;
+  const contentFn = content ?? createComponent;
```
robertknight, line 43: This could use default assignment in destructuring (`{componentName, content = createComponent, ...}`), to eliminate the `contentFn` variable.

src/components/test/common-tests.js
```diff
@@ -74,54 +94,59 @@ export function testCompositeComponent(
  */
 export function testPresentationalComponent(
   Component,
-  { componentName, elementSelector } = {}
+  { componentName, content, elementSelector } = {}
```
robertknight, line 97: Same comment as above re. the naming of `content` and use of destructuring.

src/pattern-library/components/patterns/input/InputPage.js
```diff
@@ -49,7 +49,28 @@ export default function InputPage() {
           <Library.Example>
             <Library.Demo title="Basic Input" withSource>
               <div className="w-[350px]">
-                <Input placeholder="Placeholder..." />
+                <Input
+                  aria-label="Input example"
+                  placeholder="Placeholder..."
+                />
```
robertknight, line 55: Perhaps we should put some debug checks in the `<Input>` component to check for either a label-able ID or an `aria-label` at runtime.

styles/test.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * Stylesheet for tests
+ */
+
+// legacy component styles
```
robertknight, line 5:
```suggestion
// Legacy component styles
```

Perhaps "Styles for legacy components" and "Styles for new components", to be clear that legacy/updated refers to the components and not just a different set of styles for the same components.

robertknight (approved):
I left some feedback on a few details, please take a look. Looks good otherwise.

src/components/test/common-tests.js
lyzadanger: I thought the same thing! I copied the property name from the `client`'s support for an analogous feature for accessibility tests there. I like `createContent` better, too.

src/pattern-library/components/patterns/input/InputPage.js
lyzadanger: I thought about this! What should the behavior be if both are missing? Console warning? Something more serious?

PR #584, opened 2022-08-25
Replace `*Unstyled` components with extended `*Base` components

lyzadanger:
This is a breaking change.

Related Slack thread here: https://hypothes-is.slack.com/archives/C1M8NH76X/p1661358908157759

This PR removes `ButtonUnstyled` and `LinkUnstyled` and instead exposes the already-extant `ButtonBase` and `LinkBase` components in the API. This introduces a new component sub-category: "base presentational component." These function like a presentational component unless the optional `unstyled` boolean is set. As such:

* `<ButtonBase classes="border">Click me</ButtonBase>` would apply some minimal base classes to the underlying `button` element and extend them with the `border` class provided in the `classes` prop.
* `<ButtonBase classes="border" unstyled>Click me</ButtonBase>` would apply _only_ the `border` class (truly unstyled)

This allows us to have some "base" core styling — in button's case, color transitions (independent of which colors), focus ring customization, a flex layout for content) — for most customized button situations. But there's also the `unstyled` prop should the author really want a fully unstyled button to work with. This keeps some continuity in most customization situations.

It does introduce one layer of abstraction that I want to keep an eye on: the full set of classes applied to a button is no longer visible in one component's context.

Documentation is updated in all affected places.


src/components/input/Button.js
```diff
@@ -41,9 +41,9 @@ const ButtonNext = function Button({
   return (
     <ButtonBase
       {...htmlAttributes}
-      className={classnames(
-        'focus-visible-ring transition-colors whitespace-nowrap rounded-sm',
-        'flex items-center font-semibold',
+      classes={classnames(
+        // NB: Base classes are applied by ButtonBase
```
lyzadanger, line 45: Here's the abstraction. Button base classes are now applied in `ButtonBase`. This is the maximum amount of abstraction and inheritance I feel comfortable with right now. 🖕🏻

src/components/input/ButtonBase.js
```diff
@@ -56,8 +58,19 @@ const ButtonBaseNext = function ButtonBase({
     <button
       role={role ?? 'button'}
       {...ariaProps}
+      data-base-component="ButtonBase"
+      /* data-component will be overwritten unless this component is used directly */
+      data-component="ButtonBase"
```
lyzadanger, line 63: Does this make sense?

`IconButton` renders a `ButtonBase` but provides its own `data-component` value, e.g.:

`<button data-component="IconButton" data-base-component="ButtonBase" ... />`

But if you use `ButtonBase` directly (i.e. for a custom button), you'd get:

`<button data-component="ButtonBase" data-base-component="ButtonBase" />`

src/components/input/ButtonBase.js
```diff
@@ -56,8 +58,19 @@ const ButtonBaseNext = function ButtonBase({
...
       {...htmlAttributes}
-      className={className}
+      className={classNames(
+        {
+          'focus-visible-ring rounded-sm': !unstyled,
+          'transition-colors': !unstyled,
+          // Set layout for button content
+          'whitespace-nowrap flex items-center': !unstyled,
```
lyzadanger, line 70: So, these are the canonized "core styles" for a button in our world. But we don't apply them if `unstyled` is set...

src/components/test/common-tests.js
```diff
@@ -151,34 +151,52 @@ export function testPresentationalComponent(
 /**
  * Set of tests common to all base components
  *
- * @param {FunctionComponent} Component
+ * @param {import('preact').FunctionComponent} Component
```
lyzadanger, line 154: Most of the changes in this module are me finally getting around to extending the "base" common tests to take and use the options object that the other common test suites already use. I'll call out where there's actually a change.

src/components/test/common-tests.js
```diff
@@ -151,34 +151,52 @@ export function testPresentationalComponent(
...
+      const allClasses = [...primaryEl.classList];
+      assert.equal(allClasses.at(-2), 'foo');
+      assert.equal(allClasses.at(-1), 'bar');
+    });
+
+    it('does not apply base classes if `unstyled` is set', () => {
+      const wrapper = createContent(Component, {
+        classes: 'testClass',
+        unstyled: true,
+      });
```
lyzadanger, line 183: This test is new.

src/types.js
```diff
@@ -19,15 +19,9 @@
  */
 
 /**
- * Props common to Base, abstract components. These are not part of the
- * package API.
+ * Props common to Base components.
```
lyzadanger, line 22:
```suggestion
 * Props common to base presentational components.
```

Overlooked this, will fix.

robertknight:
The description of what a "base component" is says:

> A type of presentational component that only applies minimal, core styles for the component's primary element. 

Is there a standard set of things which are "in scope" for this minimal styling, and how likely is this to change over time? Looking at `ButtonBase` and `LinkBase` it seems like minimal styling includes: basic layout of children (flex, centered) and basic accessibility (focus ring when using keyboard navigation, underline on hover). These seems like sensible things to include. I think it would be good to mention them in the "Using components" page if that is the case.

src/components/input/ButtonBase.js
```diff
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
```
robertknight, line 1:
```suggestion
import classnames from 'classnames';
```

Per convention.

src/components/input/ButtonBase.js
robertknight: I think `data-component` alone is fine here, since we're dealing with just one element. The main value in these attributes is being able to see which elements come from which components when looking at the DOM tree.

src/components/test/common-tests.js
```diff
@@ -151,34 +151,52 @@ export function testPresentationalComponent(
...
+ * @param {CommonTestOpts} opts
  */
-export function testBaseComponent(Component) {
-  describe('Common base component functionality', () => {
+export function testBaseComponent(
+  Component,
+  { componentName, createContent = createComponent, elementSelector } = {}
+) {
+  const displayName = componentName ?? Component.displayName ?? Component.name;
+  describe(`Common base presentational functionality for ${displayName}`, () => {
```
robertknight, line 162:
```suggestion
  describe(`Common base functionality for ${displayName}`, () => {
```

"presentational" seems superfluous to me.

src/pattern-library/components/patterns/UsingComponentsPage.js
```diff
@@ -21,14 +21,23 @@ export default function UsingComponentsPage() {
...
-          <li>
-            <strong>Simple components</strong>: Opinionated, simple components
-            that apply design patterns. Example: <code>Spinner</code>.
-          </li>
           <li>
             <strong>Presentational components</strong>: Composable components
             with customization flexibility. Example: <code>Button</code>,{' '}
-            <code>Link</code>.
+            <code>Link</code>. This is the most common category of component.
+            <ul>
```
robertknight, line 28: Indented lists like this could use a bit more vertical spacing around them.

src/pattern-library/components/patterns/UsingComponentsPage.js
```diff
@@ -21,14 +21,23 @@ export default function UsingComponentsPage() {
...
+              <li>
+                <strong>Base presentational components</strong>: A type of
+                presentational component that only applies minimal, core styles
+                for the {"component's"} primary element. This adds flexibility
```
robertknight, line 32: It would be useful to clarify what is "in scope" for "minimal, core styles". This might vary per component, but from the current examples it looks like there are some commonalities (eg. flex layout, whitespace handling, focus ring).

robertknight (commented):
The basic approach here seems reasonable. A couple of thoughts:

- I would be inclined to omit the `data-base-component` attribute, since that will be used on the same element that has `data-component`.
- I think it would be useful for the docs to say more about what is "in scope" for the "minimal, core" base styles

src/components/input/ButtonBase.js
lyzadanger: Right, but because we're introducing this new/first abstraction in which some of the applied CSS styles for `Button`, `IconButton`, e.g. come from `ButtonBase`, I think there's some value in having a `data-base-component` so that the developer locks on to the idea that some of the classes they're seeing there in the DOM inspector are coming from `ButtonBase`. I feel like it's one little extra helper to give context.

src/pattern-library/components/patterns/UsingComponentsPage.js
lyzadanger: Yep! Already tracked, second checklist item here: https://github.com/hypothesis/frontend-shared/issues/587

src/components/input/ButtonBase.js
lyzadanger: Argh. Artifact of editor auto-imports if you can believe it, but yeah, will def. fix.

lyzadanger:
> Is there a standard set of things which are "in scope" for this minimal styling, and how likely is this to change over time? Looking at ButtonBase and LinkBase it seems like minimal styling includes: basic layout of children (flex, centered) and basic accessibility (focus ring when using keyboard navigation, underline on hover). These seems like sensible things to include. I think it would be good to mention them in the "Using components" page if that is the case. Specific base components might have additional "minimal" styles where that makes sense.

This is excellent feedback.

Tell you what, though...I'm dying to slightly restructure this page ("Using components") to use the new added `Section` level — it's a bit poorly-formatted at the moment and it makes adding this documentation in the way I want to awkward.

So, I am adding this as a fourth item to this list of pattern-library housekeeping tasks: https://github.com/hypothesis/frontend-shared/issues/587, which I intend to attend to in the next few days.

robertknight (approved):
Hi Lyza. I leave the `data-base-component` issue to your judgement, and additional documentation around what constitutes "base styles" to https://github.com/hypothesis/frontend-shared/issues/587.

PR #589, opened 2022-08-26
Support native TypeScript modules

robertknight:
Adapt the build/lint/formatting configuration to support native TypeScript modules, following the changes in https://github.com/hypothesis/client/pull/4725. Also convert Spinner.js to Spinner.tsx as a simple test case.

The only extra thing I had to add compared to the client changes was to specify `--extensions` when calling the `babel` CLI tool.


src/components/feedback/Spinner.tsx
```diff
@@ -2,18 +2,15 @@ import classnames from 'classnames';
...
 
-/**
- * @typedef SpinnerProps
- * @prop {'sm'|'md'|'lg'} [size='sm']
- * @prop {'text-light'|'text'|'text-inverted'} [color='text-light']
- */
+type SpinnerProps = {
+  size?: 'sm'|'md'|'lg';
+  color?: 'text-light'|'text'|'text-inverted';
+}
```
robertknight, line 8: TS syntax doesn't have a way to indicate the default value. This could still be put in a JSDoc comment though.

lyzadanger:
Note: to land this, we need to update content globs in the pattern library's tailwind config and update the documentation about tailwind configuration in the pattern library. See https://github.com/hypothesis/client/pull/4737, e.g.

src/components/feedback/Spinner.tsx
lyzadanger: We have been leaning pretty heavily on the commenting element of JSDoc typing. I'll have to do a little poking around to see what common conventions are for documenting/commenting types.

lyzadanger (approved):
Let's get this landed! I rebased and re-tested this morning, and tested with a local install against the `client`, so I'm satisfied...

Issue #603, opened 2022-08-30
Vet icons and potentially remove unused and/or `alt` icons


lyzadanger:
Some of the icons in the icon set are "aspirational". They match the other newer icons in style, but aren't what's actually used in our applications. We should assess the overall set and adjust as needed.

acelaya:
Closed by https://github.com/hypothesis/frontend-shared/pull/1849

PR #604, opened 2022-08-30
Update annotate icons to match what is used in the client

lyzadanger:
I missed the annotate icon during the consolidation the other day because the one in the pattern library was _similar_ but not the _same_ as the one in the client.

Regarding the "Alt" icon: I don't think I've ever seen what was previously the `AnnotateAltIcon` used anywhere, so I've removed it from the library. I know we need to assess this more broadly, ergo https://github.com/hypothesis/frontend-shared/issues/603


robertknight (approved):
To recap a Slack conversation, with this change:

- The Annotate icon now matches what the client uses
- AnnotateAlt is an alternative proposal for this icon, which we might either use to replace Annotate or discard. This review is covered by https://github.com/hypothesis/frontend-shared/issues/603

PR #606, opened 2022-08-31
Add a couple hooks to "next" entry point

lyzadanger:
This PR explicitly adds a couple of hooks to the `next` entry point. The non-inclusion, for now at least, of `useElementShouldClose` is purposeful. I'll evaluate its inclusion as I re-implement modals.

Will be using one of these hooks immediately for https://github.com/hypothesis/client/issues/4739


src/hooks/test/use-arrow-key-navigation-test.js
```diff
@@ -0,0 +1,302 @@
+import { options as preactOptions, render } from 'preact';
```
lyzadanger, line 1: This file was renamed, and an extraneous `describe` block removed, but is otherwise unchanged.

robertknight (approved):

PR #608, opened 2022-08-31
Add `DataTable` component

### LINKED ISSUE #530: Re-implement table component
Part of #455

lyzadanger:
This PR adds the `DataTable` component. It has keyboard navigation support and attempts to get the accessibility basics right. Per a conversation today, the plan is to get this landed — it's an iterative improvement over our last implementation of a `Table` component — and track improvement tasks separately.

Changes since this was last reviewed:

* Expanded documentation around what `selected` and `confirmed` mean and done another pass over documentation in general. `DataTable` is documented separately from `Table`.
* Feedback integrated from last review cycle, all minor. I'll call out a few in comments.
* Associated legacy patterns and component deprecated. Table patterns were the last remaining "patterns" in the project, so I also removed some other pattern-related assets in 9d740966d5b229fec8701cc6ea96daabf1ec59ae (Yay!)

Fixes #530 

🎉


src/components/data/DataTable.js
```diff
@@ -0,0 +1,190 @@
...
+ * @typedef {import('preact').ComponentChildren} Children
+ *
+ * @typedef Column
+ * @prop {string} field
+ * @prop {string} label
+ * @prop {string} [classes]
+ *
+ * @typedef {Record<string, any>} Row
+ *
+ * @typedef {( row:Row, field:string) => Children|null} RenderItemCallback
```
lyzadanger, line 28: Type name changed per feedback

src/components/data/index.js
```diff
@@ -1,4 +1,5 @@
 export { default as AspectRatio } from './AspectRatio';
+export { default as DataTable } from './DataTable';
```
lyzadanger, line 2: Export ordering fixed to be strictly alpha.

src/components/data/test/DataTable-test.js
```diff
@@ -0,0 +1,299 @@
+import { mount } from 'enzyme';
```
lyzadanger, line 1: No changes to tests since last review.

src/pattern-library/components/patterns/samples.js
```diff
@@ -220,3 +220,58 @@ export function sampleTableContent() {
     ],
   };
 }
+
+export function nabokovNovels() {
```
lyzadanger, line 224: No changes here.

src/components/data/DataTable.js
```diff
@@ -0,0 +1,190 @@
+import { useContext, useEffect } from 'preact/hooks';
+
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import { downcastRef } from '../../util/typing';
+import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
+
+import { SpinnerSpokesIcon } from '../icons';
+import ScrollContext from '../data/ScrollContext';
```
robertknight, line 8:
```suggestion
import ScrollContext from './ScrollContext';
```

src/components/data/DataTable.js
```diff
@@ -0,0 +1,190 @@
+import { useContext, useEffect } from 'preact/hooks';
+
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import { downcastRef } from '../../util/typing';
+import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
```
robertknight, line 5: Can you sort these.

src/components/data/DataTable.js
```diff
@@ -0,0 +1,190 @@
...
+ * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLElement>, 'size'|'rows'|'role'|'loading'>} HTMLAttributes
+ * @typedef {import('./TableContext').TableInfo} TableInfo
+ * @typedef {import('preact').ComponentChildren} Children
+ *
+ * @typedef Column
+ * @prop {string} field
+ * @prop {string} label
+ * @prop {string} [classes]
+ *
+ * @typedef {Record<string, any>} Row
```
robertknight, line 26: If you make the row type a generic parameter of `DataTableProps`, then the callbacks props can receive the specific row type that is used in a particular DataTable:

```diff
diff --git a/src/components/data/DataTable.js b/src/components/data/DataTable.js
index f81022a..5d1807e 100644
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -17,28 +17,28 @@ import TableCell from './TableCell';
  * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLElement>, 'size'|'rows'|'role'|'loading'>} HTMLAttributes
  * @typedef {import('./TableContext').TableInfo} TableInfo
  * @typedef {import('preact').ComponentChildren} Children
- *
+ */
+
+/**
  * @typedef Column
  * @prop {string} field
  * @prop {string} label
  * @prop {string} [classes]
- *
- * @typedef {Record<string, any>} Row
- *
- * @typedef {( row:Row, field:string) => Children|null} RenderItemCallback
- * @typedef {( row: Row ) => void } RowCallback
- *
+ */
+
+/**
+ * @template {object} Row
  * @typedef DataTableProps
  * @prop {Column[]} columns
  * @prop {Row[]} rows
  * @prop {string} title
  * @prop {Row|null} [selectedRow]
  * @prop {boolean} [loading=false]
- * @prop {RowCallback} [onConfirmRow] - Callback when a row is "confirmed" by
+ * @prop {(r: Row) => void} [onConfirmRow] - Callback when a row is "confirmed" by
  *   double-click or pressing "Enter"
- * @prop {RowCallback} [onSelectRow] - Callback when a row is "selected" by
+ * @prop {(r: Row) => void} [onSelectRow] - Callback when a row is "selected" by
  *   focus or click
- * @prop {RenderItemCallback} [renderItem] - Callback to render an individual table cell
+ * @prop {(r: Row, field: string) => Children|null} [renderItem] - Callback to render an individual table cell
  * @prop {Children} [emptyMessage] - Content to render if rows is empty (and not
  *   in a loading state)
  */
@@ -46,7 +46,8 @@ import TableCell from './TableCell';
 /**
  * An interactive table of rows and columns with a sticky header.
  *
- * @param {CommonProps & DataTableProps & HTMLAttributes} props
+ * @template {object} Row
+ * @param {CommonProps & DataTableProps<Row> & HTMLAttributes} props
  */
 const DataTableNext = function DataTable({
   children,
@@ -81,7 +82,10 @@ const DataTableNext = function DataTable({
    * @param {string} field
    */
   function formatDataRowItem(row, field) {
-    return renderItem?.(row, field) ?? row[field];
+    return (
+      renderItem?.(row, field) ??
+      /** @type {Children|null} */ (row[/** @type {keyof Row} */ (field)])
+    );
   }
 
   /** @param {Row} row */
diff --git a/src/pattern-library/components/patterns/data/DataTablePage.js b/src/pattern-library/components/patterns/data/DataTablePage.js
index 9764890..9a33811 100644
--- a/src/pattern-library/components/patterns/data/DataTablePage.js
+++ b/src/pattern-library/components/patterns/data/DataTablePage.js
@@ -15,20 +15,16 @@ const nabokovColumns = [
 
 export default function DataTablePage() {
   const [selectedRow, setSelectedRow] = useState(
-    /** @type {import('../../../../components/data/DataTable').Row|null} */ (
+    /** @type {nabokovRows[number]|null} */ (
       nabokovRows[nabokovRows.length - 1]
     )
   );
 
   const [selectedRow2, setSelectedRow2] = useState(
-    /** @type {import('../../../../components/data/DataTable').Row|null} */ (
-      null
-    )
+    /** @type {nabokovRows[number]|null} */ (null)
   );
   const [confirmedRow, setConfirmedRow] = useState(
-    /** @type {import('../../../../components/data/DataTable').Row|null} */ (
-      null
-    )
+    /** @type {nabokovRows[number]|null} */ (null)
   );
   return (
     <Library.Page
```

src/components/data/DataTable.js
```diff
@@ -0,0 +1,190 @@
...
+    onSelectRow?.(row);
+  }
+
+  /** @param {Row} row */
+  function confirmRow(row) {
+    onConfirmRow?.(row);
+  }
+
+  /** @param {KeyboardEvent} event */
+  function handleKeyDown(event, row) {
```
robertknight, line 98: Param type for `row` is missing here. The `noImplicitAny` setting is currently off for this project in tsconfig.json, so TS doesn't complain. We should address that at some point.

src/components/data/DataTable.js
```diff
@@ -0,0 +1,190 @@
...
+      event.stopPropagation();
+    }
+  }
+
+  // Ensure that a selected row is visible when this table is within
+  // a scrolling context
+  useEffect(() => {
+    if (!selectedRow || !scrollContext) {
+      return;
+    }
```
robertknight, line 111: It might be possible to implement this scroll-into-view logic without needing to be passed in a `ScrollContext`, eg. by using `Element.scrollIntoViewIfNeeded()` (non-standard, but widely implemented and unlikely to ever be removed) or using hit-testing. I think that's an exercise for a future iteration though.

src/components/data/test/DataTable-test.js
```diff
@@ -0,0 +1,299 @@
...
+      // Initially select the last data row
+      const [selectedRow, setSelectedRow] = useState(fakeRows[4]);
+      return (
+        <Scroll>
+          <DataTable
+            onSelectRow={setSelectedRow}
+            rows={fakeRows}
+            columns={fakeColumns}
+            selectedRow={selectedRow}
+            renderItem={(row, field) => <div>{row[field]}</div>}
```
robertknight, line 226: This test doesn't set an explicit height for rows that I can see. Does that mean the row height could vary depending on eg. the font size? If so it would probably be wise to set an explicit height here to make it easier to reason about what rows are visible at any time.

src/pattern-library/components/patterns/data/DataTablePage.js
```diff
@@ -0,0 +1,440 @@
...
+const nabokovRows = nabokovNovels();
+const nabokovColumns = [
+  { field: 'title', label: 'Title' },
+  { field: 'year', label: 'Year' },
+  { field: 'language', label: 'Language' },
+];
+
+export default function DataTablePage() {
+  const [selectedRow, setSelectedRow] = useState(
+    /** @type {import('../../../../components/data/DataTable').Row|null} */ (
```
robertknight, line 18: See comments about `Row` type in `DataTable.js`.

src/pattern-library/components/patterns/data/DataTablePage.js
```diff
@@ -0,0 +1,440 @@
...
+            </Library.Demo>
+          </Library.Example>
+
+          <Library.Example title="renderItem">
+            <p>
+              <code>renderItem</code> is a callback for formatting the contents
+              of an individual table cell. It is given the current{' '}
+              <code>row</code> and the <code>field</code> to format. Returning{' '}
+              <code>null</code> will apply default formatting for the given{' '}
+              <code>field</code> value.
```
robertknight, line 236: This might cause some confusion if a caller wanted to render nothing for a particular cell (since returning `null` is the normal way for a rendering function to generate no output). Since the default formatting is just `row[field]`, we could move the responsibility for doing that to the `renderItem` callback if provided. In other words, if the caller provides a `renderItem` callback, it must handle all fields. Otherwise the default formatting is used for all fields. An alternative would be to specify/document that callers should return an empty string or other non-nullish but empty value in this case.

robertknight (commented):
I had some comments on the `Row` type for DataTable, as well as what value `renderItem` should return if it wants to render nothing for a particular cell. Otherwise this all looks good.

src/components/data/DataTable.js
lyzadanger: This patch gives me a type error on:

image

I'll poke at it a bit, I guess...

robertknight: I don't think I saw this yesterday. However this error makes sense as the types of property values of `Row` is not known here. Converting to `unknown` per the error message is one option (this is unfortunately rather verbose in JSDoc. In TS it would be `row[field as keyof Row] as unknown as ComponentChildren|null`. The lazy option would be to just cast to `any`.

src/components/data/test/DataTable-test.js
lyzadanger: I concur that the arithmetic ends up being fiddly and non-round. However, there's more (dimension-related) styling at play here than can be controlled from the `renderItem` callback. I can, for example, update it to:

renderItem={(row, field) => <div style="height:50px">{row[field]}</div>}

But this doesn't do what we might like it to — it results in a row height of `74px`. That's because there's other padding and whatnot in play here.

If we had a `renderRow` callback, we could precisely control row height. I don't think it's quite worth the tangent just now.

src/pattern-library/components/patterns/data/DataTablePage.js
lyzadanger: Yeah, you're right. This has had a minor code smell the whole time. Let's assert that a `renderItem` callback needs to handle all fields.

lyzadanger:
@robertknight Let's see if we can get this bad boy landed!

I've integrated the last round of feedback, the most of material of which were:

* Typing: I've applied your patch for `Row` typing
* `renderItem` is now expected to render all fields, not just selective ones

src/components/data/DataTable.js
```diff
@@ -0,0 +1,184 @@
...
+          rows.map((row, idx) => (
+            <TableRow
+              key={idx}
+              selected={row === selectedRow}
+              onClick={() => selectRow(row)}
+              onFocus={() => selectRow(row)}
+              onDblClick={() => confirmRow(row)}
+              onKeyDown={event => handleKeyDown(event, row)}
+            >
+              {fields?.map(field => (
```
robertknight, line 162:
```suggestion
              {fields.map(field => (
```

`fields` is always an array.

src/components/data/test/DataTable-test.js
robertknight: Setting an explicit height in `renderItem` would at least produce a known _minimum_ height, and then a certain amount of padding could be assumed by the test with some tolerance for future minor style tweaks.

src/pattern-library/components/patterns/data/DataTablePage.js
```diff
@@ -0,0 +1,434 @@
...
+              compatible with scrolling contexts. The examples on this page are
+              constrained within <code>Scroll</code> components.
+            </p>
+          </Library.Example>
+        </Library.Pattern>
+
+        <Library.Pattern title="Props">
+          <Library.Example title="rows">
+            <p>
+              <code>rows</code> is an Array of basic objects mapping field names
```
robertknight, line 161:
```suggestion
              <code>rows</code> is an Array of basic objects mapping field names
```

I can't remember if I mentioned this last time around, but there is no hard requirement for objects to be basic/plain as far as I can see. Class instances for example would work fine.

robertknight (approved):
Changes LGTM. Found one tiny issue.

PR #614, opened 2022-09-07
Enable `noImplicitAny` TS option

robertknight:
Fix the remaining errors when the `noImplicitAny` TS option is enabled, and then remove the override that turned it off. This matches the TS settings in our other repos and will catch missing/incorrect types.

### DIFF
--- a/src/components/input/Input.js
+++ b/src/components/input/Input.js
@@ -28,6 +28,7 @@ const InputNext = function Input({
 
   ...htmlAttributes
 }) {
+  // @ts-expect-error - "aria-label" is missing from HTMLInputAttributes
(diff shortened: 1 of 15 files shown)


src/components/input/Input.js
```diff
@@ -28,6 +28,7 @@ const InputNext = function Input({
 
   ...htmlAttributes
 }) {
+  // @ts-expect-error - "aria-label" is missing from HTMLInputAttributes
```
robertknight, line 31: An alternative solution here would be to cast `htmlAttributes` to some other type first.

lyzadanger (approved):
Sure, OK, let's do this!

Issue #627, opened 2022-09-12
Checkbox component does not have a visible focus ring


lyzadanger:
When navigating by keyboard, the Checkbox component (both the legacy version and the updated, more relevant, variant) do not have a visible focus ring. Checkbox instances are focus-able, and respond to keyboard input to check/uncheck the underlying `input type="checkbox"` element, but there is no visible ring.

Issue #529, opened 2022-07-29
Re-implement dialog and modal components


lyzadanger:
Part of #455

lyzadanger:
While there is still some ways to go to make our `Modal` as accessible as we'd like it, there is now an updated `Modal` component!

PR #628, opened 2022-09-12
WIP: Updated Dialog component

lyzadanger:
This is a draft PR to demonstrate the current approach for a replacement for the legacy `Dialog` and `Modal` components with a single `Dialog` component.

Please check out this branch and visit http://localhost:4001/feedback-dialog to get a sense of how the `Dialog` component is intended to work. There's a lot of documentation information there.[^1]

I intend to break this work down into multiple PRs, or at the very least, open a fresh PR when it's time to go for reals on this.

## General approach

* Focus capture and management via the `useFocusCapture` hook. To construct this, I generalized the guts of the pre-existing `useArrowKeyNavigation` hook into an underlying `useKeyboardNavigation` hook, used by both `useArrowKeyNavigation` and the new `useFocusCapture` hook.
* Slightly more onus put on authors using `Dialog` to handle some basic accessibility things, especially setting an `aria-describedby` and `focusRef` (this is covered in the documentation). The previous implementation tried to make an educated guess about some of this, but I think asking humans to configure this is reasonable. We don't have _that_ many dialogs! This process is aided by the new component's support for HTML attributes.
* New hook `useClickAwayListener` based on pre-existing `useElementShouldClose`. Refactored to use `ListenerCollection` and to not conflate "press-escape-to-close" into itself, as that is a separate concern.
* Significant overhaul of configuration for sizing, and significant improvements to responsive resilience. Dependably sizing dialogs should, I hope, be considerably less fiddly in the future.

## Changes from legacy variant

There are a couple of visual changes that I want to call out:

* Positioning of the dialog vis-a-vis the viewport used to depend on whether the viewport was "tall" or not, with "shorter" viewports vertically-centering the dialog and "taller" ones positioning it closer to the top of the viewport. Now dialogs are always positioned relatively close to the top of the viewport. This is more consistent and provides less complexity in terms of responsiveness.
* The border under the title of a dialog now extends all the way to the edges of the dialog, whereas it used to have some space around it. This is to make scroll shadows not look stupid, when they are present (see the scrolling-modal example in the pattern library).

## Limitations/Not done yet

* In developing this, I stumbled upon a [deficiency in Checkbox components: they don't have a visible focus ring](https://github.com/hypothesis/frontend-shared/issues/627). You may notice this if you are keyboard-navigating some of the dialog examples.
* I would like to, at some point, consider disabling body scrolling when dialogs are open. But there is enough going on for now...

[^1]: Small note: The `JSXToString` functionality doesn't make it easy to understand what's going with `ref`s in some of the examples. In the "Dialog with accessible interactive elements" example, the `Dialog`'s `focusRef` value and the `Input`'s `elementRef` should both point to the same `ref` reference. I haven't looked into solving this yet.

Part of #529


src/hooks/keyboard-navigation.js
```diff
@@ -47,33 +69,40 @@ function isElementVisible(element) {
...
+    active = true,
     autofocus = false,
-    horizontal = true,
-    vertical = true,
-    selector = 'a,button',
+    autofocusRef,
+    horizontal = false,
+    tab = false,
+    vertical = false,
+    selector = 'button, [href], input, select, textarea, [tabindex]',
```
lyzadanger, line 83: Here's that selector change I proposed verbally last week.

src/hooks/keyboard-navigation.js
```diff
@@ -47,33 +69,40 @@ function isElementVisible(element) {
...
   } = {}
 ) {
   useEffect(() => {
-    if (!containerRef.current) {
+    if (!active) {
+      return () => {};
+    }
+
+    if (active && !containerRef.current) {
```
lyzadanger, line 91: I suppose the `active` part of this conditional is extraneous :).

src/hooks/keyboard-navigation.js
```diff
@@ -47,33 +69,40 @@ function isElementVisible(element) {
...
 
     const getNavigableElements = () => {
       const elements = /** @type {HTMLElement[]} */ (
         Array.from(container.querySelectorAll(selector))
       );
+      // If the container itself has a `tabIndex`, include it in the
+      // set of navigable elements
+      if (container.getAttribute('tabIndex')) {
+        elements.unshift(container);
+      }
```
lyzadanger, line 105: This allows the container itself to be part of the navigation sequence, allowing for cases in which we need to focus the dialog container itself on open (when no `focusRef` is provided).

src/util/listener-collection.js
```diff
@@ -40,7 +40,7 @@ export class ListenerCollection {
    * @param {Target} eventTarget
    * @param {Type} eventType
    * @param {(event: EventType<Target, Type>) => void} listener
-   * @param {AddEventListenerOptions} [options]
+   * @param {AddEventListenerOptions & { useCapture?: boolean }} [options]
```
lyzadanger, line 43: I don't know why TypeScript doesn't include more properties in its  [`AddEventListener` interface](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L25-L29), but it doesn't.

robertknight: According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#syntax) the property is called `capture` rather than `useCapture`. `useCapture` was the name of the old API. Browser support for `capture` as a property looks good.

robertknight: I don't know why this is missing from TS types. It was added to the spec quite a long time ago.

PR #629, opened 2022-09-13
Add `useFocusCapture` hook

lyzadanger:
Factor out a common `useKeyboardNavigation` hook, used by existing `useArrowKeyNavigation` hook and new `useFocusCapture`. This is part of work supporting a new `Dialog` component.

This was discussed in #628


src/hooks/test/use-arrow-key-navigation-test.js
```diff
@@ -1,302 +0,0 @@
...
-  // Workaround for an issue with `useEffect` throwing exceptions during
-  // `act` callbacks. Can be removed when https://github.com/preactjs/preact/pull/3530 is shipped.
-  let prevDebounceRendering;
-  beforeEach(() => {
-    prevDebounceRendering = preactOptions.debounceRendering;
-  });
-  afterEach(() => {
-    preactOptions.debounceRendering = prevDebounceRendering;
-  });
-
```
lyzadanger, line 47: I double-checked that this has shipped and, yep, it has. I removed this from the updated tests.

src/hooks/test/keyboard-navigation-test.js
```diff
@@ -0,0 +1,459 @@
...
+        Help
+      </a>
+    </div>
+  );
+}
+describe('keyboard-navigation hooks', () => {
+  function currentItem() {
+    return document.activeElement.innerText;
+  }
+  describe('useArrowKeyNavigation', () => {
```
lyzadanger, line 28: There is only one small change to the actual tests for `useArrowKeyNavigation`, which I'll call out in a comment.

src/hooks/test/keyboard-navigation-test.js
```diff
@@ -0,0 +1,459 @@
...
+      {
+        key: 'ArrowDown',
+        vertical: false,
+        shouldHandle: false,
+      },
+      // Tab key should not be handled
+      {
+        key: 'Tab',
+        shouldHandle: false,
+      },
```
lyzadanger, line 172: This is the one test change here (addition).

src/hooks/test/keyboard-navigation-test.js
```diff
@@ -0,0 +1,459 @@
...
+      assert.equal(italicButton.tabIndex, -1);
+
+      boldButton.disabled = true;
+
+      // nb. tabIndex update is async because it uses MutationObserver
+      await waitFor(() => italicButton.tabIndex === 0);
+    });
+  });
+
+  function Modal({
```
lyzadanger, line 298: Everything from here is new.

lyzadanger (commented):
Irritatingly, git did not follow the test module rename in this case so the diff is a bit annoying. I've tried to call out relevant things in comments.

src/hooks/keyboard-navigation.js
```diff
@@ -47,33 +69,40 @@ function isElementVisible(element) {
...
- *   @param {boolean} [options.vertical] - Enable navigating elements using up/down arrow keys
- *   @param {string} [options.selector] - CSS selector which specifies the
- *     elements that navigation moves between
+ * @param {KeyboardNavigationOptions} options
  */
-export function useArrowKeyNavigation(
+function useKeyboardNavigation(
   containerRef,
   {
+    active = true,
```
lyzadanger, line 77: ARGH! Stupid codecov differences between dev and CI. Local was showing 100% coverage. I'll fix this after getting PR feedback.

src/hooks/keyboard-navigation.js
```diff
@@ -47,33 +69,40 @@ function isElementVisible(element) {
...
     autofocus = false,
-    horizontal = true,
-    vertical = true,
-    selector = 'a,button',
+    autofocusRef,
+    horizontal = false,
+    tab = false,
+    vertical = false,
+    selector = 'button, [href], input, select, textarea, [tabindex]',
```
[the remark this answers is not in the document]
robertknight: Some alternative ways I can think of to handle focus capture:

1. Use the native `<dialog>` element and its `showModal()` API. This is now supported in all browsers, but was only recently added to Firefox and Safari (~ March 2022). This involves the least amount of code, but needs a polyfill.
2. Listen for `focus` events on the document and cancel them if they would cause an element outside of the dialog to be focused. Google's [polyfill for dialog](https://github.com/GoogleChrome/dialog-polyfill/blob/master/index.js) does this. It also listens for `keydown` events on the document to detect if the user pressed Tab vs Shift+Tab just before the focus event.

robertknight (commented):
Although keyboard handling within a dialog may seem like an extension of arrow key navigation, I think it will make sense to separate the implementation and to avoid modifying `tabIndex` of elements within the dialog as part of the focus-capturing hook. One reason is to avoid conflicts over the `tabIndex` property if a dialog embeds a component that tries to use arrow key navigation itself (eg. a table or toolbar). Another reason is that focus capturing is a common UI behavior is becoming available as a built-in feature in browsers (eg. via `<dialog>` / `inert` attribute). We might not be able to use the native implementation everywhere today, but it would be a good idea to have the freedom to use it in future and I suspect that will be easier if the code is not part of code that deals with arrow-key navigation.

src/hooks/keyboard-navigation.js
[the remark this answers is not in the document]
lyzadanger: @robertknight If we went the native `dialog`-plus-polyfill route, what's your thinking on how to manage that polyfill? Would it be the responsibility of this package to register and provide this polyfill (and/or any other polyfills that come up in our journeys pertaining to common UI components)? Do you have a rough diagram in your mind?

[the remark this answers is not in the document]
robertknight: If we used https://github.com/GoogleChrome/dialog-polyfill#usage, that includes an ESM build that can be imported like any other module. I think we could then use it something like this:

```js
import dialogPolyfill from 'dialog-polyfill';
...
function MyDialog() {
  const dialogRef = useRef();
  useLayoutEffect(() => {
    dialogPolyfill.registerDialog(dialogRef.current);
  }, []);
  return <dialog ref={dialogRef}>...</dialog>
}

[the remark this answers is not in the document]
robertknight: This polyfill is quite small so I think we can get away without loading it conditionally. Another way to load it might be to have a function which gets called the first time that a component using `<dialog>` gets rendered, in order to install a global polyfill that monitors whenever `<dialog>` elements are added to the page. However I tend to prefer making polyfills work like any other code as far as possible (in terms of how they are imported and called), except that they defer to a native implementation if available.

PR #631, opened 2022-09-16
Do not set border-radius in component base styles

lyzadanger:
In doing some quick assessment and overview of how we might plan the conversion of our apps over to the new component set, I spotted a couple of use cases in which setting `border-radius` rules (`rounded-*` rules in tailwind parlance) in base styles was undesirable. When reviewing the conventions for base styling — setting core layout, transition configuration, focus ring styling, but not metrics, colors, etc.), it seemed that these rules should not be present on base components.

These changes move `rounded-*` classes from base components to presentational components.


robertknight (approved):
Makes sense.

PR #632, opened 2022-09-16
Add BucketButton component and documentation

lyzadanger:
This PR adds a `BucketButton` component to the `navigation` component group.

## Why (on earth) this work?

There are several reasons I opted to chunk this bit of work out:

* `Buckets.scss` is the last remaining external UI-component SASS module in the `client` application: this converts over to preferred tailwind inline utility class usage
* Encapsulating this design within a component and removal of last layer of abstraction allows for cleaner separation of styling concerns. The previous CSS abstractions masked that the button styling was tackling some things — specifically positioning — that are more appropriately handled on the containing element/component
* Gives us an opportunity (as a sort of test case) to see what brief pattern-library documentation of patterns like this could look like. The associated pattern-library page attempts to explain in brief what a "bucket" is and what these buttons are intended for.
* Gives us an opportunity to place the design pattern in context with, e.g., other buttons and UI

On this branch, you can see the bucket buttons at http://localhost:4001/navigation-buckets

### DIFF
--- a/src/components/navigation/BucketButton.js
+++ b/src/components/navigation/BucketButton.js
@@ -1,0 +1,124 @@
+import classnames from 'classnames';
+
+import { downcastRef } from '../../util/typing';
+
+import ButtonBase from '../input/ButtonBase';
+
+/**
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('../input/ButtonBase').HTMLButtonAttributes} HTMLButtonAttributes
+ * @typedef {import('../input/ButtonBase').ButtonCommonProps} ButtonCommonProps
+ *
+ * @typedef BucketButtonProps
+ * @prop {'left'|'up'|'down'|'lozenge'} [variant='lozenge']
+ */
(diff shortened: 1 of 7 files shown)


src/components/navigation/BucketButton.js
```diff
@@ -0,0 +1,124 @@
...
+
+import ButtonBase from '../input/ButtonBase';
+
+/**
+ * @typedef {import('../../types').PresentationalProps} CommonProps
+ * @typedef {import('../input/ButtonBase').HTMLButtonAttributes} HTMLButtonAttributes
+ * @typedef {import('../input/ButtonBase').ButtonCommonProps} ButtonCommonProps
+ *
+ * @typedef BucketButtonProps
+ * @prop {'left'|'up'|'down'|'lozenge'} [variant='lozenge']
```
lyzadanger, line 13: I agonized on this a bit. `variant` is a prop name consistent with other buttons and presentational components, but `direction` sure feels more natural...

PR #679, opened 2022-10-14
Don't use styles from `ButtonBase` in other Button components

### LINKED ISSUE #650: `inline` prop for `LinkButton` is not working
The `inline` prop for `LinkButton` is broken. A `LinkButton` with `inline` set will still have `display: flex`, which will override any `inline` class set. This is a result of the move to applying base classes to all styled buttons via `ButtonBase`, which sets a flex layout, always. Ah, the pitfalls of introducing even a single tiny abstracted inherited set of styles.

My swag is that it would be best for `LinkButton` to use an `unstyled` `ButtonBase` and manage all of its styles locally.

lyzadanger:
The abstraction of "base button classes" to `ButtonBase` led to a bug, #650, where `LinkButton` attempts to apply both `inline` and `flex` layout to a `button` when the `inline` prop is set. Remove this abstraction so that devs can see the entire styling context and avoid similar bugs in future.

This is less DRY but more safe and clear.

Fixes #650

Before fix:

image

After fix:

Screen Shot 2022-10-14 at 2 30 13 PM

### DIFF
--- a/src/components/input/Button.js
+++ b/src/components/input/Button.js
@@ -42,7 +42,7 @@ const ButtonNext = function Button({
     <ButtonBase
       {...htmlAttributes}
       classes={classnames(
-        // NB: Base classes are applied by ButtonBase
+        'focus-visible-ring transition-colors whitespace-nowrap flex items-center',
         'font-semibold rounded-sm',
         {
           // Variants
@@ -64,6 +64,7 @@ const ButtonNext = function Button({
       pressed={pressed}
       title={title}
       data-component="Button"
+      unstyled
     >
       {Icon && <Icon className="w-em h-em" />}
       {children}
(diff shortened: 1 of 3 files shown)


src/components/input/Button.js
```diff
@@ -42,7 +42,7 @@ const ButtonNext = function Button({
     <ButtonBase
       {...htmlAttributes}
       classes={classnames(
-        // NB: Base classes are applied by ButtonBase
+        'focus-visible-ring transition-colors whitespace-nowrap flex items-center',
```
robertknight, line 45: If this common set of styles is not part of the base component, it could potentially be shared via a package-internal constant as well.

robertknight (approved):

PR #725, opened 2022-11-16
Adjust typing on some `Ref` and `RefObject` properties

lyzadanger:
This allows better compatibility with refs generated by `useRef()` in applications.


robertknight (approved):
LGTM. I note that the internals of these hooks already allow for the possibility of the ref being unset.

PR #786, opened 2023-01-10
Update release process and docs to work with GH rebase-and-merge method

lyzadanger:
If we want to use `rebase-and-merge` as our GitHub merge method, we need to adjust our release process to ensure that the version tag points at the intended commit. Rebase-and-merge rewrites commit hashes, so tags should not be created until after the target commit is present on the `main` branch.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,34 +1,37 @@
-# Release process
+# How to version and publish `frontend-shared`
 
-## 1. Version and tag locally
+## Summary
 
-We use [Semantic Versioning](https://semver.org/#semantic-versioning-200).
+1. **Update the package version** in `package.json` and merge that change into the `main` branch. We use [Semantic Versioning](https://semver.org/#semantic-versioning-200).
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -1,34 +1,37 @@
...
-1. Create a new working branch
-2. Run
+## 1. Update the package version
+
+1. Create a new git branch.
+2. On the new branch, run:
 
    ```shell
-   $ yarn version
+   $ yarn version --no-git-tag-version
```
lyzadanger, line 14: `yarn version` isn't giving us a whole lot at this point: it prompts for a new version number interactively, updates `package.json` and commits it. But it's "free" and a consistent process, so why not. The `--no-git-tag-version` option does what it sounds like it would: it suppresses the creation of a git tag.

acelaya (approved):
This makes a lot of sense to me.

I've always found a bit odd that `yarn/npm version` try to take control of git tags, when IMHO, git tags should be the source of truth. You create the git tag, and that results in a new package version/release being published as a result.

In some projects, I've even taken a more radical approach in which I just ignore the `version` field in the package.json, and publish new packages with `npm version ${RESOLVE_LAST_GIT_TAG_SOMEHOW} --git-tag-version=false`. This is also not ideal, as can be confusing to gradually see the package.json version gradually get super outdated,

PR #787, opened 2023-01-10
Fix alignment of CardHeader close button; convert to TS

### LINKED ISSUE #677: `CardHeader` should align its close button further to the right
`CardHeader` (used in `Panel`) currently right-aligns its close button:

Screen Shot 2022-10-12 at 3 46 42 PM

However, there is padding in the close button that makes the alignment look wrong. This is more striking when using `Panel` in-situ in the sidebar interface.

Instead, `CardHeader` should adjust the right margin of the close button (or similar) such that the "x" in the button appears right-aligned:

Screen Shot 2022-10-12 at 3 46 19 PM

This is a blocker for updating application components to use the new `Panel`, as it looks pretty wonky. Fortunately, tiny fix.

lyzadanger:
Use a negative right margin make the close button (when present) in CardHeader appear right-aligned with the bottom border. Without this, the padding in the button causes it to look like the button has too much space on the right.

Landing this will allow us to move forward with updating `client` components that use the `Panel` shared component.

Opportunistically also converted `CardHeader` to TS.

Fixes #677

## Testing

Check out this branch, run `make dev` and visit http://localhost:4001/layout-card to see the `CardHeader` examples. Note the difference in alignment of the close button between this branch and `main`. You can also see the difference on the [`Panel` pattern library page](http://localhost:4001/layout-panel) page (`Panel` uses `CardHeader`).

### DIFF
--- a/src/components/layout/CardHeader.tsx
+++ b/src/components/layout/CardHeader.tsx
@@ -1,5 +1,7 @@
 import classnames from 'classnames';
+import type { JSX } from 'preact';
 
+import type { PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 
 import { CancelIcon } from '../icons';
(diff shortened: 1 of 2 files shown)


src/components/layout/CardHeader.tsx
```diff
@@ -1,26 +1,32 @@
...
+   * will be rendered.
+   */
+  onClose?: () => void;
+};
+
+type HTMLAttributes = JSX.HTMLAttributes<HTMLElement>;
+
+export type CardHeaderProps = PresentationalProps &
+  ComponentProps &
+  HTMLAttributes;
```
lyzadanger, line 26: It's very early days for TypeScript conversion of these package components, so I imagine we'll feel our way a bit for the best (clearest) way to express component prop types. We can express `CardHeaderProps` without defining `ComponentProps` and `HTMLAttributes` types separately, but I felt like this was a little easier to read.

acelaya (approved):
Looks good and works as expected.

PR #797, opened 2023-01-20
Add PointerButton component and documentation

lyzadanger:
This PR adds a new component to the "navigation" group: `PointerButton`. This component is intended to "point at" and navigate to groups of items elsewhere or off-screen in the UI. For more context, run `make dev` on this branch and visit http://localhost:4001/navigation-pointerbutton

The bulk of this work was completed a couple of months ago but I'd like to rehydrate it to help us complete out the transition of `client` components to updated shared components.

image

This component encapsulates part of the current [Buckets](https://github.com/hypothesis/client/blob/33ca18cb831ab854a64fe20b0f85097780a1c3cd/src/annotator/components/Buckets.js) component in the `client`. Whereas the `Buckets` component currently conflates button styling and positioning, and applies this via a convoluted use of the deprecated `LabeledButton` component, this new component encapsulates just the button pattern — leaving positioning up to the consumer — and fits the pattern in with the context of other application design patterns.

There's no evading that the CSS for this component is complex. It predates my tenure (I'm merely carrying it forward and trying to clean it up as best I can), and I haven't yet seen a simpler drop-in approach.

In a nutshell, this CSS pattern styles borders on absolutely-positioned `::before` and `::after` pseudo-elements to form the "points" on the button's arrows. https://css-tricks.com/snippets/css/css-triangle/ is cited in the component's commenting. Our particular use case is complicated because we have a "fill" color that is different than the button's border, and the button has more content and shape than just a triangle. I don't expect the reviewer to granularly understand every CSS rule here, but instead to verify that the linked example and comments in the component provide enough guideposts for a future dev.

### DIFF
--- a/src/pattern-library/components/patterns/navigation/PointerButtonPage.js
+++ b/src/pattern-library/components/patterns/navigation/PointerButtonPage.js
@@ -1,0 +1,141 @@
+import { PointerButton } from '../../../../next';
+import Library from '../../Library';
+import Next from '../../LibraryNext';
+
+export default function PointerButtonPage() {
+  return (
+    <Library.Page
+      title="PointerButton"
+      intro={
+        <>
+          <p>
+            <code>PointerButton</code> is a component for styling a button that
+            points towards a group of items elsewhere in the UI or off-screen.
+            They are specifically used to indicate and navigate to
+            &quot;buckets&quot; (groups of annotations that are near each other
+            in document content) in the Hypothesis client.
+          </p>
+          <p>
+            PointerButtons point at and navigate to these groups of annotation
+            highlights, which are either in the visible content of the source
+            document (guest page) or off-screen (above or below).
+          </p>
+        </>
+      }
+    >
+      <Library.Section
+        title="PointerButton"
+        intro={
+          <p>
+            <code>PointerButton</code> is a presentational component.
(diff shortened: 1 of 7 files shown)


styles/library.scss
```diff
@@ -5,6 +5,10 @@
 @layer components {
   // Support some basic text styling in Library components
   .styled-text {
+    :where(a):not(:where([class~='unstyled-text'] *)) {
+      @apply text-brand underline hover:text-brand-dark transition-colors;
+    }
+
```
lyzadanger, line 11: I should break this out into a separate commit. The documentation (pattern library page) for this component used a basic `a` tag and it needed some styling for underline and coloring (i.e. this was the first time I actually used an `a` in the docs and ran into a styling gap).

src/pattern-library/components/patterns/navigation/PointerButtonPage.js
```diff
@@ -0,0 +1,141 @@
+import { PointerButton } from '../../../../next';
```
lyzadanger, line 1: Much of the structure of `*Page` components are boilerplate; they should be reviewed for their content, not their code, roughly speaking.

acelaya (approved):
Looks good to me

PR #806, opened 2023-01-24
Update plop component templates to TS

lyzadanger:
This is just a dull little housekeeping task to assist us with creating future components (upcoming: Dialog and Modal).

Update the [plop](https://plopjs.com/) templates such that the components output TSX modules. We will want to update other JS output to TS(X) in the future.

Local docs https://github.com/hypothesis/frontend-shared/blob/main/docs/developing.md


acelaya (approved):
Very interesting. I didn't know this tool 😃

PR #808, opened 2023-01-26
Improve `PlaygroundRoute`, add flexibility for custom routes

lyzadanger:
The pattern library is designed (at least semi-designed) to be used within consuming projects, and allows projects to add "extra routes" to the set of its own common routes, for prototyping or documenting UI patterns specific to the given project. (See LMS, which has a single custom route to demonstrate error dialogs). It's quite rudimentary.

This PR makes it just a bit easier to provide custom routes the pattern library ("PlaygroundApp") by removing the asinine requirement to provide a `group` for each custom route (i.e. create a custom route type that doesn't require that field). Also add the option to customize the navigation list heading for the group of provided custom routes.

This is a little bit hard to conceptualize in the abstract. If I update `scripts/pattern-library` in this package to set a custom route, like so:

```js
// Entry point for local webserver pattern-library bundle
import { startApp } from '../src/pattern-library';

import UserPreferencesPage from '../src/pattern-library/components/prototypes/UserPreferences';

const extraRoutes = [
  {
    title: 'User Preferences',
    component: UserPreferencesPage,
    route: '/prototype-preferences',
  },
];

startApp({ extraRoutes, extraRoutesTitle: 'Prototypes' });
```

The (partial) navigation will look like this:

image

### Why now?

The screenshot might give it away, but I'd like to do some UI prototyping in the pattern library for a current project, and this makes it a little easier.

### DIFF
--- a/src/pattern-library/components/PlaygroundApp.js
+++ b/src/pattern-library/components/PlaygroundApp.js
@@ -8,6 +8,7 @@ import { useRoute } from '../router';
 
 /**
  * @typedef {import("../routes").PlaygroundRoute} PlaygroundRoute
+ * @typedef {import("../routes").CustomPlaygroundRoute} CustomPlaygroundRoute
  * @typedef {import("../routes").PlaygroundRouteGroup} PlaygroundRouteGroup
  * @typedef {import('preact').ComponentChildren} Children
  */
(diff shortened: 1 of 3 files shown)


acelaya (approved):
Looks good

PR #809, opened 2023-01-26
Convert pattern-library base components  and routes to TS

lyzadanger:
Continuing my attempt to convert things to TS as I encounter them. Convert the modules modified in #808 to TS. Depends on #808


src/pattern-library/components/PlaygroundApp.tsx
```diff
@@ -1,45 +1,34 @@
...
- * @prop {CustomPlaygroundRoute[]} [extraRoutes] -
- *   Local-/application-specific routes to add to this pattern library in
- *   addition to the shared/common routes
- * @prop {string} extraRoutesTitle - Title to show above any extra routes
- * provided
- */
+export type PlaygroundAppProps = {
+  baseURL?: string;
+  extraRoutes?: CustomPlaygroundRoute[];
+  extraRoutesTitle?: string;
```
acelaya, line 14: Should we maybe keep the property comments?

acelaya, line 14: Should we maybe keep the property comments?

src/pattern-library/components/PlaygroundApp.tsx
```diff
@@ -1,45 +1,34 @@
...
   extraRoutesTitle,
-}) {
+}: PlaygroundAppProps) {
   const routes = getRoutes();
 
   // Put all of the custom routes into the "custom" group
   const customRoutes = extraRoutes.map(route => {
-    return /** @type PlaygroundRoute */ ({ ...route, group: 'custom' });
+    return { ...route, group: 'custom' } as PlaygroundRoute;
   });
```
acelaya, line 32: It's probably safer to add the return type to the callback, otherwise we might be returning the wrong properties here by mistake, and the type check would not warn about it:

```ts
const customRoutes = extraRoutes.map((route): PlaygroundRoute => {
  return { ...route, group: 'custom' };
});
```

acelaya (approved):
Just a minor comment on a type check, but other than that, looks good to me.

lyzadanger:
@acelaya I pushed another commit that attempts to consolidate some types. See what you think...

acelaya (approved):
Looks good

PR #813, opened 2023-01-27
Add `Tab` and `TabList` components

lyzadanger:
This PR adds new `Tab` and `TabList` components, and pattern-library documentation for same.

Why now: I believe we may wish to use a tabbed interface in upcoming new UI in the client and before building out new tabs, I wanted to:

* Extract the design pattern: Styling is based on the sidebar tabs (`SelectionTabs`) in the client, and extraction helps to better encapsulate the design pattern vs. the `SelectionsTab`-specific logic.
* Ensure that we are ticking accessibility boxes. The new `TabList` adds arrow-key navigation per https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/
* Fix a longstanding layout complexity to do with tabs jiggling when selected because of bold text. We've put various workarounds in place in the client over the years to try to account for it, but the new `Tab` component takes aim at alleviating it systematically (see documentation on `textContent` for that component).

As always with these new-component PRs, the lion's share of the diff is the pattern-library documentation page component. The components themselves were refreshingly simple to put together: I was able to scaffold out and concentrate on the specific problems I wanted to solve (fixing the jiggle, applying keyboard navigation) and documentation. The toolbox is getting more robust and it's starting to pay dividends!

### Testing

Check out this branch, run `make dev` and visit [the Tabs page in the pattern library](http://localhost:4001/navigation-tab)

image

### DIFF
--- a/src/components/navigation/Tab.tsx
+++ b/src/components/navigation/Tab.tsx
@@ -1,0 +1,88 @@
+import classnames from 'classnames';
+import type { JSX } from 'preact';
+
+import type { IconComponent, PresentationalProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+
+import ButtonBase from '../input/ButtonBase';
+
+type ComponentProps = {
+  icon?: IconComponent;
+  /**
+   * Text string representing the content of the tab when selected. The tab
+   * button will be sized to accommodate this string in bold text. This can
+   * prevent tab jiggle.
+   */
+  textContent?: string;
+  selected?: boolean;
+  variant?: 'basic';
+};
+
+type HTMLAttributes = Omit<
+  JSX.HTMLAttributes<HTMLButtonElement>,
+  'size' | 'icon' | 'title'
+>;
+
+export type TabProps = PresentationalProps & ComponentProps & HTMLAttributes;
+
+/**
+ * Render a button with appropriate ARIA tab affordances
+ */
+const TabNext = function Tab({
+  children,
+  classes,
+  elementRef,
+
+  icon: Icon,
+  textContent,
+  selected = false,
+  variant = 'basic',
+
+  ...htmlAttributes
+}: TabProps) {
+  return (
+    <ButtonBase
+      {...htmlAttributes}
+      classes={classnames(
+        'gap-x-1.5 enabled:hover:text-brand-dark',
+        {
+          'font-bold': selected && variant === 'basic',
+        },
+        classes
+      )}
+      elementRef={downcastRef(elementRef)}
+      aria-selected={selected}
+      role="tab"
+      data-component="Tab"
+    >
+      {Icon && (
+        <Icon
+          className={classnames(
+            // A small padding value here sizes the icon down slightly in relation
+            // to the tab text, which results in nicer proportions.
+            'p-[0.125em] w-em h-em'
+          )}
+        />
+      )}
+      <span
+        data-content={textContent}
+        data-testid="sizing-wrapper"
+        className={classnames({
+          // Set the content of this span's ::before pseudo-element to
+          // `textContent` and make it bold.
+          'before:content-[attr(data-content)] before:font-bold': textContent,
+          // Make the ::before occupy space within the button, but make it
+          // invisible. This ensures that the tab button is wide enough to show
+          // bolded text even if the visible text is not currently bold. See
+          // https://css-tricks.com/bold-on-hover-without-the-layout-shift/
+          'before:block before:invisible before:h-0 before:overflow-hidden':
+            textContent,
+        })}
+      >
+        {children}
+      </span>
+    </ButtonBase>
+  );
+};
+
+export default TabNext;
--- a/src/components/navigation/TabList.tsx
+++ b/src/components/navigation/TabList.tsx
@@ -1,0 +1,58 @@
+import classnames from 'classnames';
+import type { JSX } from 'preact';
+
+import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import type { PresentationalProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+
+type HTMLAttributes = Omit<JSX.HTMLAttributes<HTMLElement>, 'size'>;
+
+type ComponentProps = {
+  /**
+   * By default, TabLists are oriented horizontally. Vertically-oriented
+   * TabLists add up/down arrow-key navigation.
+   */
+  vertical?: boolean;
+};
+
+export type TabListProps = PresentationalProps &
+  ComponentProps &
+  HTMLAttributes;
+
+/**
+ * Render a tablist container for a set of tabs, with arrow key navigation per
+ * https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/
+ */
+const TabListNext = function TabList({
+  children,
+  classes,
+  elementRef,
+
+  vertical = false,
+
+  ...htmlAttributes
+}: TabListProps) {
+  const tabListRef = useSyncedRef(elementRef);
+
+  useArrowKeyNavigation(tabListRef, {
+    selector: 'button',
+    horizontal: true,
+    vertical,
+  });
+
+  return (
+    <div
+      {...htmlAttributes}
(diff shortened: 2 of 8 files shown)


src/components/navigation/Tab.tsx
```diff
@@ -0,0 +1,77 @@
...
+        data-content={textContent}
+        data-testid="sizing-wrapper"
+        className={classnames({
+          // Set the content of this span's ::before pseudo-element to
+          // `textContent` and make it bold.
+          'before:content-[attr(data-content)] before:font-bold': textContent,
+          // Make the ::before occupy space within the button, but make it
+          // invisible. This ensures that the tab button is wide enough to show
+          // bolded text even if the visible text is not currently bold. See
+          // https://css-tricks.com/bold-on-hover-without-the-layout-shift/
```
lyzadanger, line 66: The buttons at the top of this interface itself use this trick so that they don't move around when active:

image

src/components/navigation/TabList.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  ...htmlAttributes
+}: TabListProps) {
+  const tabListRef = useSyncedRef(elementRef);
+
+  useArrowKeyNavigation(tabListRef, {
+    selector: 'button',
+    horizontal: true,
+    vertical: false,
+  });
```
lyzadanger, line 30: This was so awesomely easy to drop in. Hooray!

src/components/navigation/test/Tab-test.js
```diff
@@ -0,0 +1,45 @@
...
+
+import Tab from '../Tab';
+
+const contentFn = (Component, props = {}) => {
+  return mount(
+    <div role="tablist">
+      <Component {...props}>This is child content</Component>
+    </div>
+  );
+};
```
lyzadanger, line 13: Here and in the tests for `TabList`: our automated common and accessibility tests are working well! Accessibility tests will fail if `Tab` isn't inside of a `tablist`, and `TabList` tests will fail if the `TabList` doesn't have any `tab` children. Sweet.

lyzadanger:
I've gone ahead and pinged both @robertknight and @acelaya for review on this one. Rob more specifically for the keyboard navigation and behavior aspects.

src/components/navigation/Tab.tsx
acelaya: Clever! 😃

acelaya (approved):
A couple of minor comments, but looks good 👍🏼

src/components/navigation/TabList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  vertical = false,
+
+  ...htmlAttributes
+}: TabListProps) {
+  const tabListRef = useSyncedRef(elementRef);
+
+  useArrowKeyNavigation(tabListRef, {
+    selector: 'button',
+    horizontal: true,
+    vertical: vertical,
```
robertknight, line 41:
```suggestion
    vertical,
```

src/pattern-library/components/patterns/navigation/TabPage.js
```diff
@@ -0,0 +1,390 @@
...
+              with <code>role={'"tablist"'}</code> (or use the{' '}
+              <code>TabList</code> component).
+            </li>
+            <li>
+              You <em>should</em> provide an{' '}
+              <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls">
+                <code>aria-controls</code>
+              </a>{' '}
+              attribute to each <code>Tab</code>. This is not always feasible in
+              our applications.
```
robertknight, line 85: My initial reaction to "This is not always feasible in our applications" was "why?". Are we doing something unusual or is there an implementation problem? What are the problems caused by not using this?

src/pattern-library/components/patterns/navigation/TabPage.js
```diff
@@ -0,0 +1,390 @@
...
+              <TabList classes="gap-x-6">
+                <Tab
+                  selected={sidebarPanel3 === 'annotations'}
+                  onClick={() => setSidebarPanel3('annotations')}
+                  textContent="Annotations##"
+                >
+                  Annotations
+                  <span className="relative bottom-[3px] left-[2px] text-[10px]">
+                    {52}
+                  </span>
```
robertknight, line 219: Do you have plans to create a pattern library component for counts associated with a tab? I suspect this may be a common-enough pattern and have enough considerations around design/a11y that require thought to make it worthwhile.

src/components/navigation/Tab.tsx
```diff
@@ -0,0 +1,88 @@
...
+import ButtonBase from '../input/ButtonBase';
+
+type ComponentProps = {
+  icon?: IconComponent;
+  /**
+   * Text string representing the content of the tab when selected. The tab
+   * button will be sized to accommodate this string in bold text. This can
+   * prevent tab jiggle.
+   */
+  textContent?: string;
```
robertknight, line 16: I suppose we could try to simplify the API by determining the `textContent` automatically, eg. in a `useLayoutEffect` that reads `textContent` after render, although that would have a limitation of not being able to account for expected text changes (eg. if counts were not available and become available later), and be a bit more expensive.

robertknight (approved):
The implementation fundamentals look good. From a design perspective I think in future we may want to reconsider the lack of any borders between tabs or between tabs and tab panels, as this makes the tab-iness of the control less obvious, at least in certain contexts.

src/pattern-library/components/patterns/navigation/TabPage.js
lyzadanger: Great question! The WAI-ARIA authoring guidelines for a full tabby-widget-y interface assumes  a structure like so:

```
<tablist>
  <tab controls="panel1">One</tab>
  <tab controls="panel2" aria-selected>Two</tab>
  <tab controls="panel3">Three</tab>
</tablist>
<tabpanel id="panel1" style="visibility:hidden">Content for panel One</tabpanel>
<tabpanel id="panel2">Content for panel Two</tabpanel>
<tabpanel id="panel3" style="visibility:hidden">Content for panel Three</tabpanel>
```

To wit:

* A `tabpanel` for each `tab`
* Non-active `tabpanel`s are visually hidden

However, in the sidebar, we have:

```
<tablist>
  <tab>Annotations</tab>
  <tab aria-selected>Page Notes</tab>
  <tab>Orphans</tab>
</tablist>
<div>Page-note annotation threads</div>
```

To wit: All three tabs control the same "panel"; we don't have a discrete element "per panel". I don't think it would appropriate to assign the same `aria-controls` value to each tab (i.e. the `id` of the `div` containing threads, which, for clarity, does not currently have an `id` attribute). But I could be wrong.

I think this sort of situation is why WAI-ARIA guidelines indicate that there _should_ be an `aria-controls` on `role="tab"` elements, but does not indicate that there _must_.

src/pattern-library/components/patterns/navigation/TabPage.js
lyzadanger: We could do that, sure. It would simplify the acrobatics I did in re `textContent`. I'll give it a think.

src/components/navigation/Tab.tsx
lyzadanger: Yeah, I'm going to continue to hem and haw on this some. I don't plan to immediately cut a release of the package, so something intelligent might occur to me in the next few days.

lyzadanger:
> The implementation fundamentals look good. From a design perspective I think in future we may want to reconsider the lack of any borders between tabs or between tabs and tab panels, as this makes the tab-iness of the control less obvious, at least in certain contexts.

For sure. My goal in the "styling" part of this was simply to capture the current pattern as-is.

src/pattern-library/components/patterns/navigation/TabPage.js
robertknight: > To wit: All three tabs control the same "panel"; we don't have a discrete element "per panel".

Ah, OK. That should be an implementation detail that we are re-using the DOM elements though. From a user's perspective (screen reader or not) I think it would ideally "look" the same as if we were using separate panels. I wonder if there is a blessed way of doing this kind of virtualization. How does `aria-controls` actually get used?

PR #822, opened 2023-02-01
Make Panel content scroll if Panel height exceeds that set on parent element

lyzadanger:
This PR updates the `Panel` component such that its rendered `children` will scroll, if needed, to not exceed a height constraint set on the parent element of the `Panel`.

We have some use cases in which we need to constrain the height of content, especially in modals. The goal is to have a modal not exceed a particular height — typically some proportion of the viewport height — and have some of the modal's content automatically scroll as needed. This notoriously fiddly little problem is made more challenging by the desire not to scroll the title and buttons/action of a modal, just the interior content.

We solved this previously for our legacy `Dialog` and `Modal` components and it needs to be re-solved to work with our newer components and Tailwind in general.

Both legacy `Dialog` and `Modal` include `Panel`s[^1]. The least invasive solution I've found to solve this layout problem today is to make `Panel` quietly handle scrolling-as-needed without user intervention by adding a few CSS classes to its outer element and throwing a `Scroll` component around its rendered `children`.

The pattern goes something like this (using Tailwind utility classes for quick shorthand):

```html
<div className="h-[250px]"> <!-- outer element that sets a height constraint, perhaps a modal -->
  <div className="flex flex-col min-h-0 h-full"> <!-- Panel container -->
  <div>My Panel Title</div> <!-- this won't scroll -->
  <div className="overflow-auto min-h-0 h-full">Panel children (content)</div> <!-- SCROLLS -->
  <div><button>OK</button></div> <!-- This won't scroll either -->
</div>
```

The pieces are play are:

* An outer element setting some sort of height constraint
* An immediate child of the constrained-height element establishes a columnar flex layout -- This is the `Panel`'s outer component in our applied case
* One immediate child of the columnar-flex element has `overflow` set - this is the element containing panel `children` content — and will scroll if needed

If the combined heights of the immediate children of the columnar-flex layout exceed the height constraint, the `overflow` div will automatically scroll and the entire `Panel` height will not exceed the height constraint.

This is probably easiest to demonstrate by example, and I have added such an example to the [Panel page](http://localhost:4001/layout-panel) in the pattern library. 

These changes have no visible effect on `Panel`s not rendered in height-constrained contexts.

I will be making immediate use of this to get `Dialog` and `Modal` migrated over ASAP.

Oh, `Panel` is also converted to TSX here.

## Testing

* Make sure the added example `Panel` in the pattern library scrolls
* Take a peek at the other `Panel` examples and see that they look as expected (they should not have changed)

The real proof will be in subsequent PR(s) for `Modal` and `Dialog`.

[^1]: When we redesign these components, I'd like to evaluate decoupling them from `Panel` and letting consumers set whatever `children` they'd like, but that's for later.


rollup-tests.config.js
```diff
@@ -70,6 +70,8 @@ export default {
               '**/test/**/*.js',
               '**/test-util/**',
               'src/components/icons/**/*.js',
+              'src/pattern-library/**/*.js',
+              'src/pattern-library/**/*.ts*',
```
lyzadanger, line 74: The test for this change makes use of a little utility component, `LoremIpsum`. Without updating this config, istanbul (and, later, CodeCov) will bleat because the module that `LoremIpsum` is in doesn't have tests.

src/components/layout/Panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+import type { IconComponent, CompositeProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+
+import Card from './Card';
+import CardContent from './CardContent';
+import CardHeader from './CardHeader';
+import CardTitle from './CardTitle';
+import CardActions from './CardActions';
+
+import Scroll from '../data/Scroll';
```
lyzadanger, line 12: Oops this import is out of order. I'll fix along with other feedback on this PR.

src/components/layout/Panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+  icon: Icon,
+  onClose,
+  title,
+
+  ...htmlAttributes
+}: PanelProps) {
+  return (
+    <Card
+      {...htmlAttributes}
+      classes={'flex flex-col min-h-0 h-full'}
```
acelaya, line 53: Just out of curiosity. Any reason to do `classes={'flex flex-col min-h-0 h-full'}` instead of `classes="flex flex-col min-h-0 h-full"`?

acelaya (approved):
Very clever approach! I remember years ago trying to achieve this using bootstrap modals.

src/components/layout/Panel.tsx
acelaya: Maybe we should consider adding a linting rule that helps us keep the imports the way we want? Being able to fix this kind of things with `make format` would be very convenient.

PR #824, opened 2023-02-02
Extract `useUniqueId` hook for reuse

lyzadanger:
This PR extracts a simple existing hook `useUniqueId` for reuse. This was an internal hook in the legacy `Dialog` component. I've re-used it and added some basic tests.

This is a separate PR to keep the forthcoming `Modal` PR from being too ginormous.

### DIFF
--- a/src/hooks/use-unique-id.ts
+++ b/src/hooks/use-unique-id.ts
@@ -1,0 +1,16 @@
+import { useState } from 'preact/hooks';
+
+let idCounter = 0;
+/**
+ * Return an ID string beginning with `prefix` that is unique per component
+ * instance.
+ *
+ * This avoids different instances of a component re-using the same ID.
+ */
+export function useUniqueId(prefix: string) {
+  const [id] = useState(() => {
+    ++idCounter;
+    return `${prefix}-${idCounter}`;
+  });
+  return id;
+}
--- a/src/hooks/test/use-unique-id-test.js
+++ b/src/hooks/test/use-unique-id-test.js
@@ -1,0 +1,67 @@
+import { mount } from 'enzyme';
+
+import { useUniqueId } from '../use-unique-id';
+
+function Widget() {
+  const varietyId = useUniqueId('banana-variety');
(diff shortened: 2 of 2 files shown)


src/hooks/use-unique-id.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * This avoids different instances of a component re-using the same ID.
+ */
+export function useUniqueId(prefix: string) {
+  const [id] = useState(() => {
+    ++idCounter;
+    return `${prefix}-${idCounter}`;
+  });
+  return id;
+}
```
acelaya, line 16: I understand we use this hook for the `id` attribute and to be able to link inputs with labels via `htmlFor`.

Why are we not using the standard `useId` hook from react/preact? See https://preactjs.com/guide/v10/hooks/#utility-hooks

acelaya (approved):
Looks good to me.

I just added a question as I don't see a clear reason not to use the standard `useId` hook instead of our own implementation, but I guess there's a reason, so approved from me.

src/hooks/use-unique-id.ts
robertknight: The main reason is that `useId` did not exist at the time this hook was originally created. A remaining advantage it has is that it allows giving IDs human-readable prefixes which is useful when scanning DOM output. An upside that `useId` has is that it promises to produce the same ID sequence for client and server rendering, in case we ever do server-side rendering.

We could actually combine them to get the best of both:

```js
function useUniqueId(prefix: string) {
  return prefix + '-' + useId();
}
```

lyzadanger: 🤦🏻‍♀️ While I was working on this, I did have that tickle in the back of my head about `useId` but then ploughed dopily onward. I'm going to land this as-is because it's not actually broken or anything and it's in a chain of dependencies. I'm writing down a follow-up item for it _right now_.

acelaya: Thanks for the explanation 😃

PR #825, opened 2023-02-02
Allow pattern-library wrapper components to masquerade as the component they wrap

lyzadanger:
In the pattern library, it's often convenient to encapsulate some state or other logic in a reusable local example component. These example components tend to directly wrap the component being documented.

This PR extends the `jsx-to-string` module to allow components following the naming pattern `<ComponentWrapper>` to masquerade `<Component>` . It also allows certain props to be hidden similarly.

_Note_: I'd like to concede that this module needs attention: we've been glomming onto it for a while. I converted it to TSX here, but it could use a refactor. I considered doing that as part of this work but there's a lot else going on so I'm going to hold.

### Explanation

This change is better explained with an example. In the forthcoming PR to add a new `Modal` component, the associated pattern library has a local wrapper that takes props like so, forwarding most on to the `Modal` component it wraps:

```tsx
type ModalWrapperProps = ModalProps & {
  /** Pattern-wrapping prop. Not visible in source view */
  nonCloseableWrapperProp?: boolean;
};

function ModalWrapper({
  buttons,
  nonCloseableWrapperProp,
  children,
  ...modalProps
}: ModalWrapperProps) {
  // Renders an example, forwarding various props on to `Modal`
}
```

Then, later...

```tsx
            <Library.Demo title="Non-closeable modal" withSource>
              <ModalWrapper
                title="Non-closeable modal"
                nonCloseableWrapperProp={true}
              >
                <p>
                  This is a non-closeable modal. There is no close button at the
                  top of the panel, and the modal cannot be dismissed by
                  pressing escape or clicking outside of it. However, an escape
                  hatch has been provided for you below (typically non-closeable
                  modals do not have any buttons).
                </p>
              </ModalWrapper>
            </Library.Demo>
```

When this demo is rendered and the source panel activated:

image

This allows the pattern library to "fake" source:

* It appears that `Modal` was used directly, and
* The internal `nonCloseableWrapperProp` is hidden

### DIFF
--- a/src/pattern-library/util/jsx-to-string.tsx
+++ b/src/pattern-library/util/jsx-to-string.tsx
@@ -3,20 +3,26 @@ import hljsXMLLang from 'highlight.js/lib/languages/xml';
 import hljsJavascriptLang from 'highlight.js/lib/languages/javascript';
 import { Fragment } from 'preact';
 
+import type { ComponentChildren, VNode } from 'preact';
+
 /**
  * Escape `str` for use in a "-quoted string.
- *
- * @param {string} str
  */
-function escapeQuotes(str) {
+function escapeQuotes(str: string) {
   return str.replace(/"/g, '\\"');
 }
 
-/** @param {any} type */
-function componentName(type) {
+/**
+ * Format a component's name for display
+ */
+function componentName(type: any) {
   let name = typeof type === 'string' ? type : type.displayName ?? type.name;
 
-  name = name.replace('Wrapper', '');
+  // Remove the string "Wrapper" from component names. This allows the pattern
+  // library to create a convenience wrapper component around a given component
+  // being documented, and have it appear that the documented component is being
+  // used directly in rendered source content.
+  name = name.replace(/\BWrapper$/, '');
   // Handle (display)name conflicts if there are two components with the same
   // name. e.g. if there are two components named `Foo`, the second of those
   // encountered will have a name of `Foo$1`. Strip the `$1` in this case.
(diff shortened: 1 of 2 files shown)


src/pattern-library/util/jsx-to-string.tsx
```diff
@@ -3,18 +3,26 @@ import hljsXMLLang from 'highlight.js/lib/languages/xml';
...
+ * Format a component's name for display
+ */
+function componentName(type: any) {
+  let name = typeof type === 'string' ? type : type.displayName ?? type.name;
+
+  // Remove the string "Wrapper" from component names. This allows the pattern
+  // library to create a convenience wrapper component around a given component
+  // being documented, and have it appear that the documented component is being
+  // used directly in rendered source content.
+  name = name.replace('Wrapper', '');
```
robertknight, line 25:
```suggestion
  name = name.replace(/\BWrapper$/, '');
```

Matches "Wrapper" only at the end of the component name, and not if the component name is exactly "Wrapper".

src/pattern-library/util/jsx-to-string.tsx
```diff
@@ -71,6 +70,11 @@ export function jsxToString(vnode) {
     // (eg. from an index or item ID) so it doesn't make sense to include it either.
     let propStr = Object.entries(vnode.props)
       .map(([name, value]) => {
+        // Don't render props that are for internal pattern-library wrapper
+        // component use only
+        if (name.includes('WrapperProp')) {
```
robertknight, line 75: We could also use a prefix (eg. `$` or `_`) indicate internal props as a less verbose alternative.

robertknight (approved):

src/pattern-library/util/jsx-to-string.tsx
lyzadanger: Yeah, I may adjust this when I clean this module up. Thanks!

PR #826, opened 2023-02-02
Add updated `Modal` component and documentation

lyzadanger:
Depends on #825

Don't panic: More than half the diff is the pattern-library documentation.

This PR adds a `Modal` component that conforms to the current API and styling conventions for shared components. It combines the functionality of legacy `Dialog` and `Modal` components[^1].

image

This implementation should be considered a bridge implementation until we have a little time to 
focus on a full-featured `dialog` component.

It retains from `Dialog` and `Modal`, unchanged:

* initial focus-routing behavior
* `aria-describedby` behavior
* `useElementShouldClose` handling
* `Modal`s use a `Panel` layout still

What it does add is, I hope, more sane modal sizing and positioning configuration. There are some size defaults, and a `width` prop, and some styling guard rails to ensure that modal dialogs never get too big or stupid for their viewports. I've combined sizing and positioning use cases from LMS, the client and the previous implementations of this component. The API is simpler than previously.

The [pattern-library page](http://localhost:4001/feedback-modal) for this new component documents and exemplifies its aims. I suggest starting there...

[^1]: `Dialog` is not actually used anywhere in our applications directly. Legacy `Modal` wraps `Dialog`.

### DIFF
--- a/src/components/feedback/Modal.tsx
+++ b/src/components/feedback/Modal.tsx
@@ -22,10 +22,10 @@ type ComponentProps = {
   width?: 'sm' | 'md' | 'lg' | 'custom';
   /**
    * Element that should take focus when the Dialog is first rendered. When not
-   * provided, the dialog's outer element will take focus. Setting this prop to
-   * `null` will opt out of focus routing.
+   * provided ("auto"), the dialog's outer element will take focus. Setting this
+   * prop to "manual" will opt out of focus routing.
    */
-  initialFocus?: RefObject<HTMLElement | null> | null;
+  initialFocus?: RefObject<HTMLOrSVGElement | null> | 'auto' | 'manual';
 };
 
 export type ModalProps = PresentationalProps & ComponentProps & PanelProps;
@@ -41,7 +41,7 @@ const ModalNext = function Modal({
 
   classes,
   elementRef,
-  initialFocus,
+  initialFocus = 'auto',
 
   // Forwarded to Panel
   buttons,
@@ -58,18 +58,28 @@ const ModalNext = function Modal({
   const dialogDescriptionId = useUniqueId('dialog-description');
 
   useEffect(() => {
-    // Route focus unless `initialFocus` is set to `null`
-    if (initialFocus !== null) {
-      const focusEl = initialFocus?.current as HTMLInputElement;
-      if (focusEl && !focusEl.disabled) {
-        focusEl.focus();
-      } else {
-        // The `initialFocus` prop has not been set, so use automatic focus
-        // handling. Modern accessibility guidance is to focus the dialog itself
-        // rather than trying to be smart about focusing a particular control
-        // within the dialog.
-        modalRef.current?.focus();
-      }
+    if (initialFocus === 'manual') {
+      return;
+    }
+    if (initialFocus === 'auto') {
+      // An explicit `initialFocus` has not be set, so use automatic focus
+      // handling. Modern accessibility guidance is to focus the dialog itself
+      // rather than trying to be smart about focusing a particular control
+      // within the dialog.
+      modalRef.current?.focus();
+      return;
+    }
+
+    const focusEl = initialFocus?.current as HTMLElement & {
+      disabled?: boolean;
+    };
+
+    if (focusEl && !focusEl.disabled) {
+      focusEl.focus();
+    } else {
+      // Fall back to focusing the modal itself
+      modalRef.current?.focus();
+      return;
     }
 
     // We only want to run this effect once when the dialog is mounted.
--- a/src/components/feedback/Modal.tsx
+++ b/src/components/feedback/Modal.tsx
@@ -1,0 +1,136 @@
+import classnames from 'classnames';
+import type { RefObject } from 'preact';
+import { useEffect, useLayoutEffect } from 'preact/hooks';
+
+import { useElementShouldClose } from '../../hooks/use-element-should-close';
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import { useUniqueId } from '../../hooks/use-unique-id';
+
+import type { PresentationalProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+
+import Overlay from '../layout/Overlay';
+import Panel from '../layout/Panel';
+import type { PanelProps } from '../layout/Panel';
+
+type ComponentProps = {
+  /**
+   * Modal _should_ be provided with a close handler. We have a few edge use
+   * cases, however, in which we need to render a "non-closeable" modal.
+   */
+  onClose?: () => void;
+  width?: 'sm' | 'md' | 'lg' | 'custom';
+  /**
+   * Element that should take focus when the Dialog is first rendered. When not
+   * provided, the dialog's outer element will take focus. Setting this prop to
+   * `null` will opt out of focus routing.
+   */
+  initialFocus?: RefObject<HTMLElement | null> | null;
+};
+
+export type ModalProps = PresentationalProps & ComponentProps & PanelProps;
+
+const noop = () => {};
+
+/**
+ * Show a modal
+ */
+const ModalNext = function Modal({
+  children,
+  width = 'md',
+
+  classes,
+  elementRef,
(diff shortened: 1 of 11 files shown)


src/components/feedback/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  ...htmlAttributes
+}: ModalProps) {
+  const modalRef = useSyncedRef(elementRef);
+  const closeHandler = onClose ?? noop;
+  useElementShouldClose(modalRef, true, closeHandler);
+
+  const dialogDescriptionId = useUniqueId('dialog-description');
+
+  useEffect(() => {
```
lyzadanger, line 60: This `useEffect` is copied from previous implementation of `Dialog`.

src/components/feedback/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+        // within the dialog.
+        modalRef.current?.focus();
+      }
+    }
+
+    // We only want to run this effect once when the dialog is mounted.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useLayoutEffect(
```
lyzadanger, line 79: This `useLayoutEffect` is copied from a previous implementation of `Dialog`

src/components/feedback/test/Modal-test.js
```diff
@@ -0,0 +1,136 @@
...
+  );
+};
+
+describe('Modal', () => {
+  testPresentationalComponent(Modal, {
+    createContent: createComponent,
+    elementSelector: 'div[data-component="Modal"]',
+  });
+
+  describe('initial focus', () => {
```
lyzadanger, line 22: This `describe` block is essentially copied from pre-existing `Dialog` tests

src/components/feedback/test/Modal-test.js
```diff
@@ -0,0 +1,136 @@
...
+      );
+
+      assert.equal(
+        document.activeElement,
+        wrapper.find('[role="dialog"]').getDOMNode()
+      );
+    });
+  });
+
+  describe('closing the modal', () => {
```
lyzadanger, line 90: These tests were copied, too.

src/components/feedback/test/Modal-test.js
```diff
@@ -0,0 +1,136 @@
...
+
+      const event = new Event('keydown');
+      event.key = 'Escape';
+      document.body.dispatchEvent(event);
+      assert.called(onClose);
+      container.remove();
+    });
+  });
+
+  describe('aria-describedby', () => {
```
lyzadanger, line 112: I wrapped these tests in a `describe` block, but they already existed on the previous `Dialog` implementation.

src/tailwind.preset.js
```diff
@@ -84,6 +84,7 @@ export default /** @type {Partial<import('tailwindcss').Config>} */ ({
         DEFAULT: '2px',
       },
       screens: {
+        tall: { raw: '(min-height: 32rem)' },
```
lyzadanger, line 87: This "screen" (media query) is used by the new `Modal`

src/components/feedback/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+   * cases, however, in which we need to render a "non-closeable" modal.
+   */
+  onClose?: () => void;
+  width?: 'sm' | 'md' | 'lg' | 'custom';
+  /**
+   * Element that should take focus when the Dialog is first rendered. When not
+   * provided, the dialog's outer element will take focus. Setting this prop to
+   * `null` will opt out of focus routing.
+   */
+  initialFocus?: RefObject<HTMLElement | null> | null;
```
acelaya, line 28: I see this prop handles three possible "states":

* A ref is provided, in which case the referenced component gets focused.
* The prop is not provided at all, in which case the modal itself gets focused.
* The `null` value is provided, in which case we opt out from focusing capabilities.

I think the last two can be easily confused in consumer code. If I see `<Modal initialFocus={null} ... />` I will not remember that means the focusing is disabled.

I don't see an easy solution though. Maybe accepting a constant value instead, which is more clear in consumer code? Something in the lines of `<Modal initialFocus="DISABLE_FOCUS" ... />`. It's a bit hacky, but maybe more descriptive.

Other ideas?

src/components/feedback/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+      if (focusEl && !focusEl.disabled) {
+        focusEl.focus();
+      } else {
+        // The `initialFocus` prop has not been set, so use automatic focus
+        // handling. Modern accessibility guidance is to focus the dialog itself
+        // rather than trying to be smart about focusing a particular control
+        // within the dialog.
+        modalRef.current?.focus();
+      }
+    }
```
acelaya, line 73: Can I suggest not nesting `if`s? I find it easier to use early returns so that you can mentally discard conditions as soon as possible, as you "follow" the code. When nesting, you need to "drag" all of them.

```ts
if (initialFocus === null) {
  return;
}

const focusEl = initialFocus?.current as HTMLInputElement;
if (focusEl && !focusEl.disabled) {
    focusEl.focus();
} else {
  // ...
  modalRef.current?.focus();
}
```

acelaya (changes requested):
Looks good in general.

I just added two minor comments, but they are very much personal preference. Feel free to answer to those and I'm open to approve if the general feeling is that they make no sense.

robertknight:
The combination of single line separator that doesn't cover the full width of the content, and top-shadow for scrolled content, which does, can look a bit odd:

Top shadow and underline

src/components/feedback/Modal.tsx
robertknight: > The prop is not provided at all, in which case the modal itself gets focused.
> The null value is provided, in which case we opt out from focusing capabilities.

What uses do we have for these two separate cases? Could we just always use one of the two behaviors here?

src/components/feedback/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const modalRef = useSyncedRef(elementRef);
+  const closeHandler = onClose ?? noop;
+  useElementShouldClose(modalRef, true, closeHandler);
+
+  const dialogDescriptionId = useUniqueId('dialog-description');
+
+  useEffect(() => {
+    // Route focus unless `initialFocus` is set to `null`
+    if (initialFocus !== null) {
+      const focusEl = initialFocus?.current as HTMLInputElement;
```
robertknight, line 63: Hmm... the assumption that `focusEl` is an `HTMLInputElement` might not always be valid. For example it might be a textarea or button. I presume the reason we are doing this is that not all HTML elements have a `disabled` prop. We could make this clearer with something like:

```js
      const focusEl = initialFocus?.current as HTMLElement & { disabled?: boolean };
      if (focusEl && !focusEl.disabled) {
```

robertknight (commented):
This generally looks good. I think the most important thing to resolve once we get this landed will be to properly implement focus trapping etc, so that our dialogs are following correct behavior for keyboard navigation.

src/components/feedback/Modal.tsx
lyzadanger: Good feedback, thanks!

src/components/feedback/Modal.tsx
lyzadanger: Sure! I kind of blindly cut-and-pasted this from a previous implementation so it's good to get fresh eyes on it. I agree. Guards are easier to reason about.

src/components/feedback/Modal.tsx
lyzadanger: > The combination of single line separator that doesn't cover the full width of the content, and top-shadow for scrolled content, which does, can look a bit odd:
> 
> Top shadow and underline

I tend to agree. However, this styling is actually owned by `Panel` so I want to hit this in a follow-up. Making a note.

lyzadanger:
I've pushed a commit that adjusts `initialFocus` handling per feedback.

acelaya (approved):
Looks good!

PR #827, opened 2023-02-03
Extract `InputRoot` component

lyzadanger:
This PR establishes a root component for textual input elements. The purpose of this is to be able to share design patterns and behavior across those input elements. I'm terming this a "Root" element, as distinct from a "Base" element, because it is private to the package. Early days here; I want to feel this out.

Expect to see a `Select` component based on this imminently. A `TextArea` component is probably not far off, either, but doesn't have as immediate of a need as `Select`.

There should be no visible or behavioral changes to the existing `Input` component. You can verify that on the [Input pattern-library page](http://localhost:4001/input-input).

### DIFF
--- a/src/components/input/InputRoot.tsx
+++ b/src/components/input/InputRoot.tsx
@@ -1,0 +1,62 @@
+import classnames from 'classnames';
+import type { JSX } from 'preact';
+
+import type { PresentationalProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+
+import { inputGroupStyles } from './InputGroup';
+
+type RootComponentProps = {
+  element?: 'input' | 'select';
+  hasError?: boolean;
+};
+
+export type InputRootProps = PresentationalProps &
+  RootComponentProps &
+  (
+    | Omit<JSX.HTMLAttributes<HTMLInputElement>, 'size' | 'icon'>
+    | Omit<JSX.HTMLAttributes<HTMLSelectElement>, 'size' | 'icon'>
+  );
+
+/**
+ * Root component for various input types that applies a consistent design
+ * pattern.
+ */
+const InputRootNext = function InputRoot({
+  element: Element = 'input',
+  children,
+  classes,
+  elementRef,
+  hasError,
+
+  ...htmlAttributes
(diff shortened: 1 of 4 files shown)


src/components/input/InputRoot.tsx
```diff
@@ -1,43 +1,48 @@
...
     );
   }
+
   return (
-    <input
+    <Element
+      data-component="InputRoot"
       {...htmlAttributes}
-      type={type}
+      // @ts-ignore-next
```
lyzadanger, line 45: Ref-typing purgatory...I should add more comment here but I don't really know what to say.

src/components/input/InputRoot.tsx
```diff
@@ -1,43 +1,48 @@
...
   if (!htmlAttributes.id && !htmlAttributes['aria-label']) {
     console.warn(
       '`Input` component should have either an `id` or an `aria-label` attribute'
     );
   }
+
   return (
-    <input
+    <Element
+      data-component="InputRoot"
```
lyzadanger, line 43: In all normal uses of this component this `data-component` attribute should be overridden. That's by design.

src/components/input/test/Input-test.js
```diff
@@ -10,17 +10,4 @@ const contentFn = (Component, props = {}) => {
 
 describe('Input', () => {
   testPresentationalComponent(Input, { createContent: contentFn });
-
-  it('should warn in console if accessibility attributes are missing', () => {
```
lyzadanger, line 14: This is now the responsibility of `InputRoot`

src/components/input/InputRoot.tsx
acelaya: Just for my own understanding (not related with this PR per se). What's the purpose of `data-component` when used by consumer code?

acelaya (approved):
LGTM

src/components/input/InputRoot.tsx
lyzadanger: Mostly as a debugging affordance. It's easier to see when inspecting the DOM where elements come from.

PR #828, opened 2023-02-03
Add new `Select` component

lyzadanger:
Depends on #827 

This PR adds a basic `Select` component, building on `InputRoot`. Its purpose right now is to capture, standardize and simplify the design pattern used in the grading toolbar in LMS:

image

This is to support some UI improvements in that area.

The component is simple and young. You can see its documentation on this branch at http://localhost:4001/input-select

image

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -1,0 +1,55 @@
+import classnames from 'classnames';
+import type { ComponentChildren, JSX } from 'preact';
+
+import type { PresentationalProps } from '../../types';
+
+import InputRoot from './InputRoot';
+
+type ComponentProps = {
+  children?: ComponentChildren;
+  hasError?: boolean;
+};
+export type SelectProps = PresentationalProps &
+  ComponentProps &
+  JSX.HTMLAttributes<HTMLSelectElement>;
+
+// URI-encoded source of `CaretDownIcon`
+// Currently, the color (stroke) is hard-coded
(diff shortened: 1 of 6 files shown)


src/components/input/Select.tsx
```diff
@@ -0,0 +1,55 @@
...
+  children?: ComponentChildren;
+  hasError?: boolean;
+};
+export type SelectProps = PresentationalProps &
+  ComponentProps &
+  JSX.HTMLAttributes<HTMLSelectElement>;
+
+// URI-encoded source of `CaretDownIcon`
+// Currently, the color (stroke) is hard-coded
+const arrowImage = `url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16' viewBox='0 0 16 16' aria-hidden='true' focusable='false'%3E%3Cg fill-rule='evenodd'%3E%3Crect fill='none' stroke='none' x='0' y='0' width='16' height='16'%3E%3C/rect%3E%3Cpath fill='none' stroke='%239c9c9c' stroke-linecap='round' stroke-linejoin='round' stroke-width='2' d='M12 6l-4 4-4-4'%3E%3C/path%3E%3C/g%3E%3C/svg%3E")`;
```
lyzadanger, line 18: Alternately, we can extend the Tailwind theme in the package's preset:

```
theme: {
  extend: {
    backgroundImage: {
      'caret-down': 'url("/* ... *./")',
    }
  }
}
```

which could then be applied with the `bg-caret-down` utility class.

Putting the source of the image here in the component module means we could ostensibly set `stroke` color or other attribute values dynamically. The up side of putting it in the tailwind preset is that, ostensibly, a consumer could override it with something completely different.

src/components/input/Select.tsx
```diff
@@ -0,0 +1,55 @@
...
+        // padding (3 units) = 7 units of right padding needed.
+        'bg-no-repeat bg-[center_right_0.5rem] pr-7',
+        classes
+      )}
+      element="select"
+      type={type}
+      hasError={hasError}
+      style={{
+        backgroundImage: arrowImage,
+      }}
```
lyzadanger, line 46: The LMS implementation of this `<select>` uses a separate Icon component/element and absolutely-positions it at the right edge of a wrapping container element.

The advantage of using a background image is that no wrapping container is needed. This allows `<Select>` to take advantage of `InputGroup` styling, e.g.

src/components/input/Select.tsx
acelaya: I guess there are pros and cons to both approaches, but since it's hard to see how the needs for this component will evolve, I would suggest keeping it like this and change it later if wee see the need.

acelaya (approved):

PR #841, opened 2023-02-09
Remove explicit height from `Select` component

lyzadanger:
Setting a fixed height here is too restrictive for a presentational component. Applications should be able to style the height of this component as desired.

See related PR here https://github.com/hypothesis/lms/pull/5008

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -31,7 +31,7 @@ const SelectNext = function Select({
   return (
     <InputRoot
       classes={classnames(
-        'appearance-none h-touch-minimum',
+        'appearance-none',
         // position the down-arrow image centered at the right, offset from the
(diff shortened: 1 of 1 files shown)


robertknight (approved):
Is there still a sensible default that consumers can get without customizing this?

lyzadanger:
> Is there still a sensible default that consumers can get without customizing this?

The sensible default is implicit: it will size to local font context, provide internal scaling/padding consistent with other UI components, and there is also padding affordance in this case to ensure that options text never bump into the down-caret icon. `Input` does not set an explicit height either; this is consistent with that. 

Some of this flexibility is because this is an "early" component: the design pattern isn't fully baked, so I don't want to "pre-guess" what applications will need where it's not necessary.

PR #843, opened 2023-02-10
Add default `type="button"` to all button components

lyzadanger:
This simple PR ensures that buttons default to a `type` of `button`, which should untangle the frustrating situation in https://github.com/hypothesis/lms/pull/5017 (button elements with `type="button"` will not submit forms).

This change is on `ButtonBase`, so applies to all button components.


robertknight (approved):
Seems sensible to me. I think this is quite a common thing for other UI frameworks to do?

lyzadanger:
> Seems sensible to me. I think this is quite a common thing for other UI component frameworks to do?

At least some of them do. Material UI and Chakra do; Reactstrap doesn't.

Issue #839, opened 2023-02-09
Address visual balance of border above scrolling content in Panel


lyzadanger:
A `Panel`'s content will scroll if it overflows. The scrolling content is full width, but the bottom border of the header above the content does not extend full width. This looks a little funky.

<img width="765" alt="image" src="https://user-images.githubusercontent.com/439947/217850512-a93ab941-800a-4625-9e69-f146a627600c.png">

PR #852, opened 2023-02-13
Provide more flexibility with headers and padding of Panels

lyzadanger:
This PR provides a set of additional props for `CardHeader`, `Panel` and `Modal` to give more flexibility in layout styling.

These changes will cause a visual change to existing Modals: the border under the modal's header will now extend all the way to the edges of the modal. Existing `Panel`s are unaffected. Other props introduced here are additive (no changes to existing usage of these components).

The first problem this work attempts to solve is to make `Panel`/`Modal` headers span the full width if there is scrollable content. This is to avoid scroll hint shadows looking funky (see https://github.com/hypothesis/frontend-shared/issues/839).

I also ran into a situation recently in which it was necessary to have more control over the way that padding is applied to Panel content.

The broader fix here, so we don't end up in an arms race, is to not force all `Modal` content to be in `Panel`s (i.e. allow Modals with arbitrary content, not just Panels). I intend to make the next-generation `Dialog` component more flexible in that respect.

These changes are documented on the pattern library pages for `Card` `Panel`, and `Modal`.

Fixes https://github.com/hypothesis/frontend-shared/issues/839

### DIFF
--- a/src/components/layout/Panel.tsx
+++ b/src/components/layout/Panel.tsx
@@ -13,6 +13,19 @@ import CardActions from './CardActions';
 type ComponentProps = {
   /** Buttons are rendered at the bottom right of the Panel. */
   buttons?: ComponentChildren;
+
+  /**
+   * Make the header take the full width of the Panel, with a full-width border
+   */
+  fullWidthHeader?: boolean;
+
+  /**
+   * Forwarded to `CardContent`. If `none`, content is not wrapped with
+   * `CardContent`. This allows content to span the full width and height of
+   * the Panel's content area without any inset/padding.
+   */
+  paddingSize?: 'sm' | 'md' | 'lg' | 'none';
+
   /** Optional icon to render in the header */
   icon?: IconComponent;
 
@@ -21,6 +34,13 @@ type ComponentProps = {
    * this function as an onClick handler
    */
   onClose?: () => void;
+  /**
+   * The content of the Panel should scroll if it exceeds available vertical
+   * space. This will cause the `CardHeader` (and its bottom border) to span the
+   * full width of the Panel. This prevents scroll shadow hints from looking
+   * awkward.
+   */
+  scrollable?: boolean;
   title: string;
 };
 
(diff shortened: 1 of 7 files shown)


lyzadanger:
I can't make any sense of the CodeCov fail. Local test coverage is complete. I'll look in the morning with fresh eyes.

src/components/layout/Panel.tsx
```diff
@@ -21,6 +34,13 @@ type ComponentProps = {
    * this function as an onClick handler
    */
   onClose?: () => void;
+  /**
```
acelaya, line 37: Should there be an empty line before this comment?

acelaya (approved):
I think the code coverage issue is legit. It is complaining about [this line](https://github.com/hypothesis/frontend-shared/pull/852/files#diff-c69bca562089f7ded78b4724ef4528cee0a4fa8d5e4b8d27c9beb98f0a276503R74) not being covered.

In order to cover that you need to explicitly provide `paddingSize="none"` in some test and check that there's no `CardContent` in that case.

lyzadanger:
> I think the code coverage issue is legit. It is complaining about [this line](https://github.com/hypothesis/frontend-shared/pull/852/files#diff-c69bca562089f7ded78b4724ef4528cee0a4fa8d5e4b8d27c9beb98f0a276503R74) not being covered.
> 
> In order to cover that you need to explicitly provide `paddingSize="none"` in some test and check that there's no `CardContent` in that case.

Oh, yes, it's legit! I got so busy yesterday that I didn't get a chance to circle back to this.

acelaya:
> > I think the code coverage issue is legit. It is complaining about [this line](https://github.com/hypothesis/frontend-shared/pull/852/files#diff-c69bca562089f7ded78b4724ef4528cee0a4fa8d5e4b8d27c9beb98f0a276503R74) not being covered.
> > In order to cover that you need to explicitly provide `paddingSize="none"` in some test and check that there's no `CardContent` in that case.
> 
> Oh, yes, it's legit! I got so busy yesterday that I didn't get a chance to circle back to this.

The weird thing is that locally it does not complain about that. Only if you open the code coverage report you can see that branch is not covered, but in terms of lines, it is reported as covered, which I think it's not correct.

lyzadanger:
> The weird thing is that locally it does not complain about that. Only if you open the code coverage report you can see that branch is not covered, but in terms of lines, it is reported as covered, which I think it's not correct.

Yes, that is my general complaint as well. I did check local coverage before pushing the branch.

PR #855, opened 2023-02-15
Add flexibility for scrolling partial content in `Panel` and `Modal`

lyzadanger:
We've got some complex modal-contained layouts in our LMS app. It's been a bit of a beast, but I hope I've covered most if not all of the scrolling bases at this point.

This PR allows `scrollable` to be disabled on `Modal`s because sometimes we don't want to scroll _all_ content in a modal, but only _some_ of it. Modals still are scrollable by default.

Some use-case examples have been added to the [Panel](http://localhost:4001/layout-panel) and [Modal](http://localhost:4001/feedback-modal) pattern-library pages, which I hope demonstrate the point of these changes, because they're tedious to explain :).

This stuff is tricky 😅

### DIFF
--- a/src/components/feedback/Modal.tsx
+++ b/src/components/feedback/Modal.tsx
@@ -28,7 +28,11 @@ type ComponentProps = {
   initialFocus?: RefObject<HTMLOrSVGElement | null> | 'auto' | 'manual';
 };
 
-export type ModalProps = PresentationalProps & ComponentProps & PanelProps;
+// This component forwards a number of props on to `Panel` but always sets the
+// `fullWidthHeader` prop to `true`.
+export type ModalProps = PresentationalProps &
+  ComponentProps &
+  Omit<PanelProps, 'fullWidthHeader'>;
 
 const noop = () => {};
 
@@ -48,6 +52,7 @@ const ModalNext = function Modal({
   icon,
   onClose,
   paddingSize = 'md',
+  scrollable = true,
   title,
 
   ...htmlAttributes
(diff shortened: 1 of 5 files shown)


src/components/feedback/Modal.tsx
```diff
@@ -138,11 +143,12 @@ const ModalNext = function Modal({
       >
         <Panel
           buttons={buttons}
+          fullWidthHeader={true}
```
lyzadanger, line 146: Previously, all Modals were `scrollable`, and `Panel` will set a full-width header on when `scrollable`. To avoid confusion or inconsistency, all Modals will continue to have a full-width header whether or not they are `scrollable`.

src/components/layout/test/Panel-test.js
```diff
@@ -92,7 +92,7 @@ describe('Panel', () => {
       );
 
       assert.isAbove(
-        wrapper.find('div[data-component="CardContent"]').first().getDOMNode()
+        wrapper.find('div[data-component="CardContent"] p').first().getDOMNode()
```
lyzadanger, line 95: Because the `CardContent` in play here now has height-constraint classes applied to it, _it_ is height constrained in all cases but its children aren't, when scrolling is disabled.

src/pattern-library/components/patterns/feedback/ModalPage.tsx
```diff
@@ -388,6 +424,40 @@ export default function ModalPage() {
...
+                  Fake {'>'} breadcrumbs {'>'} example
+                </p>
+                <Scroll>
+                  <DataTable
+                    rows={nabokovRows}
+                    columns={nabokovColumns}
+                    title="Nabokov novels"
+                  />
+                </Scroll>
+              </Modal_>
```
lyzadanger, line 458: _This_ is the core use case I'm solving for here: A Modal with some non-scrolling content (like breadcrumbs at the top of a list of files) and a Table that _should_ scroll.

acelaya (approved):

PR #859, opened 2023-02-16
Handle layout of tables with sparse row data

lyzadanger:
This PR solves layout issues for tables with sparse row data (i.e. not enough rows to fill the table's allotted vertical space). This will allow us to use the newer `DataTable` component in some LMS contexts.

Before these changes:

image

This PR addresses this by:

* Adding an empty `<tfoot>` element to the table, via a new `TableFoot` component, which acts as a stretchy element to consume extra vertical space.
* Adding a very subtle, intentionally nearly-invisible, border to the last row of a table. This provides just the barest visual separation in case the table data ends on a row without a striped background.

After these changes:

image

### DIFF
--- a/src/components/data/TableCell.tsx
+++ b/src/components/data/TableCell.tsx
@@ -1,14 +1,14 @@
 import classnames from 'classnames';
+import type { JSX } from 'preact';
 import { useContext } from 'preact/hooks';
 
+import type { PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 
 import TableSectionContext from './TableSectionContext';
 
-/**
- * @typedef {import('../../types').PresentationalProps} CommonProps
- * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLElement>, 'size'>} HTMLAttributes
- */
+export type TableCellProps = PresentationalProps &
+  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'>;
 
 /**
  * Render a single table cell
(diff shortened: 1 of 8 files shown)


src/components/data/TableCell.tsx
```diff
@@ -38,6 +38,13 @@ const TableCellNext = function TableCell({
           // on scroll with sticky headers.
           'text-left border-t border-b border-b-grey-5': isHeadCell,
           'border-none': !isHeadCell,
+          // Apply a very subtle bottom border to the last row in the table (not
+          // in the head). This can help delineate the end of data in tables
+          // with sparse row data. Only apply border if row is not selected.
+          // This uses Tailwind's nested-group syntax. See
+          // https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups
+          'group-last/unselected:border-b group-last/unselected:border-grey-2 group-last/unselected:border-dotted':
+            !isHeadCell,
```
lyzadanger, line 47: I feel that most of Tailwind's naming conventions are self-documenting, or at least intuitive once you learn about them. The nested-group syntax is a stark departure from this: I don't find it easy to human-parse at all, so I've left a documentation link here.

`group-last/unselected:border-b` translated into some sort of English: _When a parent element with the `group/unselected` class is the last sibling element with `group/unselected` applied_. 😅 

Or, put another way, this could be thought of as:

```
tr.group/unselected:last-child td {
  // border rules
}
```

If you ignore for a moment that `/` is an illegal character in a classname.

acelaya (approved):

PR #866, opened 2023-02-20
Panel: Do not render container for buttons if there are no buttons

lyzadanger:
This PR fixes a regression with padding at the bottom of `Panel`s that do not have any buttons.

[This commit](https://github.com/hypothesis/frontend-shared/commit/36fd07b0605fcbf878b6bab9a11dc2202701b2a1) makes a change such that `Panel` children and `Panel` buttons are rendered within separate `CardContent` components. When there are `buttons` provided to the panel, this works fine, but when there are no `buttons`, an empty `CardContent` component was being rendered. This causes excess padding on the bottom of `Panel`s. 

Before this fix:

Screen Shot 2023-02-20 at 1 29 57 PM

After:

Screen Shot 2023-02-20 at 1 30 14 PM

### DIFF
--- a/src/components/layout/Panel.tsx
+++ b/src/components/layout/Panel.tsx
@@ -98,9 +98,11 @@ const PanelNext = function Panel({
         <CardTitle>{title}</CardTitle>
       </CardHeader>
       {scrollable ? <Scroll>{panelContent}</Scroll> : <>{panelContent}</>}
-      <CardContent>
-        {buttons && <CardActions>{buttons}</CardActions>}
-      </CardContent>
+      {buttons && (
+        <CardContent data-testid="panel-buttons">
+          <CardActions>{buttons}</CardActions>
+        </CardContent>
+      )}
     </Card>
   );
 };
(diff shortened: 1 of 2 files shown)


src/components/layout/Panel.tsx
```diff
@@ -98,9 +98,11 @@ const PanelNext = function Panel({
         <CardTitle>{title}</CardTitle>
       </CardHeader>
       {scrollable ? <Scroll>{panelContent}</Scroll> : <>{panelContent}</>}
-      <CardContent>
-        {buttons && <CardActions>{buttons}</CardActions>}
-      </CardContent>
+      {buttons && (
+        <CardContent data-testid="panel-buttons">
+          {buttons && <CardActions>{buttons}</CardActions>}
```
acelaya, line 103: You probably don't need to check the buttons here again.

```suggestion
          <CardActions>{buttons}</CardActions>
```

acelaya (changes requested):
Just spotted something minor. We are now checking the `buttons` twice.

src/components/layout/Panel.tsx
lyzadanger: 🤦🏻‍♀️

acelaya (approved):

Issue #743, opened 2022-11-29
Update package and components per new features in Tailwind 3.2


lyzadanger:
See: https://tailwindcss.com/blog/tailwindcss-v3-2

There are a few custom things that can be removed from the Tailwind configuration/preset, including `aria-`-attribute classes, which are now present out of the box.

lyzadanger:
The `aria-` variants have been accounted for as of #871. I thought some feature additions related to named groups would allow us to simplify how `InputGroup` gets its job done, but in squinting at some things, I think that would be better approached differently. So this is done!

PR #871, opened 2023-02-22
Remove custom `aria-` Tailwind variants

lyzadanger:
As of Tailwind 3.2, `aria-*` variants are now provided out of the box. We implemented our own custom versions exactly as the Tailwind-provided variants work, so these variants will continue to work as they did without having to define them in config.

Part of #743

### DIFF
--- a/src/pattern-library/components/patterns/UtilitiesPage.js
+++ b/src/pattern-library/components/patterns/UtilitiesPage.js
@@ -53,26 +53,7 @@ export default function UtilitiesPage() {
           This package also provides some tailwind variants (modifiers) that can
           be combined with other tailwind utility classes.
         </p>
-        <Library.Example title="ARIA variants">
-          <p>
-            Tailwind does not provide out-of-the-box <code>aria-*</code>{' '}
-            variants. This package provides <code>aria-pressed</code> and{' '}
-            <code>aria-expanded</code>.
-          </p>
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/patterns/UtilitiesPage.js
```diff
@@ -53,26 +53,7 @@ export default function UtilitiesPage() {
           This package also provides some tailwind variants (modifiers) that can
           be combined with other tailwind utility classes.
         </p>
-        <Library.Example title="ARIA variants">
```
lyzadanger, line 56: Note: These examples would behave correctly still as written, but there is no need to document them as a special utility as they are core to Tailwind now.

src/tailwind.preset.js
```diff
@@ -121,9 +121,6 @@ export default /** @type {Partial<import('tailwindcss').Config>} */ ({
       // only apply (set the element's background color to white) if a parent
       // element had the `.theme-clean` class.
       addVariant('theme-clean', '.theme-clean &');
-      // Tailwind does not provide variants for ARIA-attributes out of the box
```
lyzadanger, line 124: They do now!

acelaya (approved):

acelaya:
> We implemented our own custom versions exactly as the Tailwind-provided variants work, so these variants will continue to work as they did without having to define them in config.

That's convenient 😄

PR #873, opened 2023-02-22
Update `Card` and `InputGroup` pattern library documentation

### LINKED ISSUE #648: Card documentation in pattern library is missing `width` prop
Somehow managed to not document the `width` prop when documenting the new `Card` components in `CardPage`. Fix that and double-check other props.

lyzadanger:
This PR makes a few updates to pattern-library documentation, things I've noted over the past few weeks.

* Adds documentation for the `width` prop in `Card`, which was missing. This fixes #648 
* Updates `CardPage` to use the `LoremIpsum` utility component to DRY it out a little (and I added an `xs` size option to `LoremIpsum`).
* Updates the `InputGroup` documentation to note that `Select` is supported, and provides an example


src/pattern-library/components/patterns/input/InputGroupPage.tsx
```diff
@@ -34,6 +42,32 @@ export default function InputGroupPage() {
                 </InputGroup>
               </div>
             </Library.Demo>
+
+            <Library.Demo
```
lyzadanger, line 46: Rob noted yesterday how nearly-generic a pattern was in `StudentSelector`, an LMS component. Note that this is basically that pattern here: it's very atomic when you get down to it...

src/pattern-library/components/patterns/samples.tsx
```diff
@@ -88,10 +88,23 @@ export function SampleTableBody() {
 
 /**
  * @param {object} props
- *   @param {'sm'|'md'|'lg'} [props.size]
+ *   @param {'xs'|'sm'|'md'|'lg'} [props.size]
```
acelaya, line 91: Not required for this PR, but considering this file is already TS, perhaps we could define a proper type for the props.

acelaya (approved):
Works as expected

src/pattern-library/components/patterns/samples.tsx
lyzadanger: Oops. No, I should fix that. Given that I've already merged, I'll just do a quick follow-up. Thanks for spotting that!

Issue #876, opened 2023-02-23
Logistics and tasks for version 6


lyzadanger:
This is a grab-bag issue of tasks related to version 6 of this project's package. Items here can be broken out into discrete issues in the Version 6 milestone as needed.

* [ ] Figure out how to migrate and deprecate `lib/next` entrypoint
* [ ] Add selected hooks to entrypoint
* [ ] Review peer dependencies

PR #877, opened 2023-02-27
Add prettier plugin to sort imports in a consistent and automated way

acelaya:
This adds `@trivago/prettier-plugin-sort-imports` to the project, so that we can ensure a consistent and automated imports sorting on front-end components via `prettier`.

This was already introduced in client. See https://github.com/hypothesis/client/pull/5237 and https://github.com/hypothesis/client/pull/5279


package.json
```diff
@@ -75,7 +76,12 @@
   },
   "prettier": {
     "arrowParens": "avoid",
-    "singleQuote": true
+    "singleQuote": true,
+    "importOrder": [
+      "^.\\/components\\/patterns\\/(.*)Components$",
```
acelaya, line 81: This regular expression ensures the imports for legacy components in the pattern library router are placed in their own group (as they are now), to ease eventual removal.

Once those are removed we can also remove this regexp.

robertknight:
I suggest getting @lyzadanger to look at this, since she has been doing most of the work in this repo.

src/pattern-library/routes.ts
```diff
@@ -1,51 +1,44 @@
...
+import ButtonComponents from './components/patterns/ButtonComponents';
+import ContainerComponents from './components/patterns/ContainerComponents';
+import DialogComponents from './components/patterns/DialogComponents';
+import FormComponents from './components/patterns/FormComponents';
+import IconComponents from './components/patterns/IconComponents';
+import LinkComponents from './components/patterns/LinkComponents';
+import PanelComponents from './components/patterns/PanelComponents';
+import SpinnerComponents from './components/patterns/SpinnerComponents';
+import TableComponents from './components/patterns/TableComponents';
+import ThumbnailComponents from './components/patterns/ThumbnailComponents';
```
acelaya, line 13: The block of legacy page imports used to be at the bottom. However, I couldn't find an easy way to keep that behavior. The closest I found was a way to keep them as their own group, but at the top instead of the bottom (see [this comment](https://github.com/hypothesis/frontend-shared/pull/877/files#r1118444666)).

Since we are soon removing this, I thought this would be good enough for now, but if it is preferred to keep the previous behavior, I can check a bit further and try to find a better regular expression.

For context, the reason it is at the top is because this plugin matches imports against regular expressions defined in `importOrder` prettier config. Each regexp defined there becomes an import group.

The one we usually use (`^[./]`) is too broad, and matches legacy pages as well, resulting in a mixture of imports.

I tried adding `^.\/components\/patterns\/(.*)Components$` after, which only matches legacy pages, but for the reason explained above, it does not work unless we but it before `^[./]`.

I have tried using a regexp which is "the opposite" of `^.\/components\/patterns\/(.*)Components$`, but it's tricky and has weird side effects.

acelaya: @lyzadanger your feedback here would be very valuable.

package.json
lyzadanger: I'm adding removal of this to the list of small tasks for version 6: https://github.com/hypothesis/frontend-shared/issues/876

lyzadanger (approved):
Let's get this bad boy merged!

acelaya:
Something went horribly wrong while resolving conflicts in yarn.lock, so I'm just replacing this PR with https://github.com/hypothesis/frontend-shared/pull/885

Issue #858, opened 2023-02-16
Export types in some sort of sane way


lyzadanger:
Provide necessary types in a sane place in the package — known list so far includes:

* Props type definitions for each component in the API (currently exported from each component but a user would have to dig down into the package to get at them)
* API props definitions for Presentational, Composite and Base components (currently in `lib/types`)
* `IconComponent` type (current in `lib/types`) applicable to all SVG icon components

lyzadanger:
I think we've got a good start here. The `next` entrypoint (which will be _the_ entrypoint in v6.0) now exports all component props types and the package's shared types (e.g. `PresentatationalProps`, `IconComponent`).

PR #886, opened 2023-03-01
Migrate ButtonBase to TypeScript

acelaya:
This will help track down some of the type incompatibilities introduced with preact 10.13.0

### DIFF
--- a/src/components/input/ButtonBase.tsx
+++ b/src/components/input/ButtonBase.tsx
@@ -1,33 +1,53 @@
 import classNames from 'classnames';
+import type { JSX } from 'preact';
 
+import type { BaseProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 
 /**
  * Common props used for Button components.
- *
- * @typedef ButtonCommonProps
- * @prop {boolean} [expanded] - Is the element associated with this button
- *   expanded? (set `aria-expanded`)
- * @prop {never} [aria-expanded] - Use `expanded` prop instead
- * @prop {boolean} [pressed] - Is this button currently "active?" (set
- *   `aria-pressed` or `aria-selected` depending on button `role`)
- * @prop {never} [aria-pressed] - Use `pressed` prop instead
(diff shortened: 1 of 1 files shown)


src/components/input/ButtonBase.tsx
```diff
@@ -1,33 +1,53 @@
...
+
+  /**
+   * Is this button currently "active"?
+   * (sets `aria-pressed` or `aria-selected` depending on button `role`)
+   */
+  pressed?: boolean;
+
+  /**
+   * Button title (sets `aria-label` attribute)
+   */
```
lyzadanger, line 25:
```suggestion
  /** Button title (sets `aria-label` attribute) */
```

src/components/input/ButtonBase.tsx
```diff
@@ -1,33 +1,53 @@
...
  * HTML attributes accepted by button components. This eliminates conflicting
  * `icon` and `size` attributes.
- *
- * @typedef {Omit<import('preact').JSX.HTMLAttributes<HTMLButtonElement>, 'icon'|'size'>} HTMLButtonAttributes
- *
  */
+export type HTMLButtonAttributes = Omit<
+  JSX.HTMLAttributes<HTMLButtonElement>,
+  'icon' | 'size'
+>;
```
lyzadanger, line 43: Components in this package frequently intersect in `HTMLAttributes`; I acknowledge that at some point we might want to consolidate these types.

lyzadanger: Oh, also: Well done to keep exporting this. (deeply) Unfortunately, these types are used by a component or two in the `client`. There's very much a ticket open to consolidate this package's type exports: https://github.com/hypothesis/frontend-shared/issues/858

lyzadanger (approved):
Cool. Looks like this component module is still exporting multiple types — which is good because (sigh) they are used by consumers.

There's good news ahead:

A) Base components might be going away soon (at the very least from the package API)
B) We're gonna fix this package's types soon

PR #887, opened 2023-03-01
Adjust types to make project compatible with preact 10.12 and 10.13

acelaya:
This PR applies some type changes that make the build pass both with preact 10.12 and 10.13.

### DIFF
--- a/src/components/input/InputRoot.tsx
+++ b/src/components/input/InputRoot.tsx
@@ -30,7 +30,8 @@ const InputRootNext = function InputRoot({
 
   ...htmlAttributes
 }: InputRootProps) {
-  // @ts-expect-error - "aria-label" is missing from HTMLInputAttributes
+  // TODO Remove next line once updated to preact 10.13.0
+  // @ts-ignore-next - "aria-label" is missing from HTMLInputAttributes.
   if (!htmlAttributes.id && !htmlAttributes['aria-label']) {
     console.warn(
       '`Input` component should have either an `id` or an `aria-label` attribute'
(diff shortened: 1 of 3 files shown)


src/components/input/InputRoot.tsx
```diff
@@ -30,7 +30,8 @@ const InputRootNext = function InputRoot({
 
   ...htmlAttributes
 }: InputRootProps) {
-  // @ts-expect-error - "aria-label" is missing from HTMLInputAttributes
+  // TODO Remove next line once updated to preact 10.13.0
+  // @ts-ignore-next - "aria-label" is missing from HTMLInputAttributes.
```
acelaya, line 34: Preact 10.13 adds the missing `aria-label`. The `@ts-expect-error` annotation throws if TS does not find an error, so I have replaced it with `@ts-ignore-next`.

Once updated to preact 10.13, I will create another PR removing this altogether.

src/components/navigation/Tab.tsx
```diff
@@ -19,7 +19,14 @@ type ComponentProps = {
...
   JSX.HTMLAttributes<HTMLButtonElement>,
-  'size' | 'icon' | 'title'
+  | 'size'
+  | 'icon'
+  | 'title'
+  | 'selected'
+  // Omitting these aria attributes, as they are set as `never` in ButtonBase
+  | 'aria-expanded'
+  | 'aria-pressed'
+  | 'aria-label'
```
acelaya, line 29: Preact 10.13 introduces these aria attributes in the type, making it incompatible with `ButtonBase` props, where they are all explicitly set as `never` to indicate there are alternative props to be used.

I'm wondering if we should omit them there instead. It should have the same effect as `never`. but I'm not sure if we could introduce some other side effect.

lyzadanger (approved):
👍🏻 

After this series of PRs is landed, I think it would be wise to test a build of the current package (i.e. on the `main` branch) against the `client` and/or `lms` locally to ensure typecheck/tests pass. There are instructions for how to do this in the project's documentation readmes, but let me know if you need a hand!

PR #893, opened 2023-03-01
Convert `next` entrypoint to TS

lyzadanger:
Update references to this module, and plop output.


src/next.ts
```diff
@@ -2,6 +2,14 @@
 export { useArrowKeyNavigation } from './hooks/use-arrow-key-navigation';
 export { useSyncedRef } from './hooks/use-synced-ref';
 
+// Types
```
lyzadanger, line 5: More to come here, but this is a start.

acelaya (approved):

PR #899, opened 2023-03-02
Export props types for all components

lyzadanger:
This PR finishes rolling up and exporting props types for all component groups.


src/components/layout/CardContent.tsx
```diff
@@ -9,7 +9,7 @@ type ComponentProps = {
   size?: 'sm' | 'md' | 'lg';
 };
 
-export type CardActionsProps = PresentationalProps &
+export type CardContentProps = PresentationalProps &
```
lyzadanger, line 12: Oops.

acelaya (approved):

PR #901, opened 2023-03-03
Consolidate pattern-library components and modules for v6.0

lyzadanger:
Don't panic at the diff size. There are no functional changes here[^1], just housekeeping.

This PR's changes include:

* `Library` and `LibraryNext` are both modules that export multiple pattern-library components. The components in `LibraryNext` have now been added to `Library` to result in a consolidated collection of `Library`-related components for use to build pattern-library pages, and `Library` has been converted to TS(X). `LibraryNext` is still left in place, but will be removed in v6.0.
* Non-legacy documentation pages have been updated to use `Library` instead of a combination of `Library` and `LibraryNext`. I haven't updated legacy documentation, as it is going to be removed.
* Any remaining non-legacy documentation pages and pattern-library modules have been converted to TSX without any API or functional changes.
* Plopfile pattern-library templates are updated

After these changes, any non-TS(X) modules under `src/pattern-library` can be understood to be modules slated for removal in v6.0. This includes `LibraryNext`, `icons` and legacy documentation pages.

This work is in preparation for https://github.com/hypothesis/frontend-shared/issues/876

[^1]: OK, OK, I removed a section about legacy component usage from one pattern-library documentation page while I was passing by.


src/pattern-library/components/Library.tsx
```diff
@@ -0,0 +1,386 @@
+import classnames from 'classnames';
```
lyzadanger, line 1: Nothing new is happening in this module. It's been converted to TSX (types added and exported), I've collapsed some internal components that are only used once, and added components from `LibraryNext`.

src/pattern-library/router.tsx
```diff
@@ -1,16 +1,12 @@
 import { useCallback, useEffect, useMemo, useState } from 'preact/hooks';
 
-/**
- * @typedef {import('./routes').PlaygroundRoute} PlaygroundRoute
- */
+import type { PlaygroundRoute } from './routes';
```
lyzadanger, line 3: The functionality of this module needs some work but I believe we may be about to settle on a general front-end routing approach so I'm holding off.

acelaya (approved):

Issue #77, opened 2021-05-25
Re-implement accessible dialogs


lyzadanger:
This ticket serves as a container for tasks related to making our modals and dialogs more accessible and usable. This may turn into a project at some point. This issue is not actionable at this time as written.

-----

The `Dialog` component added here: https://github.com/hypothesis/frontend-shared/pull/75 was copied over from previous implementation in the `client` (which came from `lms`).

## Tasks

* Capturing and managing focus
* The logic for determining `aria-describedby` to see if we can find something a little less fragile-feeling
* The logic for generating element IDs to make sure there aren't potential collisions

### Disabling scrolling

Consider disabling scrolling on `window.document` when modal is open.

One possible implementation:
https://github.com/hypothesis/client/blob/6cae566ee5ae5d2dd88aa00954ea0f6a69dfb530/src/annotator/components/NotebookModal.js#L55-L78

### Disallowing click-elsewhere or esc-to-close

There are cases when we use a Modal component in which it shouldn't be possible to close the modal without choosing an adventure (i.e. clicking a button).

 "Cancel" may still be valid, and may still close the modal. But in some cases, it's problematic to have clicks outside of the modal component close the modal—e.g. in certain LMS contexts—or it's "too easy" to accidentally close a modal. And/or the close button (X) is misleading or distracting.

Before we do this, we need to look into proper focus-trapping (see Keyboard Support section of https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html).

One implementation approach that's been looked at here involves updating the `useElementShouldClose` hook to be configurable. 

### Initial focus to...?

Might be helpful to set initial focus on an action button of some sort?

PR #906, opened 2023-03-07
Establish new `Dialog` component and assets

lyzadanger:
This PR establishes a new `Dialog` component and its associated boilerplate, tests, documentation assets.

It is currently an exact copy of the `Modal` component and tests. A stub pattern-library page is also added here. While under development, this component will not be exported from the package's entry point. This will allow us to add features iteratively before we ship this component.

Part of #77

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -1,0 +1,158 @@
+import classnames from 'classnames';
+import type { RefObject } from 'preact';
+import { useEffect, useLayoutEffect } from 'preact/hooks';
+
+import { useElementShouldClose } from '../../hooks/use-element-should-close';
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import { useUniqueId } from '../../hooks/use-unique-id';
+import type { PresentationalProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+import Overlay from '../layout/Overlay';
+import Panel from '../layout/Panel';
+import type { PanelProps } from '../layout/Panel';
+
+type ComponentProps = {
+  /**
+   * Dialog _should_ be provided with a close handler. We have a few edge use
+   * cases, however, in which we need to render a "non-closeable" modal.
+   */
+  onClose?: () => void;
+  width?: 'sm' | 'md' | 'lg' | 'custom';
+  /**
+   * Element that should take focus when the Dialog is first rendered. When not
+   * provided ("auto"), the dialog's outer element will take focus. Setting this
+   * prop to "manual" will opt out of focus routing.
+   */
+  initialFocus?: RefObject<HTMLOrSVGElement | null> | 'auto' | 'manual';
+};
+
+// This component forwards a number of props on to `Panel` but always sets the
+// `fullWidthHeader` prop to `true`.
+export type DialogProps = PresentationalProps &
+  ComponentProps &
+  Omit<PanelProps, 'fullWidthHeader'>;
+
+const noop = () => {};
+
+/**
+ * Show a modal
+ */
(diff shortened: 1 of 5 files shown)


src/components/feedback/Dialog.tsx
```diff
@@ -0,0 +1,158 @@
...
+import type { PresentationalProps } from '../../types';
+import { downcastRef } from '../../util/typing';
+import Overlay from '../layout/Overlay';
+import Panel from '../layout/Panel';
+import type { PanelProps } from '../layout/Panel';
+
+type ComponentProps = {
+  /**
+   * Dialog _should_ be provided with a close handler. We have a few edge use
+   * cases, however, in which we need to render a "non-closeable" modal.
```
acelaya, line 17:
```suggestion
   * cases, however, in which we need to render a "non-closeable" dialog.
```

src/components/feedback/Dialog.tsx
```diff
@@ -0,0 +1,158 @@
...
+  buttons,
+  icon,
+  onClose,
+  paddingSize = 'md',
+  scrollable = true,
+  title,
+
+  ...htmlAttributes
+}: DialogProps) {
+  const modalRef = useSyncedRef(elementRef);
```
lyzadanger, line 58: I'll rename this to `dialogRef` on the next pass. It slipped through.

src/components/feedback/Dialog.tsx
```diff
@@ -0,0 +1,158 @@
...
+      disabled?: boolean;
+    };
+
+    if (focusEl && !focusEl.disabled) {
+      focusEl.focus();
+    } else {
+      // Fall back to focusing the modal itself
+      modalRef.current?.focus();
+      return;
+    }
```
acelaya, line 87: I think you can disregard this `else` block entirely, as you have already checked for `'manual'` and `'auto'` values before getting to this point.

src/pattern-library/components/patterns/feedback/DialogPage.tsx
```diff
@@ -0,0 +1,122 @@
...
+    return (
+      <Dialog
+        buttons={forwardedButtons}
+        {...dialogProps}
+        onClose={closeHandler}
+      >
+        {children}
+      </Dialog>
+    );
+  }
```
acelaya, line 73: Just picky comment, but you can avoid the `else` block, as the `if` block is returning:

```ts
if (!dialogOpen) {
  return (
    <Button ... />
  );
}

return (
  <Dialog ... />
);
```

acelaya (approved):
Looks like a good starting point. Just left some minor picky comments.

src/components/feedback/Dialog.tsx
lyzadanger: I'm going to change the accepted values for `initialFocus` very soon, and will clean this up at that time!

src/components/feedback/Dialog.tsx
lyzadanger: Accccctually, this is probably technically correct (though its inclusion was an accident). Only modal dialogs can be non-close-able :)

src/pattern-library/components/patterns/feedback/DialogPage.tsx
lyzadanger: Fair enough! I'll rectify in the next pass.

lyzadanger:
@acelaya Totally selfish because I want to avoid merge angst, but I'm going to fix these little nits in the next PR. I've noted them all! Thanks!

acelaya:
> @acelaya Totally selfish because I want to avoid merge angst, but I'm going to fix these little nits in the next PR. I've noted them all! Thanks!

All good!

PR #907, opened 2023-03-07
Add `modal` prop to `Dialog`

lyzadanger:
This PR starts the division of `Dialog` into modal and non-modal sub-variants. It is possible that one day we will peel this into two components, but for now, one component that can have modal or non-modal behaviors.

More behavior differentiation will be applied to modal vs. non-modal Dialogs, but for now, modal Dialogs get a full-screen backdrop Overlay, sizing and positioning, while non-modal Dialogs do not. A modal example has been added to the `Dialog` page.

I've also added some next-step TODOs here on `Dialog`'s pattern-library page to sketch out where this is going next.

Next step: some hooks to handle events that should close modal `Dialog`s.

Part of #77

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -1,5 +1,6 @@
 import classnames from 'classnames';
 import type { RefObject } from 'preact';
+import { Fragment } from 'preact';
 import { useEffect, useLayoutEffect } from 'preact/hooks';
 
 import { useElementShouldClose } from '../../hooks/use-element-should-close';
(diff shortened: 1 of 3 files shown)


src/components/feedback/Dialog.tsx
```diff
@@ -14,16 +15,23 @@ import type { PanelProps } from '../layout/Panel';
 type ComponentProps = {
   /**
    * Dialog _should_ be provided with a close handler. We have a few edge use
-   * cases, however, in which we need to render a "non-closeable" modal.
+   * cases, however, in which we need to render a "non-closeable" modal dialog.
```
lyzadanger, line 18: Fix requested in last PR.

src/components/feedback/Dialog.tsx
```diff
@@ -14,16 +15,23 @@ import type { PanelProps } from '../layout/Panel';
...
    * Element that should take focus when the Dialog is first rendered. When not
    * provided ("auto"), the dialog's outer element will take focus. Setting this
    * prop to "manual" will opt out of focus routing.
    */
   initialFocus?: RefObject<HTMLOrSVGElement | null> | 'auto' | 'manual';
+
+  /**
+   * Make this Dialog modal. Modal dialogs are:
+   * - Rendered with a full-screen overlay backdrop
+   */
```
lyzadanger, line 33: This comment list will get longer and be adjusted as more implementation is done...

src/components/feedback/Dialog.tsx
```diff
@@ -108,31 +117,37 @@ const DialogNext = function Dialog({
...
+            'tall:fixed tall:max-h-[80vh] tall:top-[10vh]': modal,
+          },
           {
             // Max-width rules will ensure actual width never exceeds 90vw
-            'w-[30rem]': width === 'sm',
-            'w-[36rem]': width === 'md', // default
-            'w-[42rem]': width === 'lg',
+            'w-[30rem]': modal && width === 'sm',
+            'w-[36rem]': modal && width === 'md', // default
+            'w-[42rem]': modal && width === 'lg',
```
lyzadanger, line 150: This application of sizing styles will be refined as implementation continues, with the introduction of some of the customization controls presented last week. But for now: Only assert size and positioning for modal Dialogs.

acelaya (approved):
👍🏼

PR #908, opened 2023-03-09
Update to actions/setup-node@v3 to fix deprecation warning on using Node 12

acelaya:


lyzadanger (approved):
Common-sense-wise, this looks right!

PR #910, opened 2023-03-09
Ensure `listener-collection` passes `options` to `removeEventListener`

lyzadanger:
I ran into a confusion today while converting some hooks over to more consistently use `listener-collection`. Some worked, but a focus-event-related hook did not. I realized it was because `listener-collection` wasn't passing the `options` parameter (when provided) to `removeEventListener`. This matters if you have an event that was registered with options (specifically `capture: true`) as they don't get removed if you don't pass them options!

Converted it to TS while I was in there.

The tests added to the tests module will fail without the change made to the module.

### DIFF
--- a/src/util/listener-collection.ts
+++ b/src/util/listener-collection.ts
@@ -37,10 +37,10 @@ export class ListenerCollection {
   /**
    * Add a listener and return an ID that can be used to remove it later
    */
-  add(
-    eventTarget: EventTarget,
-    eventType: string,
-    listener: (event: EventType<EventTarget, string>) => void,
+  add<ListenerTarget extends EventTarget, Type extends string>(
+    eventTarget: ListenerTarget,
+    eventType: Type,
+    listener: (event: EventType<ListenerTarget, Type>) => void,
     options?: AddEventListenerOptions
   ) {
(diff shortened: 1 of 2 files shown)


robertknight (approved):
LGTM. Once our browser support target passes Safari >= 15 we can simplify this code by using the `signal` option as a way to disconnect listeners, instead of having to remember all the `addEventListener` arguments and pass them to `removeEventListener`.

src/util/listener-collection.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Add a listener and return an ID that can be used to remove it later
+   */
+  add<ListenerTarget extends EventTarget, Type extends string>(
+    eventTarget: ListenerTarget,
+    eventType: Type,
+    listener: (event: EventType<ListenerTarget, Type>) => void,
+    options?: AddEventListenerOptions
+  ) {
```
robertknight, line 45: It might be helpful for readers to indicate explicitly that the return type is a `symbol`.

robertknight (approved):
LGTM. As noted in Slack, you could get away with abbreviations for the generics if you wanted to make the signature shorter, but I think this is readable as-is.

PR #911, opened 2023-03-09
Add granular event hooks (for use in closing `Dialog`s)

### LINKED ISSUE #900: Decompose `useElementShouldClose` into more targeted hooks
`useElementShouldClose` conflates too many events into one hook, without the ability to configure it more specifically. It currently invokes the provided close callback for any external click, external focus or the keypress of `Esc`. You can't opt in or out of any of these.

Arguments:

* With the re-implementation of `Dialog`, the away-focus handling won't make sense as the component will focus-trap.
* Sometimes we don't want to close a modal on click-away (it becomes too easy, e.g., to accidentally dismiss the modal, as is the case in the LMS file pickers)
* Sometimes we want to react to `Esc`, but not a click-away

We might consider decomposing this hook, e.g., maybe (just one option here):

* `useClickAway` and `useFocusAway`, or `useAwayEvent` (hook that could take multiple events)
* `useKeyPress` specifically for keyboard event handling

There are a number of plausible options here...

lyzadanger:
Depends on #910 
Fixes #900
Part of #77

This PR decomposes the functionality of `useElementShouldClose` into three separate, more-generic hooks, and aligns with `useArrowKeyNavigation` in the use of `ListenerCollection` for managing event listeners. This will allow more nuanced control for `Dialog`s.

There's some opportunity to DRY this out some in the future (especially in tests), but I started with the simplest approach possible here.

Just like with `Dialog`, I imagine we might make these hooks part of the package's public API (and deprecate `useElementShouldClose`). But let's do that later and separately.

### DIFF
--- a/src/hooks/test/use-click-away-test.js
+++ b/src/hooks/test/use-click-away-test.js
@@ -1,0 +1,67 @@
+import { mount } from 'enzyme';
+import { useRef } from 'preact/hooks';
+import { act } from 'preact/test-utils';
+
+import { useClickAway } from '../use-click-away';
+
+describe('useClickAway', () => {
+  let handler;
+
+  const events = [new Event('mousedown'), new Event('click')];
+
+  // Create a fake component to mount in tests that uses the hook
+  function FakeComponent({ enabled = true }) {
+    const myRef = useRef();
+    useClickAway(myRef, handler, { enabled });
+    return (
+      <div ref={myRef}>
+        <button>Hi</button>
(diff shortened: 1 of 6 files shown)


lyzadanger:
I see the test misses. I'll address shortly.

lyzadanger:
I thought test coverage was happy but now it's not again. I'll deal with it later; it's trivial.

acelaya (approved):
Looks good

PR #929, opened 2023-03-21
Split `ModalDialog` from `Dialog`

lyzadanger:
This PR splits out `ModalDialog` from `Dialog`. `ModalDialog` wraps `Dialog`, and provides modal-dialog-specific defaults and styling. It doesn't add any new functionality, but moves things around between components. Pattern-library pages added and updated.

This PR also makes an addition to the options accepted by common composite component tests.

Next steps will involve focus management: restoring after close and trapping.

## Addition of `wrapperSelector` option for composite component common tests

By default, all common tests assume that the outermost element rendered by a component is its primary element, unless a different element is identified with the (pre-existing) `elementSelector` test option. Before these changes, this element is assumed to be:

* where any HTML attributes passed as props are merged to
* where any passed CSS classes are added to
* where the `data-component` attribute is applied
* where the `data-composite-component` attribute is applied (for composite components only)

The test changes here make it possible for composite component tests to differentiate between a primary element (the first three items above) versus a _wrapper_ element (the last item, where `data-composite-component` is set). 

Perhaps easier to look at by example. In the case of `ModalDialog`, the structure is like so:

```jsx
<Overlay data-composite-component="ModalDialog">
  <Dialog ... data-component="Dialog" />
<Overlay>
```

In this case, the _primary_ element is the Dialog, but the `data-composite-component` attribute is applied to a different, _wrapper_ element. I’ve allowed for this in common composite tests by adding a new, optional `wrapperSelector` option to complement the `elementSelector` option. 

By default, tests will continue to treat to the outermost element as both _primary_ and _wrapper_ elements unless one or both of these selector options are set. 

Part of #77

### DIFF
--- a/src/components/feedback/ModalDialog.tsx
+++ b/src/components/feedback/ModalDialog.tsx
@@ -1,0 +1,74 @@
+import classnames from 'classnames';
+
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import { downcastRef } from '../../util/typing';
+import Overlay from '../layout/Overlay';
+import Dialog from './Dialog';
+import type { DialogProps } from './Dialog';
+
+type ComponentProps = {
+  size?: 'sm' | 'md' | 'lg' | 'custom';
+};
+
+export type ModalDialogProps = DialogProps & ComponentProps;
+
+/**
+ * Show a modal dialog
+ */
+const ModalDialogNext = function ModalDialog({
+  children,
+  size = 'md',
+
+  classes,
+  elementRef,
+
+  // Forwarded to Dialog
+  closeOnEscape = true,
+  closeOnClickAway = false,
+  closeOnFocusAway = false,
+  initialFocus = 'auto',
+
+  ...htmlAndPanelAttributes
+}: ModalDialogProps) {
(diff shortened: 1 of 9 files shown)


src/components/feedback/ModalDialog.tsx
```diff
@@ -0,0 +1,74 @@
...
+type ComponentProps = {
+  size?: 'sm' | 'md' | 'lg' | 'custom';
+};
+
+export type ModalDialogProps = DialogProps & ComponentProps;
+
+/**
+ * Show a modal dialog
+ */
+const ModalDialogNext = function ModalDialog({
```
acelaya, line 18: Does it make sense to continue using the `Next` suffix for components which don't have a non-next equivalent?

acelaya (approved):
👍🏼

src/components/feedback/ModalDialog.tsx
lyzadanger: My current plan is to do this in one go when working on v6 of the package; it's a task tracked here: https://github.com/hypothesis/frontend-shared/issues/876

Until then, it's a marker of a non-deprecated component.

PR #931, opened 2023-03-22
Add focus-trap and tab-key keyboard navigation to `ModalDialog`

lyzadanger:
This PR adds a new hook, `useTabKeyNavigation`, modeled after `useArrowKeyNavigation`, and enables it by default for `ModalDialog`.

This establishes trap-focus behavior and tab/shift-tab keyboard navigation for `ModalDialog`s, which we need for accessibility.

Note that `useTabKeyNavigation` duplicates behavior from `useArrowKeyNavigation`, per voice discussion earlier today. The life expectancy of `useTabKeyNavigation` is short-ish, so this time-saving measure is perhaps justifiable. 

This PR also contains a commit that removes a no-longer-needed testing workaround for `useArrowKeyNavigation`'s tests.

### DIFF
--- a/src/hooks/test/use-tab-key-navigation-test.js
+++ b/src/hooks/test/use-tab-key-navigation-test.js
@@ -1,0 +1,245 @@
+import { render } from 'preact';
+import { useRef } from 'preact/hooks';
+import { act } from 'preact/test-utils';
+
+import { waitFor } from '../../test-util/wait';
+import { useTabKeyNavigation } from '../use-tab-key-navigation';
+
+function Toolbar({ navigationOptions = {} }) {
+  const containerRef = useRef();
+
+  useTabKeyNavigation(containerRef, navigationOptions);
+
+  return (
+    <div ref={containerRef} data-testid="toolbar">
+      <button data-testid="bold">Bold</button>
+      <button data-testid="italic">Italic</button>
+      <button data-testid="underline">Underline</button>
+      <a href="/help" target="_blank" data-testid="help">
+        Help
+      </a>
+    </div>
+  );
+}
+
+describe('useTabKeyNavigation', () => {
+  let container;
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    document.body.append(container);
+    renderToolbar();
+  });
+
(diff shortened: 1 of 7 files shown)


src/hooks/test/use-tab-key-navigation-test.js
lyzadanger, line 1: Most of these tests, or at least their structure, were copied from the tests for `useArrowKeyNavigation`. I'll call out where they differ.

src/hooks/test/use-tab-key-navigation-test.js
```diff
@@ -0,0 +1,245 @@
...
+  function findElementByTestId(testId) {
+    return container.querySelector(`[data-testid=${testId}]`);
+  }
+
+  /**
+   * @param {string|Partial<KeyboardEvent>} key - Either a string name of a
+   *   key (e.g. 'ArrowLeft') or a partial `KeyboardEvent` object, e.g.
+   *   `{ key: 'Tab', shiftKey: true }`
+   */
+  function pressKey(key) {
```
lyzadanger, line 57: I adjusted the parameters for this utility function so it can imitate more types of `KeyboardEvent`s.

src/hooks/test/use-tab-key-navigation-test.js
```diff
@@ -0,0 +1,245 @@
...
+  it('should not respond to keys if hook is not enabled', () => {
+    renderToolbar({
+      enabled: false,
+    });
+    ['Tab', { key: 'Tab', shiftKey: true }].forEach(key => {
+      findElementByTestId('bold').focus();
+      pressKey(key);
+      assert.equal(currentItem(), 'Bold');
+    });
+  });
```
lyzadanger, line 176: This hook can be "enabled" or "disabled", which is a feature that `useArrowKeyNavigation` does not have.

src/hooks/use-tab-key-navigation.ts
lyzadanger, line 1: This implementation is very close to `useArrowKeyNavigation` but slightly simpler. It does add an `enabled` option.

PR #933, opened 2023-03-24
Adapt handling of initial focus in `useTabKeyNavigation`

lyzadanger:
This PR adapts the implementation of `useTabKeyNavigation` to work appropriately in its targeted use case within modal dialogs. These changes are concerned with how the hook handles initial focus state.

### When the initially-focused element is not the first element in the tab sequence

Our `Dialog` and `ModalDialog` take an optional `Ref` as an `initialFocus` prop and will set initial focus on that element when provided. That element may or may not be the first element in the hook's derived tab sequence within the dialog (container). When initial focus is on an interactive descendant of `container` but not the _first_ interactive descendant, ensure that tab indexes are set properly.

Before these changes, if a non-first element had initial focus, pressing "Tab" would always move focus to the second element in the tab sequence (because the hook incorrectly assumed the first element had focus). 

You can see this working correctly in the "Basic ModalDialog" example on http://localhost:4001/feedback-dialog, which sets initial focus on an input element (not the first navigable element in the sequence).

### When the initial focus is on the dialog element itself (container)

I gave some thought as to whether we should find another default initial focus element (other than the dialog `container`), but decided to retain this approach, in some part to make migration to these new components easier, but also because it seems as simple and valid as any other approach I could come up with.

The changes here allow the dialog `container` to have focus initially, but it isn't part of the subsequent tab sequence (i.e. if you tab or shift-tab around you won't go back to it). This seems to be in line with accessible examples I explored (e.g. the "Verification Result" modal on https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/# , which sets initial focus on a non-interactive element).

You can see this working correctly in most of the other `ModalDialog` examples on http://localhost:4001/feedback-dialog

### Avoiding initial-focus ambiguity

To prevent ambiguity, I've removed the `autofocus` option from this hook and explicitly noted that initial focus is the responsibility of the consuming component.

### DIFF
--- a/src/hooks/test/use-tab-key-navigation-test.js
+++ b/src/hooks/test/use-tab-key-navigation-test.js
@@ -1,17 +1,29 @@
 import { render } from 'preact';
-import { useRef } from 'preact/hooks';
+import { useEffect, useRef } from 'preact/hooks';
 import { act } from 'preact/test-utils';
 
 import { waitFor } from '../../test-util/wait';
 import { useTabKeyNavigation } from '../use-tab-key-navigation';
 
-function Toolbar({ navigationOptions = {} }) {
+function Toolbar({
+  focusContainer = false,
+  initialFocusIndex = -1,
+  navigationOptions = {},
+}) {
   const containerRef = useRef();
+
+  useEffect(() => {
+    if (focusContainer) {
+      containerRef.current.focus();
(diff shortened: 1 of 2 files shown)


acelaya (approved):
Looks good, and works as expected.

Must have been fun wrapping your head around this 😅

PR #934, opened 2023-03-24
Add `Dialog`, `ModalDialog` to package API, consolidate and document

### LINKED ISSUE #77: Re-implement accessible dialogs

lyzadanger:
This PR completes initial implementation of the `Dialog` and `ModalDialog` components. It:

* Adjusts the naming of a few props to better match conventions and make migration from `Modal` easier
* Completes pattern-library documentation and adds migration notes
* Deprecates `Modal`
* Exports `Dialog` and `ModalDialog` from the package (next) entrypoint

Regarding the changing of a few prop names:

* `size` -> (back to) `width`: I got a little zealous early on in the implementation of the new `ModalDialog` and wanted to start moving towards the standard theme-/customization-related prop names I presented to the team several weeks ago. However, it's distracting right now and will make migration from `Modal` more tedious. Change this prop name back to `width` so that it is unchanged from `Modal`.
* The following boolean props were renamed on `ModalDialog` because their defaults differ from `Dialog`, and our conventions suggest that booleans should default false and be named for their effect when set (`true`):
    * `closeOnEscape` => `disableCloseOnEscape`
    * `restoreFocus` => `disableRestoreFocus`

Fixes #77
~~Depends on #933~~ 

Leaving in draft until #933 is resolved and some more testing done against our applications.

### DIFF
--- a/src/components/feedback/Modal.tsx
+++ b/src/components/feedback/Modal.tsx
@@ -36,6 +36,7 @@ const noop = () => {};
 
 /**
  * Show a modal
+ * @deprecated - Use `ModalDialog` instead
  */
 const ModalNext = function Modal({
   children,
--- a/src/components/feedback/ModalDialog.tsx
+++ b/src/components/feedback/ModalDialog.tsx
@@ -8,7 +8,7 @@ import Dialog from './Dialog';
 import type { DialogProps } from './Dialog';
 
 type ComponentProps = {
-  size?: 'sm' | 'md' | 'lg' | 'custom';
+  width?: 'sm' | 'md' | 'lg' | 'custom';
 
   /**
    * Disable WAI-ARIA-specific modal-dialog focus trap and tab/shift-tab
@@ -24,7 +24,7 @@ export type ModalDialogProps = DialogProps & ComponentProps;
  */
 const ModalDialogNext = function ModalDialog({
   children,
-  size = 'md',
+  width = 'md',
   disableFocusTrap = false,
 
   classes,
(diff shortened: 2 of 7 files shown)


lyzadanger:
Closing for now: I'd like to add `Dialog` to package exports, but not `ModalDialog`, yet.

PR #941, opened 2023-03-29
Consolidate `*Dialog` props and documentation; export `Dialog`

lyzadanger:
This PR consolidates props naming, organization and documentation for `Dialog` and `ModalDialog`, and adds `Dialog` to the package API[^1]. It:

* Adjusts the naming of a few props to better match conventions and make migration from `Modal` easier
* Completes pattern-library documentation for both components and adds migration notes
* Exports `Dialog` from the package (next) entrypoint (not `ModalDialog`, yet, because of some deficiencies in the `useTabKeyNavigation` hook when combined with embedded interactive widgets like `DataTable` and `TabList`.

Regarding the changing of a few prop names:

* `size` -> (back to) `width`: I got a little zealous early on in the implementation of the new `ModalDialog` and wanted to start moving towards the standard theme-/customization-related prop names I presented to the team several weeks ago. However, it's distracting right now and will make migration from `Modal` more tedious. Change this prop name back to `width` so that it is unchanged from `Modal`.
* The following boolean props were renamed on `ModalDialog` because their defaults differ from `Dialog`, and our conventions suggest that booleans should default false and be named for their effect when set (`true`):
    * `closeOnEscape` => `disableCloseOnEscape`
    * `restoreFocus` => `disableRestoreFocus`

[^1]: Holding for now on adding `ModalDialog` to API until some nuances with trapped-focus and embedded interactive widgets (e.g. `DataTable`, `TabList`) can be perfected.

### DIFF
--- a/src/components/feedback/ModalDialog.tsx
+++ b/src/components/feedback/ModalDialog.tsx
@@ -7,8 +7,13 @@ import Overlay from '../layout/Overlay';
 import Dialog from './Dialog';
 import type { DialogProps } from './Dialog';
 
+type ModalWidth = 'sm' | 'md' | 'lg' | 'custom';
+
 type ComponentProps = {
-  width?: 'sm' | 'md' | 'lg' | 'custom';
+  /**
+   * @deprecated  - use `size` instead
+   */
+  width?: ModalWidth;
 
   /**
    * Do not close the modal when the Escape key is pressed
@@ -26,6 +31,11 @@ type ComponentProps = {
    * the dialog is closed.
    */
   disableRestoreFocus?: boolean;
+
+  /**
+   * Relative size (width) of modal dialog
+   */
+  size?: ModalWidth;
 };
 
 export type ModalDialogProps = Omit<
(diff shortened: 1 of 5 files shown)


acelaya:
> `size` -> (back to) `width`: I got a little zealous early on in the implementation of the new `ModalDialog` and wanted to start moving towards the standard theme-/customization-related prop names I presented to the team several weeks ago. However, it's distracting right now and will make migration from `Modal` more tedious. Change this prop name back to `width` so that it is unchanged from `Modal`.

Idea: Keep both `width` and `size`, marking `width` as deprecated. This will allow a more direct migration, but highlight the fact that this is not the path we want in the future.

Then inside the component we can do something in the lines of `const size = props.size ?? props.width ?? <default_value_if_any>`.

acelaya (approved):
👍🏼

lyzadanger:
> > `size` -> (back to) `width`: I got a little zealous early on in the implementation of the new `ModalDialog` and wanted to start moving towards the standard theme-/customization-related prop names I presented to the team several weeks ago. However, it's distracting right now and will make migration from `Modal` more tedious. Change this prop name back to `width` so that it is unchanged from `Modal`.
> 
> Idea: Keep both `width` and `size`, marking `width` as deprecated. This will allow a more direct migration, but highlight the fact that this is not the path we want in the future.
> 
> Then inside the component we can do something in the lines of `const size = props.size ?? props.width ?? <default_value_if_any>`.

This is a sensible idea, thank you. I've pushed another commit that does exactly this. Docs are updated also.

acelaya:
> This is a sensible idea, thank you. I've pushed another commit that does exactly this. Docs are updated also.

Looks good :)

PR #957, opened 2023-04-13
Enhance Dialog so that it allows providing a TransitionComponent

acelaya:
This PR extends the `Dialog` component, so that a `transitionComponent` can be provided, and the `Dialog` transparently handle all its focusing capabilities.

The `transitionComponent` can be any component that has these properties:

```ts
type TransitionComponentProps = {
  visible: boolean;
  onTransitionEnd?: () => void;
};
```

Some UI libraries also handle the transition start callback, but since we don't yet have a real use-case for it, I have decided to leave it out for now, reducing the code and tests to maintain.

This PR also includes some basic documentation and an example on how to use this property, which is now part of the pattern library. Visit http://localhost:4001/feedback-dialog and search for `transitionComponent`.

> **Note**
> This PR is easier to review hidding whitespaces.

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -1,6 +1,5 @@
 import classnames from 'classnames';
 import type { RefObject } from 'preact';
-import type { JSX } from 'preact';
 import { Fragment } from 'preact';
 import {
   useCallback,
@@ -16,15 +15,10 @@ import { useFocusAway } from '../../hooks/use-focus-away';
 import { useKeyPress } from '../../hooks/use-key-press';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
 import { useUniqueId } from '../../hooks/use-unique-id';
-import type { PresentationalProps } from '../../types';
+import type { PresentationalProps, TransitionComponent } from '../../types';
 import { downcastRef } from '../../util/typing';
 import Panel from '../layout/Panel';
 import type { PanelProps } from '../layout/Panel';
-
-type TransitionComponent = JSX.ElementType<{
-  visible: boolean;
-  onTransitionEnd?: () => void;
-}>;
 
 type ComponentProps = {
   closeOnClickAway?: boolean;
@@ -49,11 +43,7 @@ type ComponentProps = {
   restoreFocus?: boolean;
 
   /**
-   * Providing this has the next implications:
-   * - The component will be used to wrap the Dialog contents.
-   * - If initialFocus === 'auto', the Dialog will be focused once the open
-   *   transition has finished.
-   * - onClose will be invoked after the close transition has finished.
+   * Allows to provide a transition animation to the Dialog contents
    */
   transitionComponent?: TransitionComponent;
 };
@@ -95,12 +85,13 @@ const DialogNext = function Dialog({
   const restoreFocusEl = useRef<HTMLElement | null>(
     document.activeElement as HTMLElement | null
   );
-  const [visible, setVisible] = useState(false);
+  const [transitionComponentVisible, setTransitionComponentVisible] =
+    useState(false);
   // If a TransitionComponent was provided, closing the Dialog should just set
   // it to not visible. The TransitionComponent will take care of actually
-  // closing the dialog once transition has finished
+  // closing the dialog once the "out" transition has finished
   const closeHandler = TransitionComponent
-    ? () => setVisible(false)
+    ? () => setTransitionComponentVisible(false)
     : onClose ?? noop;
   const setInitialFocus = useCallback(() => {
     if (initialFocus === 'manual') {
@@ -126,6 +117,14 @@ const DialogNext = function Dialog({
       modalRef.current?.focus();
     }
   }, [initialFocus, modalRef]);
+  const onTransitionEnd = (direction: 'in' | 'out') => {
+    if (direction === 'in') {
+      // We can't check the initial focus until transition "in" has finished
+      setInitialFocus();
+    } else {
+      onClose?.();
+    }
+  };
 
   useClickAway(modalRef, closeHandler, {
     enabled: closeOnClickAway,
@@ -146,9 +145,11 @@ const DialogNext = function Dialog({
   );
 
   useEffect(() => {
-    // Trigger initial open animation
-    setVisible(true);
-    setInitialFocus();
+    // Trigger initial "in" transition animation
+    setTransitionComponentVisible(true);
+    if (!TransitionComponent) {
+      setInitialFocus();
+    }
 
     // We only want to run this effect once when the dialog is mounted.
     // eslint-disable-next-line react-hooks/exhaustive-deps
@@ -191,16 +192,8 @@ const DialogNext = function Dialog({
 
   return (
     <Wrapper
-      visible={visible}
-      onTransitionEnd={() => {
-        if (!visible) {
-          onClose?.();
-        } else {
-          // Once transition on a visible Dialog has finished, we re-check the
-          // initial focus
-          setInitialFocus();
-        }
-      }}
+      visible={transitionComponentVisible}
+      onTransitionEnd={onTransitionEnd}
     >
       <div
         data-component="Dialog"
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -1,6 +1,15 @@
 import classnames from 'classnames';
 import type { RefObject } from 'preact';
-import { useEffect, useLayoutEffect, useRef } from 'preact/hooks';
+import type { JSX } from 'preact';
+import { Fragment } from 'preact';
+import {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'preact/hooks';
 
 import { useClickAway } from '../../hooks/use-click-away';
 import { useFocusAway } from '../../hooks/use-focus-away';
@@ -11,6 +20,11 @@ import type { PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 import Panel from '../layout/Panel';
 import type { PanelProps } from '../layout/Panel';
+
+type TransitionComponent = JSX.ElementType<{
+  visible: boolean;
+  onTransitionEnd?: () => void;
+}>;
 
 type ComponentProps = {
   closeOnClickAway?: boolean;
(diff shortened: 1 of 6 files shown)


src/components/feedback/Dialog.tsx
```diff
@@ -86,13 +115,20 @@ const DialogNext = function Dialog({
   });
 
   const dialogDescriptionId = useUniqueId('dialog-description');
+  const Wrapper = useMemo(
+    () => TransitionComponent ?? Fragment,
+    [TransitionComponent]
+  );
```
acelaya, line 121: I'm not sure if it's worth memoizing this. Maybe `const Wrapper = TransitionComponent ?? Fragment` would suffice 🤷🏼‍♂️

src/components/feedback/Dialog.tsx
```diff
@@ -152,31 +190,44 @@ const DialogNext = function Dialog({
...
-      {...htmlAttributes}
-      className={classnames(
-        // Column-flex layout to constrain content to max-height
-        'flex flex-col',
-        classes
-      )}
-      ref={downcastRef(modalRef)}
+    <Wrapper
+      visible={visible}
+      onTransitionEnd={() => {
```
lyzadanger, line 195: Ah, interesting, so we can rely on an `onTransitionEnd` event for elements with no transition defined? I'll have to look that up!

src/components/feedback/Dialog.tsx
```diff
@@ -33,6 +47,15 @@ type ComponentProps = {
...
    */
   restoreFocus?: boolean;
+
+  /**
+   * Providing this has the next implications:
+   * - The component will be used to wrap the Dialog contents.
+   * - If initialFocus === 'auto', the Dialog will be focused once the open
+   *   transition has finished.
+   * - onClose will be invoked after the close transition has finished.
+   */
```
lyzadanger, line 57: We'll want to tighten down this comment before we land this as most of this is implementation detail. The consumer probably doesn't need to care about most of this :)

src/components/feedback/Dialog.tsx
```diff
@@ -109,8 +124,31 @@ const DialogNext = function Dialog({
...
+  const dialogDescriptionId = useUniqueId('dialog-description');
+  const Wrapper = useMemo(
+    () => TransitionComponent ?? Fragment,
+    [TransitionComponent]
+  );
+
+  useEffect(() => {
+    // Trigger initial open animation
+    setVisible(true);
+    setInitialFocus();
```
lyzadanger, line 151: As we discussed: don't invoke this if a transition has been set.

src/components/feedback/Dialog.tsx
```diff
@@ -68,31 +92,22 @@ const DialogNext = function Dialog({
...
-  });
-
-  useFocusAway(modalRef, closeHandler, {
-    enabled: closeOnFocusAway,
-  });
-
-  const dialogDescriptionId = useUniqueId('dialog-description');
-
-  useEffect(() => {
+  const [visible, setVisible] = useState(false);
```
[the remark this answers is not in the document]
lyzadanger: Might consider naming this something that clearly associates it with transitions. A Dialog is always "open" (it's stateless), so this naming of "visible" could be confusing to future devs, especially when there is no transition provided.

lyzadanger (commented):
As we chatted about with voices, this is moving in a good direction. Next steps/feedback includes:

* Add one or more examples on the `DialogPage` in the pattern library to show a transition (You might have to whip up a quick fake `TransitionComponent` locally to demo this in lieu of bringing `Slider` into the package, yet). Add some rudimentary documentation for the `transitionComponent` prop; we'll need to expand this once we start providing any `TransitionComponent` components from this package, but I don't want to blow up the scope of this work _too_ much.
* Provide the `onTransitionEnd` callback with an argument to identify which "direction" the transition has just completed (e.g. `'in' | 'out'` or similar). Then use that internally in `Dialog` to determine "what to do".
* Ensure that the `setInitialFocus` call in the `useEffect` hook in `Dialog` doesn't get invoked if there is a provided transition.
* Re-examine the logic for composing the close handling function internal to `Dialog`, just to make sure it's as intuitive as it can be for future devs to understand.
* Extract transition-component props API type to the `types` module and export from the `next` entry point.

We will approach bringing over `Slider` and documenting transitions as a separate chunk of work.

acelaya:
@lyzadanger I think everything has been addressed, so feel free to review again.

Two things that might still need some discussion:

* Think on how this "affects" the ModalDialog component.
* e-examine the logic for composing the close handling function internal to `Dialog`, just to make sure it's as intuitive as it can be for future devs to understand.

src/components/feedback/Dialog.tsx
```diff
@@ -33,6 +41,11 @@ type ComponentProps = {
    * Restore focus to previously-focused element when unmounted/closed
    */
   restoreFocus?: boolean;
+
+  /**
+   * Allows to provide a transition animation to the Dialog contents
```
lyzadanger, line 46:
```suggestion
   * Optional TransitionComponent for open (mount) and close (unmount) transitions
```

src/components/feedback/Dialog.tsx
```diff
@@ -53,46 +66,42 @@ const DialogNext = function Dialog({
...
-  });
-
-  const dialogDescriptionId = useUniqueId('dialog-description');
+  const closeHandler = useCallback(() => {
+    if (TransitionComponent) {
+      setTransitionComponentVisible(false);
+    } else {
+      onClose();
+    }
+  }, [onClose, TransitionComponent]);
```
lyzadanger, line 97: If you felt like it added useful context, you _could_ add a brief comment or two in this handler function.

src/components/feedback/Dialog.tsx
```diff
@@ -53,46 +66,42 @@ const DialogNext = function Dialog({
...
 
-  useEffect(() => {
+  const setInitialFocus = useCallback(() => {
```
lyzadanger, line 99: Although the name of this function is totally accurate, you could consider renaming it something more contextual, like `initializeDialog`. That would represent it as a sort of bag of operations that get executed once when the Dialog is mounted and "ready" (what "ready" means differs based on whether a `TransitionComponent` was provided) and might make the code a little easier to scan. OTOH, this really does just deal in initial focus, so perhaps the abstraction is not helpful...

src/components/feedback/Dialog.tsx
```diff
@@ -53,46 +66,42 @@ const DialogNext = function Dialog({
...
     if (initialFocus === 'manual') {
       return;
     }
     if (initialFocus === 'auto') {
-      // An explicit `initialFocus` has not be set, so use automatic focus
+      // An explicit `initialFocus` has not been set, so use automatic focus
```
lyzadanger, line 104: Hee hee, thanks.

src/components/feedback/Dialog.tsx
```diff
@@ -109,7 +118,41 @@ const DialogNext = function Dialog({
     } else {
       // Fall back to focusing the modal itself
       modalRef.current?.focus();
-      return;
+    }
+  }, [initialFocus, modalRef]);
+
+  const onTransitionEnd = (direction: 'in' | 'out') => {
+    if (direction === 'in') {
+      // We can't check the initial focus until transition "in" has finished
```
lyzadanger, line 126: See my above comment about perhaps renaming `setInitialFocus` to `initializeDialog` or similar. That would obviate the need for this comment.

src/components/feedback/Dialog.tsx
```diff
@@ -109,7 +118,41 @@ const DialogNext = function Dialog({
...
+  });
+
+  const dialogDescriptionId = useUniqueId('dialog-description');
+  const Wrapper = useMemo(
+    () => TransitionComponent ?? Fragment,
+    [TransitionComponent]
+  );
+
+  useEffect(() => {
+    // Trigger initial "in" transition animation
```
lyzadanger, line 152: Might be able to do without this comment.

src/pattern-library/components/patterns/FadeComponent.tsx
```diff
@@ -0,0 +1,31 @@
...
+    [visible, onTransitionEnd]
+  );
+
+  return (
+    <div
+      className={classnames('transition-opacity duration-500', {
+        'opacity-100': visible,
+        'opacity-0': !visible,
+      })}
+      // @ts-expect-error react uses "ontransitionend" rather than "onTransitionEnd".
```
lyzadanger, line 22:
```suggestion
      // @ts-expect-error preact uses "ontransitionend" rather than "onTransitionEnd".
```

src/pattern-library/components/patterns/FadeComponent.tsx
```diff
@@ -0,0 +1,31 @@
+import classnames from 'classnames';
+import { useCallback } from 'preact/hooks';
+
+import type { TransitionComponent } from '../../../types';
+
+const FadeComponent: TransitionComponent = ({
```
lyzadanger, line 6: Can you add a docstring/comment here to explain the purpose of this component (which is, as I understand it, to demonstrate the `TransitionComponent` functionality?)?

src/types.ts
```diff
@@ -25,3 +25,12 @@ export type BaseProps = PresentationalProps & { unstyled?: boolean };
  * valid `<svg>` element attribute as props.
  */
 export type IconComponent = FunctionComponent<JSX.SVGAttributes<SVGSVGElement>>;
+
+/**
+ * A component type used by other components that can optionally support a
+ * transition animation
+ */
```
lyzadanger, line 32: These components will still be usable as standalone components. e.g. 

```jsx
<Slider visible>
  <div>Some sort of content</div>
</Slider>
```

should totally work.

As such, perhaps:

```suggestion
/**
 * A component that describes an `in` and an `out` transition, typically to animate the 
 * mounting and unmounting of a child component.
 */
```

I dunno, this one is hard to write!

lyzadanger (approved):
Wow, this is coming along very nicely!

src/components/feedback/Dialog.tsx
acelaya: I think I'll just a short note on the first if, explaining "why" this is done like this.

src/components/feedback/Dialog.tsx
acelaya: I like the idea. It will quickly help realize that anything that needs to happen "at the beginning", should be there, or invoked from there.

Otherwise, if this component evolves, it's easy to end up putting more stuff in the existing `useEffect`, making it work properly only when there's no `TransitionComponent`.

I'll rename it.

src/pattern-library/components/patterns/FadeComponent.tsx
acelaya: Oops 😅

PR #958, opened 2023-04-13
Add `ModalDialog` example to demonstrate keyboard navigation

lyzadanger:
This PR adds a pattern-library example to the Dialogs pattern-library page (http://localhost:4001/feedback-dialog) to demonstrate keyboard navigation within a `ModalDialog` that contains complex content, in this case multiple ARIA widgets (a `tablist` and a `grid`).

While the tag-key navigation handling is working well for most modals, it breaks down when a modal dialog contains widgets which themselves have keyboard-navigation handling.

I'm adding this example as documentation of what the existing keyboard-navigation problems are. I have some fixes in mind that I've been prototyping. This extant example will then help us to verify when keyboard behavior is fixed. Sort of like committing a failing test.

Here is the current tab-navigation behavior in this example:

modal-keyboard-nav

What's happening here:

* Tab-key navigation _mostly_ works with the `tablist` tabs, but this is a sort of "accidental mostly works" because the tabs themselves are buttons, which get picked up by the modal's tab-key navigation hook automatically. What _should_ happen is that the tab sequence should select the `tablist` container (not the individual buttons) and then the tablist's own arrow-key navigation should take over within the tablist. This is a sort of subtle difference, but, implemented correctly, will correctly navigate to the last tab focused when returning to the tablist.
* Data-table navigation is completely not working. It should be possible to tab to the table, and then navigate the rows in the table with arrow keys. Alas, this is not as simple as adding `DataTable` elements (`[role="grid"]`) to the `useTabKeyNavigation` hook's selector, i.e. we can't just add them to the tab sequence. It's more fiddly than that. The good news is that I think I do have a workable solution. Similar to with the `tablist`, the behavior should be: the modal's tab sequence includes the `DataTable` container, and then the user can navigate rows with arrow keys. Tabbing out of the table, then tabbing back and using arrow keys again should navigate based on the last row selected (not starting from the top row).

i.e. in both widget cases, "tabbing back" to the widgets after tabbing out of them should retain state about the last widget sub-item that was active, not reset the navigation sequence in the widget. THIS IS REALLY HARD TO EXPLAIN, especially this late in the day :). It may be easier to demonstrate when I push a proposed fix.

### DIFF
--- a/src/pattern-library/components/patterns/feedback/DialogPage.tsx
+++ b/src/pattern-library/components/patterns/feedback/DialogPage.tsx
@@ -15,9 +15,12 @@ import {
   InputGroup,
   Link,
   Scroll,
+  Tab,
+  TabList,
 } from '../../../../next';
 import Library from '../../Library';
 import { LoremIpsum, nabokovNovels } from '../samples';
+import type { NabokovNovel } from '../samples';
 
 const nabokovRows = nabokovNovels();
 const nabokovColumns = [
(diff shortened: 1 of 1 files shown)


acelaya (approved):
👍🏼

PR #959, opened 2023-04-14
Support keyboard navigation within and across nested widgets

lyzadanger:
This PR introduces changes to the `useArrowKeyNavigation` and `useTabKeyNavigation` hooks such that they behave more concordantly when multiple keyboard-navigable widgets are used and nested.

See the video in https://github.com/hypothesis/frontend-shared/pull/958 for the before state: tab and arrow navigation were not melding well in `ModalDialog`s. It was impossible to reach the `DataTable` by tabbing with the keyboard.

Behavior after these changes:

modal-keyboard-nav-fixed

What's going on here:

* `ModalDialog` opens
* `Tab` forward to the "link". Note that as I tab to the `tabList`, the tablist container is focused (and the tabs themselves are, by design, not in the tab sequence).
* `Shift-tab` back to the tablist, then using arrow keys to navigate to `Tab2`.
* `Tab` to "link" then `Shift-Tab`: Notice that `Tab2` is focused, not the container now: The state of that widget's focus is retained.
* `Tab` forward: The `DataTable` container is focused. `Tab` again to "Maybe" button, then `Shift-tab` back to the table.
* Now using arrow keys to select rows in the `DataTable`. 
* Tabbing out of the `DataTable` and then back again resumes navigation within the table based on the last selected row.

In summary:

* Initial tab sequence in a `ModalDialog` (well, anything using the tab-key hook) will include the _containers_ of `TabList` and `DataTable` components. 
* Once a widget with arrow-key support is "dropped into" (user starts navigating with arrow keys), the focused item is retained in state.
* Exiting and re-entering a widget will resume navigation from the last-focused/selected item

_Note_: We haven't had need to use tabs in a modal yet, but we definitely do use `DataTable`s. 

This PR also adds some styling for focus rings for consistency in these widgets.

This. was. so. hard. And fiddly. It is my hope that this is the final tweak needed to be able to ship the updated `ModalDialog` component so we can finally roll it out.


src/hooks/use-arrow-key-navigation.ts
```diff
@@ -82,9 +87,17 @@ export function useArrowKeyNavigation(
...
         container.querySelectorAll(selector)
       );
-      return elements.filter(
+      const filtered = elements.filter(
         el => isElementVisible(el) && !isElementDisabled(el)
       );
+      // Include the container itself in the set of navigable elements if it
+      // is currently focused. It will not be part of the tab sequence once it
+      // loses focus. This allows, e.g., a modal container to be focused when
+      // opened but not be part of the subsequent trapped tab sequence.
```
lyzadanger, line 96: This comment was directly pasted from `useTabKeyNavigation` and isn't quite applicable. I'll adjust after review.

acelaya (approved):
It looks like this one was "fun" to figure out 😅

PR #968, opened 2023-04-17
Version 6: Remove legacy assets, eliminate `Next` naming

lyzadanger:
This PR completes the tasks under "Remove legacy assets" in https://github.com/hypothesis/frontend-shared/issues/876, a large chunk of getting to version 6 of this package.

This is a large diff, but primarily removal. It:

* removes all legacy components, SASS, icons, icon registration and pattern-library documentation of same
* removes the `Next` name suffix from components and templates that generate components
* makes the `next` and `index` entrypoints identical for now and updates imports to use the "regular" (`index`) entrypoint instead of `next`
* Updates pattern-library and README documentation to remove references to legacy components, SASS, "next", etc.

Note that this PR wants to merge into a feature branch (`version-6`), not `main`. As this branch's changes touch basically every module and file in this project, I'll need to keep it aggressively rebased on `main` until `version-6` is merged.

The other piece of version 6 work is the tasks under "Package Housekeeping" in https://github.com/hypothesis/frontend-shared/issues/876 and some testing.

TODOs before landing and release:

* [x] test build against `client` application
* [x] test build against `lms` application


acelaya (approved):
Good one! 💪🏼 

Note for the future. Since we no longer use the `Next` suffix in "new" components, perhaps we could at some point move from this:

```tsx
const SomeComponent = function SomeComponent() {
  // [...]
};

export default SomeComponent;
```

To simply do this:

```tsx
export default function SomeComponent() {
  // [...]
}
```

That of course assuming there's not other reason that I'm missing, and not on this PR, as it should be backwards compoatible.

acelaya:
> Ah, yes! I thought about this while I was making changes. Basically, I didn't do it now simply because the diff was already huge, and it wasn't critical for this release. Absolutely valid to do it in a follow-up pass.

Yep! I guessed, that was the reason, and it totally makes sense.

PR #969, opened 2023-04-18
Use `wouter` for routing

### LINKED ISSUE #557: Fragment navigation within pattern-library page not working
At present, clicking on a link in a pattern-library page to a fragment within the same page does not work. The URL bar is updated, but the viewport does not scroll to the fragment location. Entering the same URL (with hash) directly works.

There's something about the `navigate` callback in the `router` that isn't working for this; I haven't investigated it. If I do something hamfisted to make the `clickListener` return without trying to handle navigating the link (i.e. `return` before calling `navigate`), it works.

lyzadanger:
This PR replaces the custom routing we've been using in the pattern library with `wouter-preact`, which we also use in the `lms` project. By using `wouter`, I was also able to find a solution for fragment navigation. It is now possible to link between pattern-library pages, and to use links with hashes. The API for passing routes to the app has not changed (so, no breaking changes here).

There's plenty of potential follow-up work here: establishing IDs (for hash-linking) on library heading elements, providing tables of contents for pages, making the new `<Library.Link>` component support external links. But all for separate PRs!

## Testing

Run `make dev` on this branch and visit the [pattern library](http://localhost:4001/). Ensure that:

* Left navigation items work, and show "active" state
* The page scrolls to the top when you navigate (you might not see the scroll, but the page should be at the top and not scrolled down).
* Visit [the page for the deprecated Modal component](http://localhost:4001/feedback-modal) and ensure that the two links under the "Status" section (which contain fragments) navigate and scroll to the right place on the Dialogs page.

Fixes #557


src/pattern-library/components/Library.tsx
```diff
@@ -48,7 +49,10 @@ export type LibraryPageProps = {
 function Page({ children, intro, title }: LibraryPageProps) {
   return (
     <section className="max-w-6xl pb-16 space-y-8 text-slate-7">
-      <div className="sticky top-0 z-4 h-16 flex items-center bg-slate-0 border-b">
+      <div
+        className="sticky top-0 z-4 h-16 flex items-center bg-slate-0 border-b"
+        id="page-header"
```
lyzadanger, line 54: This is the page's sticky header. Just giving it an ID so I can query for it in the `PlaygroundApp`'s navigation/scroll handling `useEffect` hook.

src/pattern-library/components/Library.tsx
```diff
@@ -365,12 +369,30 @@ function Usage({ componentName, size = 'md' }: LibraryUsageProps) {
...
+ * Render an internal link to another pattern-library page.
+ * TODO: Support external links
+ */
+function Link({ children, href }: LinkProps) {
+  return (
+    <RouteLink href={href}>
+      <UILink underline="always">{children}</UILink>
+    </RouteLink>
+  );
+}
```
lyzadanger, line 387: So many things named `Link`, heh.

src/pattern-library/components/PlaygroundApp.tsx
```diff
@@ -1,13 +1,85 @@
...
+} from 'wouter-preact';
 
 import type { PlaygroundAppProps } from '../';
 import { Link, LogoIcon } from '../../';
-import { useRoute } from '../router';
 import { componentGroups, getRoutes } from '../routes';
 import type { PlaygroundRoute } from '../routes';
 import Library from './Library';
 
+/**
```
lyzadanger, line 19: No meaningful changes to `NavHeader`, `NavSection` or `NavList`: just extracting them so that they are not inside of the `PlaygroundApp` component function.

src/pattern-library/components/PlaygroundApp.tsx
```diff
@@ -19,130 +91,125 @@ export default function PlaygroundApp({
...
-              <Link href={baseURL + '/'} classes="grow flex gap-x-2">
-                <LogoIcon />
-                Component Library
-              </Link>
-            </h1>
-          </div>
-          <nav id="nav" className="pt-8 pb-16 space-y-4 mr-4">
-            <NavHeader>Foundations</NavHeader>
-            <NavList routes={getRoutes('foundations')} />
+    <Router base={baseURL}>
```
lyzadanger, line 152: NB: Except for wrapping the whole thing in a `Router`, there are no `wouter`-specific changes here.

src/pattern-library/components/PlaygroundApp.tsx
```diff
@@ -19,130 +91,125 @@ export default function PlaygroundApp({
...
+          </div>
+          <div className="bg-white pb-16">
+            <Switch>
+              {pageRoutes}
+              <Route>
+                <Library.Page title=":( Sorry">
+                  <h1 className="text-2xl">Page not found</h1>
+                </Library.Page>
+              </Route>
+            </Switch>
```
lyzadanger, line 209: Here's where the component routing actually happens.

src/pattern-library/routes.ts
```diff
@@ -53,7 +53,7 @@ export type PlaygroundRoute = {
    * not provided, a placeholder entry for this route will be added to the
    * navigation, with no link.
    */
-  route?: RegExp | string;
+  route?: string;
```
lyzadanger, line 56: Adjusted to work with `wouter` types. We only used `RegExp` in one route, which isn't needed anymore.

src/pattern-library/components/PlaygroundApp.tsx
lyzadanger, line 209: `<Switch>` is a `wouter` component that will render (only) the first matching `<Route>` inside of it.

src/pattern-library/components/Library.tsx
acelaya: We should call this component `LinkLink`

PR #971, opened 2023-04-20
Migrate type definition example from jsdoc to TypeScript

acelaya:
I saw this jsdoc example, and I supposed it's there only for historical reasons.

### DIFF
--- a/src/pattern-library/components/patterns/UsingComponentsPage.tsx
+++ b/src/pattern-library/components/patterns/UsingComponentsPage.tsx
@@ -64,17 +64,15 @@ export default function UsingComponentsPage() {
           </p>
           <Library.Code
             size="sm"
-            content={`/**
- * @typedef PresentationalProps
- * @prop {import('preact').ComponentChildren} [children]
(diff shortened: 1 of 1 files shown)


src/pattern-library/components/patterns/UsingComponentsPage.tsx
```diff
@@ -64,17 +64,15 @@ export default function UsingComponentsPage() {
...
-`}
+            content={`export type PresentationalProps = {
+  children?: ComponentChildren;
+  /** Optional extra CSS classes appended to the component's className */
+  classes?: string | string[];
+  /** Use \`classes\` or props instead */
+  className?: never;
+  /** Ref associated with components outermost or primary element */
+  elementRef?: Ref<HTMLElement | undefined>;
+};`}
```
acelaya, line 75: I have used the type as defined in the `types` module, but the comments for every field are different. I'm not sure if it's because those here were outdated, or it was intended.

lyzadanger (approved):
Thank yoU!

PR #974, opened 2023-04-20
Update document title when the route changes

### LINKED ISSUE #972: Fix page titles in pattern library
Ah, we lost useful (HTML) page titles in the pattern library with the migration to `wouter` routing. This is probably not too hard to fix.

acelaya:
This PR recovers the logic to dynamically set the document title based on the active route.

Fixes #972


lyzadanger (approved):
One of the fun things about working in the pattern library is that it's a small app and non-production-critical, so it can feel good to move fast and figure things out quickly and just get it done! Thank you! I'll merge this.

Issue #966, opened 2023-04-17
Add `Slider` transition to library; document `TransitionComponent`s


lyzadanger:
The `Dialog` component currently takes a `transitionComponent: TransitionComponent` prop, and the API for `TransitionComponent` is defined in exported types. The intent is that relevant components can take a `TransitionComponent`, and that this package should define, provide and document relevant transition components. As such:

* Bring the `Slider` component (in the `client`) into the package
* Document transition components in the pattern library (this may involve adding a new navigation section) and update `Dialog`'s pattern-library documentation if needed

acelaya:
I'm thinking `transition` might be a possible group name for these new components (being `Slider` the first one).

That's how they call it in libraries like [Material UI](https://mui.com/material-ui/transitions/).

Other libraries I've checked don't have nothing specific for "transitions".

PR #975, opened 2023-04-21
Slider component

acelaya:
> This PR is part of https://github.com/hypothesis/frontend-shared/issues/966

This PR brings the `Slider` component from [client](https://github.com/hypothesis/client/blob/6f3a741d630375c9dfd780af3a66991c92027c7e/src/sidebar/components/Slider.tsx) to this library.

This component will be the first one fulfiling the `TransitionComponent` contract. These components can be passed to any other component supporting them, or used standalone.

I have also deleted the dummy `FadeComponent` that was created just for `transitionComponent` documentation purposes, and used the `Slider` instead.

### TODO

- [x] We still need to decide on the final group name for `TransitionComponent`s (as in `feedback`, `input`, `layout`, etc). I have provisionally used `transition`.
- [x] We also need to document the `Slider` specifically and `TransitionComponents` in general in the pattern library. -> [On a follow-up PR](https://github.com/hypothesis/frontend-shared/pull/981) to avoid a too big diff

### Testing steps

The slider can be seen in action in the [Dialog page](http://localhost:4001/feedback-dialog). Once there, search for `transitionComponent` and there's an example using the `Slider` component.


src/components/transition/Slider.tsx
```diff
@@ -0,0 +1,95 @@
+import { useCallback, useEffect, useRef, useState } from 'preact/hooks';
+
+import type { TransitionComponent } from '../../types';
+
+const Slider: TransitionComponent = ({
```
acelaya, line 5: I would like to be able to export this as `export default function Slider() {...}`, but then there's no way to type it as `TransitionComponent`.

It would be nice to be able to do something like `export default function Slider() {...} satisfies TransitionComponent` 😅

src/components/transition/Slider.tsx
```diff
@@ -0,0 +1,95 @@
...
+        // focus rings, which may extend outside the bounds of the Slider content,
+        // are visible.
+        overflow: isFullyVisible ? 'visible' : 'hidden',
+        transition: `height 0.15s ease-in`,
+      }}
+    >
+      {children}
+    </div>
+  );
+};
```
acelaya, line 93: This component is brought from [client](https://github.com/hypothesis/client/blob/6f3a741d630375c9dfd780af3a66991c92027c7e/src/sidebar/components/Slider.tsx) verbatim.

lyzadanger: Pro tip! GitHub now has the ability to add a comment at a module level, if that's helpful (the comment box icon at the top of each file's diff):

image

src/pattern-library/components/patterns/feedback/DialogPage.tsx
```diff
@@ -13,13 +13,13 @@ import {
...
   Link,
   Scroll,
+  Slider,
   Tab,
   TabList,
 } from '../../../../';
 import { ModalDialog } from '../../../../components/feedback';
 import type { ModalDialogProps } from '../../../../components/feedback/ModalDialog';
 import Library from '../../Library';
-import FadeComponent from '../FadeComponent';
```
lyzadanger, line 22: Nice work on cleaning up.

lyzadanger (commented):
Looking great so far! I think `transition` is a fine name for this group of components. Presumably the human-readable section in the pattern library would be `Transitions`?

Do you want to take on the challenge of adding a basic pattern library page for this component (hint: you'll need to add a new "component group" to the routes typing, but I bet you're able to figure that out pretty quick)? I can help with the details once it's in place. We can tackle documenting `TransitionComponent` as a concept as a separate task (i.e. probably would document it on the "Using Components" page, which needs content updates anyway).

src/components/transition/Slider.tsx
acelaya: I actually knew that! 😅

acelaya:
I'm putting this PR as ready for review, and will work on the documentation in parallel.

src/index.ts
lyzadanger, line 1: I think it's OK to only update `index` (not `next`, which is going to go away in an upcoming minor release).

lyzadanger (approved):
Yay!

Follow-up if you're up to the challenge: Update the `plopfile.js` and add a new template for `transition` components, such that we can run `plop` to scaffold a new transition component in the future.

acelaya:
> Follow-up if you're up to the challenge: Update the plopfile.js and add a new template for transition components, such that we can run plop to scaffold a new transition component in the future.

On my way

PR #977, opened 2023-04-21
Add prototype section and page for re-use annotations features

lyzadanger:
This PR:

* Adds a navigation section, route group and place to put pattern-library pages for prototyping UI stuff
* Brings over a bunch of client `Menu` component stuff (more on this below)
* Adds a couple of icons useful for the prototyping examples
* Adds a page for prototyping some wireframe/low-fidelity UI sketches for the re-use annotations projects

Don't panic at the diff size. Most of this is on account of my bringing over `Menu`, `MenuItem`, `MenuArrow` from the client (and putting them in the prototype area as utility components — this is not a PR to actually add those components to this package). Alas, the current structure of the `AnnotationPublishControl` is such that it has dependencies on all of these. I agonized a while before bringing these over. But I needed to get some sketch work done, and I needed the annotation-publish interface to be reasonably believable.

Thus, this PR is sort of three chunks:

1. Add a structure for prototyping in the pattern library: a directory, route, route group
2. Add some components temporarily (in the prototyping area) to support a specific UI prototyping need
3. Adding the prototyping page itself and supporting assets

The prototype page itself is loose and is the basis for some discussions next week.

Detail of prototype page:

image


src/pattern-library/components/patterns/prototype/FakeAnnotationPublishControl.tsx
lyzadanger, line 1: This is a simplified, shortened and adapted variant of the client's `AnnotationPublishControl` for UI-sketching purposes

src/pattern-library/components/patterns/prototype/Menu.tsx
lyzadanger, line 1: This should not require review; it is only used for UI sketching purposes. It's brought over from the client and lightly adapted (primarily imports).

src/pattern-library/components/patterns/prototype/MenuArrow.tsx
lyzadanger, line 1: Also no need for review on this one.

src/pattern-library/components/patterns/prototype/.gitkeep
lyzadanger, line 1: This will ensure that the `prototype` menu sticks around if we remove any pages inside of it.

acelaya (approved):
👍🏼

PR #978, opened 2023-04-21
Use "getting started" page as home page for pattern library

lyzadanger:
For a long time, we've maintained a fairly useless "landing page" in the pattern library because of routing constraints. This is easier to wrangle now. The "Getting started" page makes more sense as a landing page.


acelaya (approved):
👍🏼

PR #979, opened 2023-04-21
Update `Library` section component structure

lyzadanger:
OK, this is the last fun/housekeeping PR in this package as we head into a more product-focused sprint/cycle/skirmish/whatever you want to call it week next week.

One of the other pieces of feedback I've heard about the pattern library over time is that it's kind of fiddly and hard to remember which "section"-like component to use on a pattern-library page: `<Section>`, `<Pattern>` or `<Example>` and which goes where.

This PR refactors and consolidates the `Library` and components such that there are additional available page-sectioning options:

1. The previous/current API: You can still use `Section`, `Pattern` and `Example` as before with no change to behavior/design. Internally, `Pattern` and `Example` now wrap `Section` and a new `Heading` component.
2. Use only `Section` elements. You can then either:
    * Provide an explicit `level` (`number`) prop to dictate which level the section is (1, 2, 3, 4...), or
    * Don't provide a `level` prop and it will render a section based on the current nesting depth (I used a `context` for this)

I updated the reuse-annotations prototype page to show both variations of the `Section`-only API.

This work also normalizes the components to behave more like our package components (merging in HTML attributes, e.g.).

Depends on #978 because of a change I made in that PR branch to the `Section` component's `title` prop (d'oh).

I'd like us to do a bit more page authorship before we decide what combination feels best. For example, using the `Example` component sometimes feels intuitive, while `Pattern` was never more than just a heading/section level.


src/pattern-library/components/Library.tsx
```diff
@@ -10,33 +10,10 @@ import { jsxToHTML } from '../util/jsx-to-string';
...
- *         // More Demos if desired...
- *       </Library.Example>
- *       // More Examples if desired...
- *     </Library.Pattern>
- *   // more Patterns if desired...
- *   </Library.Section>
- *
- *   // More Sections...
- *
- * </Library.Page>
```
lyzadanger, line 37: Now that it could be as simple as

```
<Page>
  <Section>
    <Section />
  </Section>
</Page>
```

I'm removing this long-winded example. That doesn't mean that we have to kill all documentation on this, I'm just not sure what would be most useful.

src/pattern-library/components/Library.tsx
```diff
@@ -68,78 +45,115 @@ function Page({ children, intro, title }: LibraryPageProps) {
...
-function Section({ children, id, intro, title }: LibrarySectionProps) {
+function Section({
+  children,
+  intro,
+  level,
+  title,
+  ...htmlAttributes
+}: LibrarySectionProps) {
+  const sectionDepth = useContext(SectionDepthContext);
+  const depth = level ?? sectionDepth + 1;
```
lyzadanger, line 104: So: If provided with a `level` prop it will render at that level, otherwise it will automatically set depth based on parent depth.

Note to self, mostly: If you provide a `level`, that will effectively reset the depth to that `level` for descendants, but that's probably desirable anyway...

src/pattern-library/components/patterns/prototype/SharedAnnotationsPage.tsx
```diff
@@ -159,6 +159,7 @@ export default function SharedAnnotationsPrototypePage() {
             </p>
           </>
         }
+        level={3}
```
lyzadanger, line 162: In this case, the automatic level would be `2` but I wanted a 3rd-level section.

src/pattern-library/components/patterns/prototype/SharedAnnotationsPage.tsx
```diff
@@ -240,8 +241,8 @@ export default function SharedAnnotationsPrototypePage() {
                 <FakeAnnotation isOwn />
               </FakeSidebar>
             </Library.Demo>
-          </Library.Example>
-        </Library.Pattern>
+          </Library.Section>
+        </Library.Section>
```
lyzadanger, line 245: I'm not entirely convinced that nested `Section` components will be any less fiddly to work with than section-components-with-unique-names, but we'll see.

PR #981, opened 2023-04-24
Slider component docs

acelaya:
> This PR is based on https://github.com/hypothesis/frontend-shared/pull/975 and is part of https://github.com/hypothesis/frontend-shared/issues/966

This adds some documentation to the pattern library, around the new `Slider` component and `TransitionComponent`s in general.

### Slider

New page, which also requires a new **Transitions** route group: http://localhost:4001/transitions-slider

image

### TransitionComponent

Added as a new sub-section in **Using components**: http://localhost:4001/using-components

Also modified the `Dialog` page, removing the `TransitionComponentProps` definition from there, and just referencing `TransitionComponents` instead.

image


src/pattern-library/components/patterns/UsingComponentsPage.tsx
```diff
@@ -166,6 +171,37 @@ export default function UsingComponentsPage() {
...
+            <p>
+              Example: <code>Slider</code>.
+            </p>
+          </>
+        }
+      >
+        <Library.Pattern title="Props API">
+          <p>
+            <strong>Transition components</strong> take these common props
+            unless documented otherwise:
```
acelaya, line 192: I'm not completely sure if the `unless documented otherwise` part applies here. I left it because it is never technically wrong, but I have mixed feelings.

lyzadanger (approved):
This looks great!

PR #992, opened 2023-04-24
Add support to create transition components via plop

acelaya:
In order to test this works, follow these steps:

* Run `yarn plop`, and introduce a component name (let's say `Fade`), then input `transition` as the group directory, and select `transition` as category.
  image
* Make sure generated code is valid by running `make lint`.
* Make sure generated test passes: Assuming the component is called `Fade`, run `yarn test --grep Fade`.

### DIFF
--- a/plop-templates/transition-component-test.hbs
+++ b/plop-templates/transition-component-test.hbs
@@ -9,7 +9,7 @@ describe('{{name}}', () => {
   const create{{name}} = (props = {}) => {
     const style = { width: 100, height: 200 };
     return mount(
-      <{{name}} visible={false} {...props}>
+      <{{name}} {...props}>
         <div style={style}>Test content</div>
       </{{name}}>,
       { attachTo: container }
@@ -29,12 +29,12 @@ describe('{{name}}', () => {
     'should pass a11y checks',
     checkAccessibility([
       {
-        name: 'visible',
-        content: () => create{{name}}({ visible: true }),
+        name: 'in',
+        content: () => create{{name}}({ direction: 'in' }),
       },
       {
-        name: 'hidden',
-        content: () => create{{name}}({ visible: false }),
+        name: 'out',
+        content: () => create{{name}}({ direction: 'out' }),
       },
     ])
   );
--- a/plop-templates/transition-component-test.hbs
+++ b/plop-templates/transition-component-test.hbs
@@ -1,0 +1,41 @@
+import { mount } from 'enzyme';
+
+import { checkAccessibility } from '../../../../test/util/accessibility';
+import {{name}} from '../{{name}}';
+
+describe('{{name}}', () => {
+  let container;
+
+  const create{{name}} = (props = {}) => {
+    const style = { width: 100, height: 200 };
+    return mount(
+      <{{name}} visible={false} {...props}>
+        <div style={style}>Test content</div>
+      </{{name}}>,
+      { attachTo: container }
+    );
+  };
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    document.body.appendChild(container);
+  });
+
+  afterEach(() => {
+    container.remove();
+  });
+
+  it(
+    'should pass a11y checks',
+    checkAccessibility([
+      {
+        name: 'visible',
+        content: () => create{{name}}({ visible: true }),
+      },
+      {
+        name: 'hidden',
+        content: () => create{{name}}({ visible: false }),
+      },
+    ])
+  );
+});
--- a/plop-templates/pattern-library-page.hbs
+++ b/plop-templates/pattern-library-page.hbs
@@ -19,11 +19,9 @@ export default function {{name}}Page() {
         </Library.Pattern>
 
         <Library.Pattern title="Props">
-          <Library.Example title="size">
+          <Library.Example title="exampleProp">
             <Library.Demo withSource>
-              <{{name}} size="sm">Small (sm)</{{name}}>
-              <{{name}} size="md">Medium (md) (default)</{{name}}>
-              <{{name}} size="lg">Large (lg)</{{name}}>
+              <p>TODO: Example of component with prop</p>
             </Library.Demo>
           </Library.Example>
         </Library.Pattern>
--- a/plop-templates/transition-component.hbs
+++ b/plop-templates/transition-component.hbs
@@ -1,0 +1,36 @@
+import { useCallback } from 'preact/hooks';
+
+import type { TransitionComponent } from '../../types';
+
+/**
+ * TODO
+ */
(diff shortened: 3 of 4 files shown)


plop-templates/transition-component-test.hbs
```diff
@@ -0,0 +1,41 @@
...
+import {{name}} from '../{{name}}';
+
+describe('{{name}}', () => {
+  let container;
+
+  const create{{name}} = (props = {}) => {
+    const style = { width: 100, height: 200 };
+    return mount(
+      <{{name}} visible={false} {...props}>
+        <div style={style}>Test content</div>
```
acelaya, line 13: We can't inline the `style` object, because then there's nested curly braces (`style={{ width: 100, height: 200 }}`), making plop treat it as a template, and failing to parse it.

lyzadanger: In the first round of plop templates, we were still using JSDoc typing, so I had to figure this out. This appears to be a workaround.

```
 * {{!See https://github.com/handlebars-lang/handlebars.js/issues/1404 }}
 * @param { {{~name~}} Props} props
```

lyzadanger (changes requested):
If you opt into a pattern-library page for the new component, `typecheck` fails (details below). 

This is more the pattern-library plop template's fault than yours! 

```
$ make lint
yarn run lint
yarn run v1.22.19
$ eslint .
✨  Done in 2.74s.
yarn run typecheck
yarn run v1.22.19
$ tsc --build src/tsconfig.json
src/pattern-library/components/patterns/transition/FadePage.tsx:16:16 - error TS2322: Type '{ children: string; }' is not assignable to type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.
  Property 'visible' is missing in type '{ children: string; }' but required in type '{ visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; }'.

16               <Fade>Example of Fade</Fade>
                  ~~~~

  src/types.ts:34:3
    34   visible: boolean;
         ~~~~~~~
    'visible' is declared here.

src/pattern-library/components/patterns/transition/FadePage.tsx:24:21 - error TS2322: Type '{ children: string; size: string; }' is not assignable to type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.
  Property 'size' does not exist on type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.

24               <Fade size="sm">Small (sm)</Fade>
                       ~~~~

src/pattern-library/components/patterns/transition/FadePage.tsx:25:21 - error TS2322: Type '{ children: string; size: string; }' is not assignable to type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.
  Property 'size' does not exist on type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.

25               <Fade size="md">Medium (md) (default)</Fade>
                       ~~~~

src/pattern-library/components/patterns/transition/FadePage.tsx:26:21 - error TS2322: Type '{ children: string; size: string; }' is not assignable to type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.
  Property 'size' does not exist on type 'IntrinsicAttributes & { visible: boolean; onTransitionEnd?: ((direction: "in" | "out") => void) | undefined; } & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<...> | undefined; }>'.

26               <Fade size="lg">Large (lg)</Fade>
                       ~~~~


Found 4 errors.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [lint] Error 1
```

You could change the pattern-library template to something like this patch:

```diff
diff --git a/plop-templates/pattern-library-page.hbs b/plop-templates/pattern-library-page.hbs
index 441fdbe..78bc535 100644
--- a/plop-templates/pattern-library-page.hbs
+++ b/plop-templates/pattern-library-page.hbs
@@ -19,11 +19,9 @@ export default function {{name}}Page() {
         </Library.Pattern>

         <Library.Pattern title="Props">
-          <Library.Example title="size">
+          <Library.Example title="exampleProp">
             <Library.Demo withSource>
-              <{{name}} size="sm">Small (sm)</{{name}}>
-              <{{name}} size="md">Medium (md) (default)</{{name}}>
-              <{{name}} size="lg">Large (lg)</{{name}}>
+              <p>TODO: Example of component with prop</p>
             </Library.Demo>
           </Library.Example>
         </Library.Pattern>
```

acelaya:
> If you opt into a pattern-library page for the new component, `typecheck` fails

True! I actually did all my tests skipping the page creation. My bad.

acelaya:
I actually realized that even with that change, the generated page does not pass type checks, because the transition components have a mandatory `visible` property.

I don't think we can define this template in a way that it's valid for any possible component right away (at least not in a simple way), so perhaps we should expect that the person generating it makes the required changes manually.

lyzadanger:
> I actually realized that even with that change, the generated page does not pass type checks, because the transition components have a mandatory `visible` property.
> 
> I don't think we can define this template in a way that it's valid for any possible component right away (at least not in a simple way), so perhaps we should expect that the person generating it makes the required changes manually.

Any chance that `visible` prop could default to `true`? Or would that be dangerous?

lyzadanger (approved):
Thanks for going the extra mile, errr, kilometer, whatever :)

acelaya:
> Any chance that `visible` prop could default to `true`? Or would that be dangerous?

Yeah, I think we can do that safely.

acelaya:
With https://github.com/hypothesis/frontend-shared/pull/996 merged, and after doing a couple changes needed by that, both `make lint` and `make test` pass after creating a new transition component, including the pattern library page, so I'm merging this.

I will now create a PR extracting common tests, and probably making some changes to the test plop template.

PR #996, opened 2023-04-26
Enhance transition components

acelaya:
While creating the `plop` files for `TransitionComponent`s, we realized there was a design mistake on their properties. The `visible` one should be optional, and default to `true`.

However, this presented another problem, based on how we have decided to use boolean optional props. We want them to have a default value which allows to set them with no value in order to overwrite it.

If `visible` defaults to `true`, we would not be able to follow that rule, because it would have to be set as `<Component visible={false} />` in order to overwrite it.

Because of this, we considered replacing the prop with something else. The first logical option was `hidden`, which could default to `false` and be overwritten with just `<Component hidden />`, but while implementing it, it didn't feel right, and most of the logic started with `const visible = !hidden`, to be able to continue dealing with `visible` as a concept.

After some discussions, we have moved away from a flag, and I have finally introduced an optional `direction` prop, which can have values `in` and `out`, with a default value of `in`.

The benefits of this approach are:

* It matches nicely with the param provided to `onTransitionEnd`, making it easier to handle internally, and to reason about.
* It feels right when implementing it (which didn't happen with `hidden`).
* It removes the problem of flag props with default value.

It has one main consideration, though:

Technically speaking, it is a breaking change.

In `client`, where we are using `Dialog` and providing a local component as `transitionComponent`, that local component will no longer fulfill the proper API.

However, there's already a [draft PR](https://github.com/hypothesis/client/pull/5424/files) which replaces that component by the equivalent one on this library, which has been adapted as part of this PR, so the migration should be easy.

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -196,7 +196,7 @@ const Dialog = function Dialog({
 
   return (
     <Wrapper
-      visible={transitionComponentVisible}
+      direction={transitionComponentVisible ? 'in' : 'out'}
       onTransitionEnd={onTransitionEnd}
     >
       <div
--- a/src/pattern-library/components/patterns/UsingComponentsPage.tsx
+++ b/src/pattern-library/components/patterns/UsingComponentsPage.tsx
@@ -194,7 +194,7 @@ export default function UsingComponentsPage() {
           <Library.Code
             size="sm"
             content={`type TransitionComponentProps = {
-  visible: boolean;
+  direction?: 'in' | 'out';
   onTransitionEnd?: (direction: 'in' | 'out') => void;
 };`}
             title="Common transition-component props"
--- a/src/pattern-library/components/patterns/UsingComponentsPage.tsx
+++ b/src/pattern-library/components/patterns/UsingComponentsPage.tsx
@@ -194,7 +194,13 @@ export default function UsingComponentsPage() {
           <Library.Code
             size="sm"
             content={`type TransitionComponentProps = {
+  /** Sets current direction of the component. Defaults to "in" */
   direction?: 'in' | 'out';
+
+  /**
+   * If provided, it is invoked when the component direction changes,
+   * once the transition has finished.
+   */
   onTransitionEnd?: (direction: 'in' | 'out') => void;
 };`}
             title="Common transition-component props"
--- a/src/pattern-library/components/patterns/transition/SliderPage.tsx
+++ b/src/pattern-library/components/patterns/transition/SliderPage.tsx
@@ -78,9 +78,8 @@ export default function SliderPage() {
 
         <Library.Pattern title="Props">
           <Library.Example title="direction">
-            This prop tells if the Slider is currently displayed (
-            <code>in</code>) or hidden (<code>out</code>). It is <code>in</code>{' '}
-            by default.
+            This prop tells if the Slider is currently expanded (<code>in</code>
+            ) or collapsed (<code>out</code>). It is <code>in</code> by default.
           </Library.Example>
 
           <Library.Example title="onTransitionEnd">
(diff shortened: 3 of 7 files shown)


src/types.ts
```diff
@@ -31,6 +31,6 @@ export type IconComponent = FunctionComponent<JSX.SVGAttributes<SVGSVGElement>>;
  * animate the mounting and unmounting of a child component.
  */
 export type TransitionComponent = FunctionComponent<{
-  visible: boolean;
+  direction?: 'in' | 'out';
```
lyzadanger, line 34: Does this mean that all `TransitionComponent`s have to support both an `in` and an `out` transition? Is this an OK assumption to make?

src/pattern-library/components/patterns/UsingComponentsPage.tsx
```diff
@@ -194,7 +194,7 @@ export default function UsingComponentsPage() {
           <Library.Code
             size="sm"
             content={`type TransitionComponentProps = {
-  visible: boolean;
+  direction?: 'in' | 'out';
```
lyzadanger, line 197: Can we document that default is `in`?

src/components/feedback/Dialog.tsx
```diff
@@ -196,7 +196,7 @@ const Dialog = function Dialog({
 
   return (
     <Wrapper
-      visible={transitionComponentVisible}
+      direction={transitionComponentVisible ? 'in' : 'out'}
```
lyzadanger, line 199: I'm curious: this `Wrapper` may be a reference to a `TransitionComponent` functional component or it might be a `Fragment`. Do `Fragment`s happily take any props without grumbling? Presumably typechecking would bark if not...I just haven't thought about it before.

src/pattern-library/components/patterns/transition/SliderPage.tsx
```diff
@@ -7,16 +7,17 @@ import Library from '../../Library';
 
 const Slider_: FunctionComponent<
   ComponentProps<TransitionComponent> & { _transitionStatus?: 'in' | 'out' }
-> = ({ children, visible, onTransitionEnd, _transitionStatus }) => {
-  const [isVisible, setIsVisible] = useState(visible);
-  const toggleSlider = () => setIsVisible(prev => !prev);
+> = ({ children, direction, onTransitionEnd, _transitionStatus }) => {
```
lyzadanger, line 10: This typing is a little hard to read (the multi-line wrapping). Is it possible to construct the type elsewhere?

src/pattern-library/components/patterns/transition/SliderPage.tsx
```diff
@@ -76,17 +77,17 @@ export default function SliderPage() {
         </Library.Pattern>
 
         <Library.Pattern title="Props">
-          <Library.Example title="visible">
-            This mandatory boolean prop tells if the Slider is currently
-            displayed or hidden.
+          <Library.Example title="direction">
+            This prop tells if the Slider is currently expanded (<code>in</code>
+            ) or collapsed (<code>out</code>). It is <code>in</code> by default.
```
lyzadanger, line 82: A task for me, especially now that we have reliable routing and fragment navigation, is to create a better centralized documentation for common component props. That is, this documentation probably shouldn't be necessary specifically for `Slider`, instead, we should be able to reuse or link to generalized documentation for this prop.

lyzadanger (approved):
Cool. Let's get the tests and plop templates sorted out now!

src/pattern-library/components/patterns/transition/SliderPage.tsx
acelaya: Actually, this is now redundant, and `_transitionStatus` can be completely discarded in favor of `direction`. It was needed when `visible` was used, but not anymore.

This can be simplified as `const Slider_: TransitionComponent = ...`.

acelaya: Ok no, my mistake. Doing that has other side effects.

I will extract the type in multiple definitions for better readability.

PR #997, opened 2023-04-26
Make sure all dev dependencies are defined in devDependencies

acelaya:
This PR moves `wouter-preact` and `highlight.js` as `devDependencies`, since they are used only for the pattern library, and this way we will save a bit of bandwidth when installing dependencies in `client`, `lms` and `browser-extension`.


acelaya:
Never mind. The pattern-library is actually used by `@hypothesis/frontend-shared` consumers. My mistake.

Issue #973, opened 2023-04-20
Host pattern-library independently


lyzadanger:
Find a way, any straightforward way, to deploy the pattern-library to a public URL independently of the rest of this package. This will allow us to share patterns and prototypes with a wider audience. The pattern library is getting mature enough now that this makes sense.

acelaya:
There are a couple of options that come to my mind.

* Build a nodejs-based docker image which serves the pattern library as we do in development, using the serve-pattern-library script which uses express to serve the page.
* Build an nginx-based docker image which builds the static pattern-library assets on a first stage, and then copies them to the nginx stage.
* Build the assets, upload them to S3 and serve them via Cloudflare (like we do with the client).

The first two options have the benefit that we can incorporate the building process and deployment to other existing workflows, and then send it to Elastic Beanstalk. However, if we cannot re-use enough of the existing tooling, it might require extra infra work.

From those two, using nginx requires a bit of work on replicating express config into nginx config, but it shouldn't be super hard.

On the other hand, using express has the problem that express, mustache, etc would become a production dependency, but we don't want them to be a prod dependency of the library, so we need to find a way to "split" that.

Finally, the option about using Cloudflare comes with multi-region capabilities out of the box (something we don't have with Elastic Beanstalk, unless we put Cloudflare in front of our deployment anyway), and in theory we should also be able to reuse some deployment logic from the client.

acelaya:
The pattern library is now available in https://patterns.hypothes.is/

acelaya:
Closing this issue, as the pattern library is now hosted.

There are a couple of improvements and missing docs that we can do separately anyway.

PR #998, opened 2023-04-26
Add Dockerfile and required config to build and serve pattern library independently

acelaya:
> This PR is part of https://github.com/hypothesis/frontend-shared/issues/973

This PR adds the required configuration to build a docker image which statically builds the pattern library, and then serves it with nginx on port 5001.

Deployment configs and such will follow this PR:
- Deployment workflow: https://github.com/hypothesis/frontend-shared/pull/999

### Testing steps

1. Check out this branch.
2. Run `docker build . -t hypothesis_frontend_shared`. This will build the docker image and tag it as `hypothesis_frontend_shared`.
3. Run a docker container based on the image just built: `docker run --rm -p 5001:5001 hypothesis_frontend_shared`
    * *`--rm` makes sure the container is discarded once stopped*
    * *The container will expose nginx on port `5001` of the host machine. In the unlikely case you are using this port, try with a different number*
4. Navigate to http://localhost:5001 (or the port you used) and check the pattern library looks ok.
    * Check also that once you have navigated somewhere other than `/`, you can refresh the page and the same section is loaded.
    * Check that `/_status` returns a 200 response. 
5. Verify "not found" URLs:
    * Dynamic "not-found" routes are handled client side, and nginx falls back to `index.html`. Navigate to something like http://localhost:5001/foo, and verify that you see this:
      image
    * Invalid static assets (any URL containing a dot is considered a file with an extension). Navigate to something like http://localhost:5001/foo.css, and verify that you see an regular nginx 404 error:  
      image

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -1,0 +1,23 @@
+server {
+    listen 80 default_server;
+    charset utf-8;
+    root /usr/share/nginx/html;
+    index index.html;
+
+    # Handle /_status endpoint with a 200 response once the container is up
+    location = /_status {
+        add_header Content-Type text/plain;
+        return 200 'Success!';
+    }
+
+    # When requesting static paths with an extension, try them, and return a 404 if not found
+    location ~* .+\..+ {
+        try_files $uri $uri/ =404;
+    }
+
+    # When requesting a path without extension, try it, and return the index if not found
+    # This allows for client-side route handling
+    location / {
+        try_files $uri $uri/ /index.html$is_args$args;
+    }
+}
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -1,5 +1,5 @@
 server {
-    listen 80 default_server;
+    listen 5001 default_server;
     charset utf-8;
     root /usr/share/nginx/html;
     index index.html;
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+./.git*
+./build
+./docs
+./lib
+./node_modules
+./test
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM node:19.8.1-alpine as builder
+
+COPY . /frontend-shared
+RUN cd /frontend-shared && \
+    yarn install --frozen-lockfile && \
+    yarn build-pattern-lib
+
+
+FROM nginx:1.24.0-alpine
+
+RUN rm -r /usr/share/nginx/html && rm /etc/nginx/conf.d/default.conf
+# Copy pattern library static assets, and put them in nginx document root folder
+COPY --from=builder /frontend-shared/build /usr/share/nginx/html
+COPY ./templates/pattern-library.mustache /usr/share/nginx/html/index.html
+COPY conf/nginx.conf /etc/nginx/conf.d/default.conf
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,3 +13,5 @@ RUN rm -r /usr/share/nginx/html && rm /etc/nginx/conf.d/default.conf
 COPY --from=builder /frontend-shared/build /usr/share/nginx/html
 COPY ./templates/pattern-library.mustache /usr/share/nginx/html/index.html
 COPY conf/nginx.conf /etc/nginx/conf.d/default.conf
+
+EXPOSE 5001
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
   "scripts": {
     "build-lib": "babel src/ --out-dir lib/ --extensions '.js,.ts,.tsx' --source-maps --ignore '**/test' --ignore '**/karma.config.js'",
     "build": "yarn build-lib && tsc --build src/tsconfig.json",
+    "build-pattern-lib": "yarn gulp bundle-css && yarn rollup -c rollup.config.js",
     "typecheck": "tsc --build src/tsconfig.json",
     "lint": "eslint .",
     "checkformatting": "prettier --check '**/*.{js,scss,ts,tsx,md}'",


config/nginx.conf
acelaya, line 1: I wasn't sure where to put this file. I'm open to suggestions.

acelaya:
@lyzadanger I would like to discuss a couple points on this, an have your feedback on some decisions before we move forward.

* First, what you mentioned yesterday about having to serve static images at some point.

  This is currently configured to be able to serve any static asset. However, it is only copying the css and js bundles, together with the mustache template (more on this in next point).

  Is there any directory where you would expect static images and other assets to be placed, that we can copy in the docker image by default to make sure it works when needed?

* And second, this is now making the naive assumption that the mustache template is actually static HTML, and copying it as `index.html`.

  This is true now, but it could stop being true at some point, and it is not obvious when you are editing the mustache template.

  If we change the template later and add handlebars-specific logic, it will no longer work as a static `index.html`.

conf/nginx.conf
```diff
@@ -0,0 +1,23 @@
...
+    index index.html;
+
+    # Handle /_status endpoint with a 200 response once the container is up
+    location = /_status {
+        add_header Content-Type text/plain;
+        return 200 'Success!';
+    }
+
+    # When requesting static paths with known extensions, try them, and return a 404 if not found
+    location ~* .+\.(css|js|html|png|jpe?g|gif|bmp|ico|json|csv|otf|eot|svg|svgz|ttf|woff|woff2|ijmap|pdf|tif|map) {
```
lyzadanger, line 14: BUT WHAT ABOUT `TIFF`?! No seriously, is this regexp reasonable for maintenance?

conf/nginx.conf
```diff
@@ -0,0 +1,23 @@
...
+        try_files $uri $uri/ =404;
+    }
+
+    # When requesting a path without extension, try it, and return the index if not found
+    # This allows for client-side route handling
+    location / {
```
lyzadanger, line 20: What if there is no trailing slash, e.g. the request is for `https://whateverWeCallThisThing.hypothes.is`. Will this routing still work?

Blah, the age-old trailing-slash headache. :)

conf/nginx.conf
acelaya: Haha. Yeah, I think you have a good point there.

I actually took this from an old project where I first used this kind of approach and didn't stop to think it through too much.

I think just checking if it's a URL with an extension would be enough to safely assume it's a static file URL, without checking for specific known extensions.

Of course, if at some point we use dots in dynamic routes, it might not work, but I suppose that's not very likely.

indigobravo:
I've tested the process of creating a container and running it. I see this...

Screenshot 2023-04-27 at 15 14 49

conf/nginx.conf
acelaya: Yeah, it works too, because even though you can type a URL which is only the domain with no path (or empty path), in practice that does not "exist". HTTP clients always send `/` as the path, which matches this config:

See this for example: this is this docker image being run on port 8765. The browser does not display any path, but you can see the actual request includes `/`.

image

Same happens if you try using `cURL`: The request is performed to `/` path, even if you don't explicitly include it:

```shell
{16:18}~ ➭ curl -v http://localhost:8765
*   Trying 127.0.0.1:8765...
* Connected to localhost (127.0.0.1) port 8765 (#0)
> GET / HTTP/1.1
> Host: localhost:8765
> User-Agent: curl/7.85.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: nginx/1.24.0
< Date: Thu, 27 Apr 2023 14:23:58 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 362
< Last-Modified: Mon, 09 Jan 2023 13:40:07 GMT
< Connection: keep-alive
< ETag: "63bc1937-16a"
< Accept-Ranges: bytes
< 
<!DOCTYPE html>
<html>
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <title>UI component pattern library</title>
  <link rel="stylesheet" href="/styles/pattern-library.css">
</head>
<body>
  <div id="app"></div>
  <script type="module" src="/scripts/pattern-library.bundle.js"></script>
</body>
</html>
* Connection #0 to host localhost left intact
```

And explicitly including `/`:

```shell
{16:18}~ ➭ curl -v http://localhost:8765/
*   Trying 127.0.0.1:8765...
* Connected to localhost (127.0.0.1) port 8765 (#0)
> GET / HTTP/1.1
> Host: localhost:8765
> User-Agent: curl/7.85.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: nginx/1.24.0
< Date: Thu, 27 Apr 2023 14:24:21 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 362
< Last-Modified: Mon, 09 Jan 2023 13:40:07 GMT
< Connection: keep-alive
< ETag: "63bc1937-16a"
< Accept-Ranges: bytes
< 
<!DOCTYPE html>
<html>
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <title>UI component pattern library</title>
  <link rel="stylesheet" href="/styles/pattern-library.css">
</head>
<body>
  <div id="app"></div>
  <script type="module" src="/scripts/pattern-library.bundle.js"></script>
</body>
</html>
* Connection #0 to host localhost left intact
```

indigobravo (approved):
Looking good. I have tested the creation of the Docker container and all appears to work as it should. If you can address the points raised by Lyza we can get this hosted.

acelaya:
@indigobravo I have changed the `Dockerfile` and nginx config to serve on port `5001`.

I have also edited the comments and descriptions on this PR.

lyzadanger:
@acelaya We might be able to set up the local express server to apply the `mustache` templating engine to `html` files and rename the `pattern-library.mustache` file to `pattern-library.html` (or whatever). If that would ease confusion. See https://expressjs.com/en/4x/api.html#app.engine

acelaya:
> `/images/` :) -- it already exists
> 
> It could potentially be useful to have the source `.svg`s from the icon components (already in `images/icons`) served. e.g. `https://patterns.hypothes.is/images/icons/arrow-up.svg`

@lyzadanger this should allow to serve static images from the `images` folder https://github.com/hypothesis/frontend-shared/pull/1025

PR #1001, opened 2023-04-26
Add `secondary` variant to `CardTitle` and `CardHeader`

lyzadanger:
This PR expresses a design pattern we are using in the `lms` application by updating `variants` for `CardTitle` and `CardHeader` to add a new `secondary` variant (default is `primary`, the way it "looked before.").

This captures the pattern we've started using in panels in the `lms` app. It will make markup more straightforward for LMS components, and make prototyping within panels of that style faster and more consistent.

Partial screen grab of updated `Card` pattern library page, in the `CardHeader` section:

image

### DIFF
--- a/src/components/layout/CardTitle.tsx
+++ b/src/components/layout/CardTitle.tsx
@@ -4,7 +4,15 @@ import type { JSX } from 'preact';
 import type { PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 
+type ComponentProps = {
+  /** Wrap the title with this HTML heading tag  */
+  tagName?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5';
+
+  variant?: 'primary' | 'secondary';
+};
+
 export type CardTitleProps = PresentationalProps &
+  ComponentProps &
   JSX.HTMLAttributes<HTMLElement>;
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/layout/CardTitle.tsx
```diff
@@ -4,7 +4,15 @@ import type { JSX } from 'preact';
 import type { PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 
+type ComponentProps = {
+  /** Wrap the title with this HTML heading tag  */
+  tagName?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5';
```
lyzadanger, line 9: I realized that we lost something in the most recent round of `Card` and `Panel` implementations: we aren't using heading elements. For HTML semantics and accessibility it is best if we do. This prop allows a consumer to use a different heading level than default (h1) if so desired.

lyzadanger:
I'll fix the test miss tomorrow.

src/components/layout/CardTitle.tsx
```diff
@@ -15,16 +23,28 @@ const CardTitle = function CardTitle({
...
+  const WrapperElement = tagName;
   return (
     <div
       data-component="CardTitle"
       {...htmlAttributes}
-      className={classnames('grow text-lg text-brand font-semibold', classes)}
+      className={classnames(
+        {
+          'text-lg text-brand font-semibold': variant === 'primary',
+          'text-xl text-slate-7 font-normal': variant === 'secondary',
```
acelaya, line 39: Is the text supposed to be bigger on `secondary` variant? Is that intended?

acelaya (approved):
LGTM

src/components/layout/CardTitle.tsx
lyzadanger: Yes, it actually is!

PR #1002, opened 2023-04-27
Transition common tests

acelaya:
Create common logic via `testTransitionComponent` helper function, that can be used by all transition components tests.

I have moved there everything that made sense from `Slider-test` (our only real transition component so far). However, I have left most of the stuff out of this, because it was too specific for the `Slider`, which animates the content height, and has a transition which either fully displays its content, or fully hides it.

Other transition components might animate different aspects, or have an `in`/`out` state where the content is not fully displayed or hidden, but something in between.

Because of that I don't want to make any further assumptions, until we have more `TransitionComponent` examples.

### DIFF
--- a/src/components/test/common-tests.js
+++ b/src/components/test/common-tests.js
@@ -5,6 +5,7 @@ import { checkAccessibility } from '../../../test/util/accessibility';
 
 /**
  * @typedef {import('preact').FunctionComponent} FunctionComponent
+ * @typedef {import('../../types').TransitionComponent} TransitionComponent
  */
 
 const createComponent = (Component, props = {}) => {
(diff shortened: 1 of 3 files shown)


src/components/test/common-tests.js
```diff
@@ -237,3 +238,40 @@ export function testSimpleComponent(Component) {
...
+  const displayName = componentName ?? Component.displayName ?? Component.name;
+
+  describe(`Common transition component functionality for ${displayName}`, () => {
+    it('renders', () => {
+      const wrapper = createComponent(Component);
+      assert.isTrue(wrapper.find(Component).exists());
+    });
+
+    ['in', 'out'].forEach(direction => {
+      it('should handle unmounting on either direction', () => {
```
lyzadanger, line 261: Can you help me understand exactly what this is testing (specifically, why unmounting is tested)?

src/components/test/common-tests.js
```diff
@@ -237,3 +238,40 @@ export function testSimpleComponent(Component) {
...
+        const wrapper = createComponent(Component, { direction });
+        wrapper.setProps({ direction: direction === 'in' ? 'out' : 'in' });
+        wrapper.unmount();
+      });
+    });
```
lyzadanger, line 266: Do we need a way to test that the `TransitionComponent` will invoke the `onTransitionEnd` callback? I know that this introduces a time component into the testing...

src/components/transition/test/Slider-test.js
```diff
@@ -24,6 +24,8 @@ describe('Slider', () => {
     container.remove();
   });
 
+  testTransitionComponent(Slider);
+
```
lyzadanger, line 28: Regarding the description of the next test: Does it render _collapsed_ or is it actually just plain old _hidden_?

lyzadanger (commented):
This is moving along nicely! I am wondering if we need a way to test the `onTransitionEnd` prop?

Otherwise I think we're darned near ready to land this. See what you think!

src/components/transition/test/Slider-test.js
acelaya: Since this is a Slider-specific test, I thought it made sense to keep the original test descriptions which use the "collapse"/"expand" concepts.

For other TransitionComponents it might indeed not apply.

src/components/test/common-tests.js
acelaya: Good point. Actually, I think the purpose of this test was probably that, and it got lost at some point (I migrated it from `client` first, and then here. I may have messed it up 😅)

src/components/test/common-tests.js
acelaya: See https://github.com/hypothesis/frontend-shared/pull/1002/files/f1eabd4ac120780602a06b28347481136ece4d0e#r1179255457

src/components/test/common-tests.js
acelaya: I have changed the test. No it covers the invocation to `onTransitionEnd`.

lyzadanger (approved):
Woot!

PR #1003, opened 2023-04-27
UI Sketch "Tennis" with some possible options for buttons in the LMS content-picking interface

lyzadanger:
This PR is a set of sketch prototypes riffing on the LMS content-picking layout, with a goal of finding a way to "fix the buttons." 

This iterative "tennis" starts from the current UI:

image

and explores some options, from the simpler and more mundane:

image

to the more fanciful or aspirational:

image

This represents two free-flowing sessions that were loose and not intended to be a one-size-fits-all sort of solution but to spark some chats about what we might do to improve this interface. Of course, we can mix and match any part of this, or push into round 3 based on other people's thoughts.

Visit http://localhost:4001/lms-content-selection in the pattern library on this branch to see things.


src/components/icons/Vitalsource.tsx
robertknight, line 1: I suggest to name this `VitalSource.tsx` since that's how the name is normally written. Incidentally `JSTOR` is an acryonym/abbreviation for "Journal Storage".

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
```diff
@@ -0,0 +1,758 @@
...
+function Button5({
+  children,
+  icon,
+  contentType = 'pdf',
+}: {
+  children: ComponentChildren;
+  contentType?: string;
+  icon?: IconComponent;
+}) {
+  const Icon = icon ?? GoogleDriveIcon;
```
robertknight, line 100: You could use our standard PDF icon as the fallback icon, since that is the only content type supported by the sources that don't currently have their own icon.

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
```diff
@@ -0,0 +1,758 @@
...
+                    </div>
+                    <div className="space-y-6 text-[14px] leading-none">
+                      <div className="flex flex-row p-y-2">
+                        <div className="grid grid-cols-4 gap-2 w-full">
+                          <Button6 contentType="web page | PDF" icon={LinkIcon}>
+                            URL
+                          </Button6>
+                          <Button6>Canvas</Button6>
+                          <Button6>Google Drive</Button6>
+                          <Button6 contentType="article">JSTOR</Button6>
```
robertknight, line 739: Should this button use the JSTOR icon?

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
```diff
@@ -0,0 +1,758 @@
...
+            </Library.Demo>
+          </Library.Example>
+        </Library.Pattern>
+
+        <Library.Pattern title="Iterative: Round 1">
+          <Library.Example title="Use grid layout">
+            <p>
+              It is quick to put the buttons in a two-column grid layout.
+              However, because of their heavy, dark color, one ends up with an
+              unpleasant optical illusion between button corners.
```
robertknight, line 241: I see a light grey "diamond" illusion appear in the white space where the four button corners meet. Is that what you mean by optical illusion?

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
```diff
@@ -0,0 +1,758 @@
...
+                            PDF from <strong>Google Drive</strong>
+                          </Button1>
+                          <Button1>
+                            <strong>JSTOR</strong> article
+                          </Button1>
+                          <Button1>
+                            PDF from <strong>OneDrive</strong>
+                          </Button1>
+                          <Button1>
+                            book from <strong>VitalSource</strong>
```
robertknight, line 401:
```suggestion
                            Book from <strong>VitalSource</strong>
```

robertknight (approved):
I like the directions presented here. I have a few ideas on the details, but I think it might be easiest to discuss them by walking through the sketches on a call.

src/components/icons/Vitalsource.tsx
lyzadanger: Oh, I agree on the casing. I was just moving fast. One thing is certain: This is not the icon I would want to use in production because it fills its viewbox too much (I'd want to adjust it to have a little space around it because of the inverse nature of the background "box").

This leads me to think that I might want to add a task to the background longlist to possibly isolate icon components used for prototypes from production/shipping/package icons. This would require some adjustments to the icon-generating script.

Annnnyway, I can go adjust casing right now.

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
lyzadanger: That's a great point. I will do that now.

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
lyzadanger: Yes, but finding a workable JSTOR logo in SVG format was eluding me and I didn't want to drop more time into it. Their more elegant logo, which I was able to find in (COMPLEX) SVG  is not workable as an icon:

image

There is a variant I've seen that is just a "J", but I could not find an SVG of it:

image

I was able to find Google Drive, OneDrive and, surprisingly, a VitalSource image in SVG format, which allowed me to drop them in quickly here. 

I can add a callout to the affected sketches to alert anyone else who might look at them about this.

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
lyzadanger: Yes. Is there clearer wording I could use?

src/pattern-library/components/patterns/prototype/LMSContentSelectionPage.tsx
lyzadanger: Oh, derp, you probably noticed that I did try to create an icon from the elaborate logo and it's in the library but I didn't use it. It didn't work at small sizes. I'll remove the `JSTORLogoIcon` component.

PR #1005, opened 2023-04-28
Add grow class to CardTitle when used from Panel

acelaya:
Another PR removed the `grow` class from `CardTitle`, to make the `CardHeader` pass it explicitly instead.

However, I have just noticed it makes the close button be rendered next to the title in `Panel`s, because those render a `CardTitle` on their own, instead of relying on the logic inside `CardHeader`:

image

Dynamically adding `grow` class when `onClose` has been provided to `Panel`, fixes this:

image

### Testing

- While in `main` branch, go to http://localhost:4001/layout-panel and search for `onClose`. You will notice the panel renders the close button incorrectly:
  image
- Check out this branch, and then go again to http://localhost:4001/layout-panel. The close button should render properly now:
  image

### DIFF
--- a/src/components/layout/Panel.tsx
+++ b/src/components/layout/Panel.tsx
@@ -1,3 +1,4 @@
+import classnames from 'classnames';
 import type { ComponentChildren, JSX } from 'preact';
 
 import type { IconComponent, CompositeProps } from '../../types';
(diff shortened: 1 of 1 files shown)


src/components/layout/Panel.tsx
```diff
@@ -94,7 +95,7 @@ const Panel = function Panel({
     >
       <CardHeader onClose={onClose} fullWidth={scrollable || fullWidthHeader}>
         {Icon && <Icon className="w-em h-em" />}
-        <CardTitle>{title}</CardTitle>
+        <CardTitle classes={classnames({ grow: !!onClose })}>{title}</CardTitle>
```
acelaya, line 98: I have added the class only hen `onClose` is provided, but maybe `classes="grow"` would be enough.

acelaya:
Closing this in favor of #1006

PR #1006, opened 2023-04-28
Ensure proper layout of `CardHeader`

lyzadanger:
This PR is an alternative fix to https://github.com/hypothesis/frontend-shared/pull/1005 that adds some updated examples to the pattern library to make this kind of thing easier to spot in future.

It fixes an alignment problem in `CardHeader`s that could cause the close button not to align right properly.

Fix regression in certain compositions of `CardHeader` by using a spacing div that will ensure that the close button is always aligned right.

### DIFF
--- a/src/components/layout/CardHeader.tsx
+++ b/src/components/layout/CardHeader.tsx
@@ -60,12 +60,9 @@ const CardHeader = function CardHeader({
       )}
       ref={downcastRef(elementRef)}
     >
-      {title && (
-        <CardTitle classes="grow" variant={variant}>
-          {title}
-        </CardTitle>
-      )}
+      {title && <CardTitle variant={variant}>{title}</CardTitle>}
       {children}
+      <div className="grow" />
       {onClose && (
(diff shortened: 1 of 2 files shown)


src/components/layout/CardHeader.tsx
```diff
@@ -60,12 +60,9 @@ const CardHeader = function CardHeader({
...
       ref={downcastRef(elementRef)}
     >
-      {title && (
-        <CardTitle classes="grow" variant={variant}>
-          {title}
-        </CardTitle>
-      )}
+      {title && <CardTitle variant={variant}>{title}</CardTitle>}
       {children}
+      <div className="grow" />
```
lyzadanger, line 65: This isn't my favorite solution (empty "spacer" divs), but it can be helpful in situations where otherwise the `flex-grow` would need to be assigned to a different element depending on context. This is a little more foolproof.

acelaya (approved):
Thanks!

PR #1026, opened 2023-05-15
Simplify dev server

acelaya:
This PR simplifies the script that runs the development express server. The changes are:

* Remove mustache entirely. The only template we use is actually static html, so we don't need a templating system.
* Expose the `images` folder as a static source, so that we can access source icon images.

Those two changes make the dev server closer to how the patterns docker container behaves.

With these changes, the `templates` folder might not make a lot of sense anymore. Maybe we can later consider having a `public` folder containing the index.html file and the `images` folder on it. So basically everything which is already static, does not need any building to be served, and we want to be accessible via HTTP.

```tree
public
├── images
│   └── icons
│       └── ...
└── index.html
```

### Testing steps

1. Check out this branch
2. Run dev server: `make dev`
3. Make sure the overall navigation works in http://localhost:4001
4. Verify static images can also be accessed locally http://localhost:4001/images/icons/annotate.svg (any ther icon should also work).


scripts/serve-pattern-library.js
```diff
@@ -1,19 +1,20 @@
...
+
+  // Map paths from which we serve static files
   app.use('/scripts', express.static(path.join(dirname, '../build/scripts')));
   app.use('/styles', express.static(path.join(dirname, '../build/styles')));
+  app.use('/images', express.static(path.join(dirname, '../images')));
+
+  // For any other path, serve the index.html file to allow client-side routing
   app.get('/:path?', (req, res) => {
-    res.render('pattern-library');
+    res.sendFile(path.join(dirname, '../templates/index.html'));
```
acelaya, line 17: From express docs:

> How do I render plain HTML?
> You don’t! There’s no need to “render” HTML with the res.render() function. If you have a specific file, use the res.sendFile() function

https://expressjs.com/en/starter/faq.html#how-do-i-render-plain-html

lyzadanger (approved):
Thanks! I do agree that we can do further work to simplify the structure of the static assets.

Note that the `images/icons` directory is referenced by the `generate-icons` script and will need to be updated if those icons are moved.

PR #1028, opened 2023-05-17
Extend `Link` with Styling API, update documentation

lyzadanger:
~~Depends on #1029~~

This PR pilots the styling API proposed in https://docs.google.com/presentation/d/1uTHdWrXKqVuivMI6Z3JR91ZP-Qb_0Ki2fvIAht7JSbk by updating the `Link` component and a whole lot of related documentation in the pattern library.

There's maybe 50 lines of actual functional change here and a whole bunch of supporting documentation and formatting.

The basic chunks of this work are:

* ~~Adjustments to some `Library` components to fix a few glitches (margins, mostly) and add a way to format API props more cleanly~~ (These were chunked off into a separate PR #1029)
* Updates to `Link` and its pattern-library documentation
* Restructure of the "Using Components" page: Add the API details of the Styling API
* Replacement of the "Customizing Components" page with a "Styling Components" page. This covers how to _use_ the Styling API

This work is additive, not breaking.

## Testing

I suggest checking out this branch, running `make dev` and visiting the:

* [Documentation for `Link`](http://localhost:4001/navigation-link)
* [Using Components page](http://localhost:4001/using-components)
* [Styling Components page](http://localhost:4001/styling-components)

## TODOs

* [x] Determine appropriate tests to add for this

## Future enhancements that would make this better

* Auto-generated permalinks and tables of contents would make pattern-library pages easier to navigate

image


src/components/navigation/Link.tsx
```diff
@@ -23,35 +28,45 @@ const Link = function Link({
...
+      className={classnames(
         {
-          // color
-          'text-brand hover:text-brand-dark': color === 'brand', // default
-          'text-color-text-light hover:text-brand': color === 'text-light',
-          'text-color-text hover:text-brand-dark': color === 'text',
+          'focus-visible-ring rounded-sm': styled,
+          // underline
+          // TODO: Underline should be controlled by `variant` and should default
+          // to `always`
```
lyzadanger, line 52: I want to acknowledge that `underline` is a problematic prop that doesn't fit into this model neatly. Which means I think it should go in the short future.

src/components/navigation/Link.tsx
```diff
@@ -23,35 +28,45 @@ const Link = function Link({
...
+
   return (
-    <LinkBase
+    <a
       data-component="Link"
+      rel="noopener noreferrer"
       {...htmlAttributes}
-      classes={classnames(
-        'rounded-sm',
-        // NB: Base classes are applied by LinkBase
```
lyzadanger, line 36: Not anymore. This is a good thing because even this small abstraction caused a regression in one of these components recently.

src/pattern-library/components/Library.tsx
```diff
@@ -63,20 +63,26 @@ function Heading({
 }: LibraryHeadingProps) {
   if (level <= 2) {
     return (
-      <h2 className="text-3xl text-slate-600 font-bold" {...htmlAttributes}>
+      <h2
+        className="text-3xl text-slate-600 font-bold mb-4"
```
lyzadanger, line 67: Margin fixes. If the smattering of changes in this module feel noisy, you might do better reviewing commit by commit.

src/pattern-library/components/Library.tsx
```diff
@@ -124,7 +130,7 @@ function Section({
         {intro && (
           <div className="text-base space-y-3 leading-relaxed">{intro}</div>
         )}
-        <div className="leading-relaxed">{children}</div>
+        {children}
```
lyzadanger, line 133: Removing containing element to fix margins.

src/pattern-library/components/patterns/UsingComponentsPage.tsx
```diff
@@ -1,331 +1,320 @@
...
+              </Library.InfoItem>
+              <Library.InfoItem label="example">
+                <Library.Code
+                  content={`<Link id="my-own-id" href="https://www.example.com">A Link with attributes</Link>`}
+                  size="sm"
+                />
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
+                <code>
+                  {'preact.JSX.HTMLAttributes<T extends HTMLElement>'}
```
lyzadanger, line 84: I'm not at all convinced I got "types" right in these docs so I'm open to feedback.

src/pattern-library/components/patterns/UsingComponentsPage.tsx
```diff
@@ -1,331 +1,320 @@
...
-            </p>
-            <p>
-              Example: <code>ScrollBox</code>, <code>Panel</code>.
-            </p>
-          </>
-        }
-      >
-        <Library.Pattern title="Props API">
+        <Library.Section
+          id="presentational-components-styling-api"
```
lyzadanger, line 92: Added a few IDs by hand in these pages so I could cross link. Hope to automate these soon.

src/pattern-library/components/patterns/UsingComponentsPage.tsx
```diff
@@ -1,331 +1,320 @@
...
-              </Card>
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Many presentational components have multiple variants. Variants{' '}
+                {'"theme"'} a component, affecting such aspects as color,
+                transitions, shadows, etc. Set to <code>{"'custom'"}</code> to
+                disable variant styling.
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
+                <code>{`'custom' | string`}</code>
```
lyzadanger, line 120: I'm 99% sure this isn't the way I should express this "type".

src/components/navigation/Link.tsx
```diff
@@ -23,35 +28,45 @@ const Link = function Link({
...
-          'text-color-text-light hover:text-brand': color === 'text-light',
-          'text-color-text hover:text-brand-dark': color === 'text',
+          'focus-visible-ring rounded-sm': styled,
+          // underline
+          // TODO: Underline should be controlled by `variant` and should default
+          // to `always`
+          'no-underline hover:no-underline': styled && underline === 'none', // default
+          'underline hover:underline': styled && underline === 'always',
+          'no-underline hover:underline': styled && underline === 'hover',
         },
```
acelaya, line 56: Since `styled` is a condition for each one of the styles, perhaps we could consider a different formula.

```tsx
className={classnames(
  {
    'focus-visible-ring rounded-sm': styled,
  },
  styled && {
    'no-underline hover:no-underline': underline === 'none', // default
    'underline hover:underline': underline === 'always',
    'no-underline hover:underline': underline === 'hover',
  },
  // [...]
)}
```

It would allow not having to add the condition every time, but perhaps it makes is less clear, or too subtle. What do you think?

acelaya (commented):
I like how it looks like :)

src/components/navigation/Link.tsx
robertknight: My intuition is that a single enum "variant" is probably not going to be enough to accommodate the different "standard"/built-in styles that a component may have. I wouldn't be surprised if we need flag props for example to turn certain elements of the standard style on/off. If the component is unstyled or using a custom variant or size, then these props wouldn't apply though.

robertknight (commented):
Generally LGTM. If I understand the intended use of `variant`, I do suspect we will find that a single `variant` value may not be enough to express commonly used knobs/switches that still fit within the scope of a "standard style". A component could say that certain props are only applicable when using particular sizes/variants though, so we don't try to make them work if the component is unstyled or using eg. a custom variant.

acelaya (approved):
LGTML

PR #1029, opened 2023-05-18
Pattern library: Adjust margins, add `Info`, `InfoItem` components

lyzadanger:
This PR chunks off changes made to the pattern library components to support #1028 so they can be available for other work. You can see them in play, if you so desire, in #1028.


acelaya (approved):
👍🏼

Issue #1044, opened 2023-05-22
Dialog is rendering a close button when no `onClose` is provided


lyzadanger:
`Dialog` supports a legacy use case of "non-closeable" modals. As such, the `onClose` prop is currently optional. When not provided, no close button should be rendered.

As of the addition of `TransitionComponent` support, there is always an internal close handler generated and, as such, a close button is always rendered. You can see a close button appearing even when no close callback is provided in one of the examples on the Dialog's pattern-library documentation page:

<img width="787" alt="image" src="https://github.com/hypothesis/frontend-shared/assets/439947/c0e28147-84eb-4260-9b3a-97ee6f853d06">

We shouldn't render a close button if no `onClose` is provided to `Dialog`.

PR #1045, opened 2023-05-22
Make sure Dialogs do not render close button when onClose callback is not provided

acelaya:
Closes https://github.com/hypothesis/frontend-shared/issues/1044

This PR changes the logic in `Dialog` to make sure an `onClose` callback is not passed down to the `Panel`, if the `Dialog` itself did not get one.

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -54,8 +54,6 @@ export type DialogProps = PresentationalProps &
   ComponentProps &
   Omit<PanelProps, 'fullWidthHeader'>;
 
-const noop = () => {};
-
 /**
  * Show a dialog
  */
@@ -74,7 +72,7 @@ const Dialog = function Dialog({
   // Forwarded to Panel
   buttons,
   icon,
-  onClose = noop,
+  onClose,
   paddingSize = 'md',
   scrollable = true,
   title,
@@ -87,6 +85,7 @@ const Dialog = function Dialog({
   );
   const [transitionComponentVisible, setTransitionComponentVisible] =
     useState(false);
+  const isClosableDialog = !!onClose;
 
   const closeHandler = useCallback(() => {
     if (TransitionComponent) {
@@ -94,7 +93,7 @@ const Dialog = function Dialog({
       // called by that component, once the "out" transition has finished
       setTransitionComponentVisible(false);
     } else {
-      onClose();
+      onClose?.();
     }
   }, [onClose, TransitionComponent]);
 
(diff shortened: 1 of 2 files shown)


src/components/feedback/Dialog.tsx
```diff
@@ -216,7 +215,7 @@ const Dialog = function Dialog({
           buttons={buttons}
           fullWidthHeader={true}
           icon={icon}
-          onClose={closeHandler}
+          onClose={isClosableDialog ? closeHandler : undefined}
```
acelaya, line 218: I could just have done `onClose ? closeHandler : undefined`, but I feel starting with `isClosableDialog` explains better the "why".

src/components/feedback/test/Dialog-test.js
```diff
@@ -282,6 +282,13 @@ describe('Dialog', () => {
         assert.called(onClose);
       });
     });
+
+    context('when no onClose callback is provided', () => {
+      it('does not render a close button', () => {
+        const wrapper = mount(<Dialog title="My dialog" />);
+        assert.isFalse(wrapper.find('CancelIcon').exists());
+      });
+    });
```
acelaya, line 291: This test fails in `main`.

src/components/feedback/Dialog.tsx
```diff
@@ -87,14 +85,15 @@ const Dialog = function Dialog({
...
+  const isClosableDialog = !!onClose;
 
   const closeHandler = useCallback(() => {
     if (TransitionComponent) {
       // When a TransitionComponent is provided, the actual "onClose" will be
       // called by that component, once the "out" transition has finished
       setTransitionComponentVisible(false);
     } else {
-      onClose();
+      onClose?.();
```
lyzadanger, line 96: Technically we know that `onClose` is present at this point, but TS wouldn't, I suppose. I'm fine leaving this as-is.

src/components/feedback/Dialog.tsx
lyzadanger: I think that's fine.

src/components/feedback/test/Dialog-test.js
lyzadanger: Nice job, thanks!

lyzadanger (approved):

src/components/feedback/Dialog.tsx
acelaya: That's actually a good point 🤔 

I kind-of want to give this a thought and see if I can find a way to make TS know this is actually set at this point.

Issue #676, opened 2022-10-11
Make it possible to "turn sizing off" for `Button`, perhaps other sized components


lyzadanger:
Padding and spacing within buttons can be adjusted via the `size` prop, which applies some canned padding and spacing for `sm`, `md`, `lg` values, e.g. However, we have an awful lot of legacy buttons with custom or mis-matched padding and spacing.

To match current custom button styling using `Button`, it's currently necessary to either:

* Pass `!important` `classes` (e.g. `classes="!px-1.5 !py-2"`) or
* Use `ButtonBase` instead (which won't apply sizing, but also won't apply colors, etc., so you have to re-invent the wheel a bit)

One option might be to introduce an `unsized` prop similar to `ButtonBase`'s `unstyled` prop. When set, padding and spacing would not be applied. The user of `Button` could then set non-important size-related classes via `classes`.

robertknight:
> One option might be to introduce an unsized prop similar to ButtonBase's unstyled prop. When set, padding and spacing would not be applied. 

How about an explicit `size={null}` or `size="unsized"`, to prevent a user from using an invalidation combination such as `<FooButton size="md" unstyled>`.

lyzadanger:
> > One option might be to introduce an unsized prop similar to ButtonBase's unstyled prop. When set, padding and spacing would not be applied.
> 
> How about an explicit `size={null}` or `size="unsized"`, to prevent a user from using an invalidation combination such as `<FooButton size="md" unstyled>`.

I'm torn, as this had also occurred to me. The up-side to another `size` value is as you stated. However, it's somewhat ergonomically awkward to have to remember and manage a value of another type (e.g. `null`).

The nice thing about a Boolean value is that it is a signal that it's putting the component in a non-default state (cf. `unstyled` on `ButtonBase`), it's easy to read, and doesn't require `size` to take a value of a different type. On the other hand, as you mentioned, it would allow for the invalid combination of `size="sm" unsized`. Hmmm, stroking chin.

lyzadanger:
I think this is part and parcel with #853 and should be given a deep think heading into v6.

Issue #1030, opened 2023-05-18
Consolidate structure of component documentation in Pattern Library


lyzadanger:
Why now? Because we need to ensure that documentation pages in the pattern library have a sensible and conventional structure for creating permalink section IDs. We want to add section IDs/permalinks so that we can create links to sections in the pattern library. This is useful for collaboration and reference and keeps coming up as a need.

Component documentation pages in the Pattern Library have grown over time and aren't chaos or anything, but I'd like them to have a predictable section/heading structure so that they are more internally consistent.

Based on work in #1028 , a structure that seems to make sense for component documentation pages is one top-level section per component, a la:

* ComponentName
  import code example
  "Basic ComponentName" demo
  * Using ComponentName (optional, for extra use-case demos or more free-form documentation)
  * Component API
    * one heading per prop...with prop info
  * Styling API (optional, for components that support it)
    * one heading per supported styling prop...

In practice, this looks like:

![screencapture-localhost-4001-data-aspectratio-2023-05-18-11_38_46](https://github.com/hypothesis/frontend-shared/assets/439947/5afb8b75-e984-419d-b84f-c4caf2ba0b85)

TODO:

* [x] AspectRatio
* [x] DataTable
* [x] Icons
* [x] ScrollBox
* [x] Table
* [x] Thumbnail
* [x] Dialogs
* [x] Spinner
* [x] Button
* [x] Checkbox
* [x] Input
* [x] InputGroup
* [x] OptionButton
* [x] Select
* [x] Card
* [x] Panel
* [x] Overlay
* [x] Link
* [x] LinkButton
* [x] PointerButton
* [x] Tabs
* [x] Slider

PR #1048, opened 2023-05-22
Make Button components implement Styling API, deprecate `ButtonBase`

lyzadanger:
This PR:

* Updates `Button` and `IconButton` to implement the styling API: `size`, `unstyled` and `variant`
* Extracts common styling-API basic tests to `common-tests` for reuse
* Deprecates `ButtonBase`
* Updates the heck out of Button* documentation in the pattern library

Previously, `IconButton` and `Button` both wrapped `ButtonBase`. Now, `IconButton` wraps `Button`. The integration of styling API props obviates the need for `ButtonBase`.

These are all additive changes.

As usual with these PR diffs, the bulk is the documentation changes.

### Checking it out

I recommend reviewing the documentation and examples on the [Button pattern library page](http://localhost:4001/input-button)

Fixes https://github.com/hypothesis/frontend-shared/issues/676
Part of #1030

### DIFF
--- a/src/components/input/Button.tsx
+++ b/src/components/input/Button.tsx
@@ -95,9 +95,9 @@ const Button = function Button({
       {...ariaProps}
       {...htmlAttributes}
       className={classnames(
-        styled && {
+        {
           'focus-visible-ring transition-colors whitespace-nowrap flex items-center':
-            true,
+            styled,
         },
         themed && {
           'font-semibold rounded-sm': true,
--- a/src/components/input/Button.tsx
+++ b/src/components/input/Button.tsx
@@ -1,28 +1,56 @@
 import classnames from 'classnames';
+import type { JSX } from 'preact';
 
 import type { IconComponent, PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
-import ButtonBase from './ButtonBase';
-import type { ButtonCommonProps, HTMLButtonAttributes } from './ButtonBase';
 
 type ComponentProps = {
-  /** Adjusts padding on button */
-  size?: 'xs' | 'sm' | 'md' | 'lg';
-  variant?: 'primary' | 'secondary';
+  /**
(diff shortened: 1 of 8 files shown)


src/components/input/Button.tsx
```diff
@@ -32,45 +60,66 @@ const Button = function Button({
...
-      classes={classnames(
-        'focus-visible-ring transition-colors whitespace-nowrap flex items-center',
-        'font-semibold rounded-sm',
-        {
-          // Variants
+      className={classnames(
+        styled && {
+          'focus-visible-ring transition-colors whitespace-nowrap flex items-center':
+            true,
+        },
```
acelaya, line 101: Maybe this would be more clear as:

```suggestion
        {
          'focus-visible-ring transition-colors whitespace-nowrap flex items-center':
            styled,
        },
```

src/components/input/IconButton.tsx
```diff
@@ -43,37 +47,46 @@ const IconButton = function IconButton({
...
       classes={classnames(
-        'focus-visible-ring transition-colors whitespace-nowrap flex items-center',
-        'justify-center gap-x-2 rounded-sm',
-        {
+        styled && {
+          'focus-visible-ring transition-colors rounded-sm whitespace-nowrap':
+            true,
+          'flex items-center justify-center': true,
+          // Adapt styling when in an InputGroup
+        },
```
acelaya, line 68: I think the approach of `<condition> && {...}` makes sense when every entry in the object has their own different conditions, but they all share a common one.

In any other case, perhaps it's more clear to go with this:

```suggestion
        {
          'focus-visible-ring transition-colors rounded-sm whitespace-nowrap':
            styled,
          'flex items-center justify-center': styled,
          // Adapt styling when in an InputGroup
        },
```

acelaya (approved):
Looks good

PR #1051, opened 2023-05-26
Extract and polish select LMS content button design

lyzadanger:
This PR adds a [new pattern-library page](http://localhost:4001/lms-content-button) that extracts a button component per the selected design approach for LMS content-selection buttons.

We can merge this and deploy the pattern library to get feedback.

I also updated uses of `ButtonBase` to `Button` in the content-selection sketches.

image


acelaya (approved):
👏🏼

PR #1064, opened 2023-05-30
Add optional input and output directory arguments to `scripts/generate-icons.js`

robertknight:
Add optional arguments to `scripts/generate-icons.js` to allow custom input/output dirs to be specified. This allows the script to be re-used to generate icons in other projects, or for other purposes. Per commit notes, I'm not using a proper CLI parser because the args are so simple. This can be changed if the syntax gets more complex.

The initial use case is to generate icon components for the video player app. The changes in this PR are enough for that. It would be convenient if the script was included in the package, then we could add a `package.json` script to run it in the downstream project.

While testing this I noticed that incremental runs of ESLint and Prettier were taking a long time because caching was not enabled, so I fixed that in a separate commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage/
 # Use Yarn rather than npm to manage the lockfile.
 package-lock.json
 
+.eslintcache
--- a/package.json
+++ b/package.json
@@ -67,9 +67,9 @@
     "build": "yarn build-lib && tsc --build src/tsconfig.json",
     "build-pattern-lib": "yarn gulp bundle-css && yarn rollup -c rollup.config.js",
     "typecheck": "tsc --build src/tsconfig.json",
-    "lint": "eslint .",
-    "checkformatting": "prettier --check '**/*.{js,scss,ts,tsx,md}'",
-    "format": "prettier --list-different --write '**/*.{js,scss,ts,tsx,md}'",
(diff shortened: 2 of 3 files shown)


acelaya:
Should we perhaps add all icons used by our apps in frontend-shared instead, and then use the already generated icons there?

That would mean we don't need to modify the script, but I might be missing something.

robertknight:
> Should we perhaps add all icons used by our apps in frontend-shared instead, and then use the already generated icons there?

If icons are generally useful, it makes sense to include them in this package. However I do think it would be useful to have a way to build and iterate on project-specific app icons without having to always add them to the shared package. For example if we have SVG icons for the different content sources in the LMS app, I think it makes sense to keep them in that project alongside the code that uses them.

acelaya:
Good point.

It will also allow for faster prototyping with the option to move the icons here later, if it makes sense.

acelaya (approved):
👍🏼

lyzadanger:
Just wanted to add a comment: I agree that _typically_ we want to include icon components in this package, so that they can be seen as a collection in the pattern library and are generally organized in one place, but I think it's reasonable to generate icon components in other projects from time to time, especially if you're prototyping or otherwise unsure that the icon in question will have broader use.

PR #1065, opened 2023-06-01
Add new `OptionButton` component

lyzadanger:
This PR adds a new `OptionButton` component. It is a semi-generalized implementation of the new button we need to satisfy https://github.com/hypothesis/lms/issues/2955 

I ended up creating a new button that is intended for use "when you might otherwise use a radio button or select element but you need a button instead." In full purity, the set of buttons in the LMS content-selection interface that we're targeting should maybe instead be radio buttons or a select element. But it's impractical to make that big of a change right now, so this component serves as an option-like button.

It takes an optional `details` prop (for content to render at the right edge of the button), aliases `selected` to `pressed` and otherwise forwards all other non-styling props to `Button`. Eh, the [pattern-library docs](http://localhost:4001/input-option-button) explain this more clearly.

I've also updated the [pattern-library prototype page](http://localhost:4001/lms-content-button) for LMS content selection to use this component.

image

### DIFF
--- a/src/components/input/OptionButton.tsx
+++ b/src/components/input/OptionButton.tsx
@@ -1,0 +1,76 @@
+import classnames from 'classnames';
+import type { ComponentChildren } from 'preact';
+
+import { CheckIcon } from '../icons/';
+import Button from './Button';
+import type { ButtonProps } from './Button';
+
+export type OptionButtonProps = {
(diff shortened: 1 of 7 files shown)


acelaya (approved):
👏🏼

Issue #1067, opened 2023-06-01
Make writing pattern-library documentation/markup easier


lyzadanger:
Composing pattern-library documentation isn't sheer hell or anything, but I miss markdown and it can be fiddly. I get annoyed when I have to write markup like this:

```JSX
<p><code>SomeComponent</code> {"doesn't"} <strong>really</strong> have a type of {`'round' | 'solid'`}</p>
```

cf. in markdown:

```
`SomeComponent` doesn't *really* have a type of `'round' | 'solid'`
```

Perhaps look at where [MDX](https://mdxjs.com/) is at or brainstorm other ways to make writing the actual documentation a little nicer.

acelaya:
+1 to MDX and using markdown in general to write docs.

robertknight:
MDX seems like one reasonable option to explore. Another would be some kind of `<Markdown/>` component within our existing JSX component pages that renders its text children as markdown. A solution that does the markdown -> JSX conversion at build time would be preferable from a performance perspective.

PR #1085, opened 2023-06-06
Add dependencies to be able to render/compile MDX files into pages

acelaya:
Very rough and early proof of concept on being able to write documentation pages in MDX.

It lacks:

- Proper styling. Generated elements do not have all the tailwind classes, but that can be addressed by adding custom `components` to the `MDXProvider`:
  ```tsx
  <MDXProvider
    components={{
      h1: props => (
        <h1 className="text-3xl text-slate-600 font-bold" {...props} />
      ),
      // ...
    }}
  >
    
  </MDXProvider>
  ```
- Support for advanced components, like the `Library.Demo` component. In an ideal world, we should be able to map `pre` or `code` to it via the same approach above, but I'm sure there will be some challenges.
- Proper types for mdx components. Supposedly adding `@types/mdx` should do, but TS is currently complaining about importing mdx files

This PR also provides two different approaches to use MDX in docs:

* [Writing the whole doc as a single markdown file](https://github.com/hypothesis/frontend-shared/pull/1085/files#diff-b0d5daf6d323f364324830c04ef3bdc09f677dc464d5f81db95dc6606cea141c)
  * 🟢 Allows you to abstract away from target "platform" when writing docs.
  * 🟢 Makes it easier to read individal doc pages.
  * 🔴 Requires further `MDXProvider` components configuration to support all requirements from current structure.
  * 🔴 Requires accepting some limitations and adapting the mental model.
* [Continue explicitly using `Library` components (`Section`, `Pattern`, etc), and using small pieces of markdown in between](https://github.com/hypothesis/frontend-shared/pull/1085/files#diff-3d91f962c3865a2c843d29b0e2b856279ade2bed05eb1a3b5dd9345e45a54372)
  * 🟢 Covers the purpose of reducing the awkwardness of manually writing paragraphs, code blocks, escaping characters, etc, without having a major impact on how docs are structured
  * 🟢 Requires less customization of the `MDXProvider`.
  * 🔴 Might be harder to read as a whole, specially when reviewing a PR, where you want to focus on the content and not the structure.
* Something in between.
  * We don't have to choose one or the other. It is always possible to delegate some structural decisions to something outside doc pages themselves, like`Library.Page` or `Library.Section`, while we still explicitly use components like `Library.Example` or `Library.Code`.

> This PR relates to https://github.com/hypothesis/frontend-shared/issues/1067


lyzadanger:
Hi @acelaya ! Just a confirmation that "I see this!" Thanks for looking into this a bit. I'll review when I have some spare brain and time cycles!

acelaya:
> Hi @acelaya : I was wondering if there might be another way of persisting this concept and what you learned that is not a PR. I do want us to get back to it but it might be a while.

The options I can think of:

* Just close the PR. GitHub will keep it around and it's still referenced from the issue, so we can come back to it eventually.
* Close the PR, but move some of the info in the description to a comment in the issue.

lyzadanger:
> > Hi @acelaya : I was wondering if there might be another way of persisting this concept and what you learned that is not a PR. I do want us to get back to it but it might be a while.
> 
> The options I can think of:
> 
> * Just close the PR. GitHub will keep it around and it's still referenced from the issue, so we can come back to it eventually.
> * Close the PR, but move some of the info in the description to a comment in the issue.

I'll let you decide! Whatever you prefer.

Issue #1084, opened 2023-06-05
Improve Cloudflare cache handling in patterns deployment


acelaya:
Cloudflare is caching the patterns page for too long in production (https://patterns.hypothes.is).

We need to find a way to properly instruct Cloudflare for how long does it have to be cached.

acelaya:
This proved to be challenging to get done in the right way.

I'll deprioritize it for now, until we have more capacity. If it suddenly becomes an urgent problem to solve, https://github.com/hypothesis/frontend-shared/pull/1077 can be a quick and dirty workaround.

acelaya:
Closing, as per the [comment](https://github.com/hypothesis/frontend-shared/issues/1084#issuecomment-1641724172) above.

PR #1088, opened 2023-06-07
Patterns deployment cache

acelaya:
This PR introduces a number of changes in order to improve and more granularly control the patterns deployment cache in Cloudinary:

- Update rollup's config to generate JS bundles with a content hash for production
- Update nginx config to cache assets in the next way:
    * `index.html` file should never be cached, as it contains references for the hashed bundles (alternatively we could change this and use a small cache period, like 5min)
    * CSS/JS bundles can be cached "forever", because once they change, the hash will be different.
    * *Image files cache to be determined*
- Create a new gulp task that manually replaces references to CSS/JS bundles with the ones including the hashes.
  There are rollup plugins to do this automatically, but since we are bundling CSS separately, we would need to orchestrate plugins and bundlings, so this seemed like a simple solution.

### Considerations

- The CSS bundle is still not generated with a hash, because it requires some changes in `frontend-build`'s `buildCSS` function.
  I tested it locally, and there are a couple of possible approaches for this:
    * Use a postcss plugin
    * Extend our `bundleCSS` function to accept an option to generate a hash or not.
      I prefer this option, because it basically requires adding this snippet, and a couple changes here and there:
      ```js
      import crypto from 'crypto';

      const hash = (content: string) => 
        crypto
          .createHash('sha256')
          .update(content)
          .digest('hex')
          .substr(0, 10);
      ```
- The task to update the bundle references in the index file is quite coupled with this use case.
  We could also get rid of it using other building mechanisms, but for now this is the simplest option I could find.
- The solutions applied here are perhaps too custom for this project specifically. We could consider other solutions that allow us to reuse existing tools.

### Testing steps

*TODO*

> This PR relates to https://github.com/hypothesis/frontend-shared/issues/1084

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM node:19.8.1-alpine as builder
 COPY . /frontend-shared
 RUN cd /frontend-shared && \
     yarn install --frozen-lockfile && \
-    yarn build-pattern-lib
+    NODE_ENV=production yarn build-pattern-lib
 
 
 FROM nginx:1.24.0-alpine
@@ -11,7 +11,7 @@ FROM nginx:1.24.0-alpine
 RUN rm -r /usr/share/nginx/html && rm /etc/nginx/conf.d/default.conf
 # Copy pattern library static assets, and put them in nginx document root folder
 COPY --from=builder /frontend-shared/build /usr/share/nginx/html
-COPY ./templates/index.html /usr/share/nginx/html/index.html
+COPY --from=builder /frontend-shared/templates/index.html /usr/share/nginx/html/index.html
 COPY ./images /usr/share/nginx/html/images
 COPY conf/nginx.conf /etc/nginx/conf.d/default.conf
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,7 +12,10 @@ gulp.task('serve-pattern-library', () => {
 // during development. Bundled JS and CSS are not published with the package.
 
 gulp.task('bundle-css', () =>
-  buildCSS(['./styles/pattern-library.scss'], { tailwindConfig })
+  buildCSS(['./styles/pattern-library.scss'], {
+    tailwindConfig,
+    withHash: true,
+  })
 );
 
 gulp.task(
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,4 +1,5 @@
 import { buildCSS, runTests, watchJS } from '@hypothesis/frontend-build';
+import * as fs from 'fs';
 import gulp from 'gulp';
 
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -3,15 +3,15 @@ FROM node:19.8.1-alpine as builder
 COPY . /frontend-shared
 RUN cd /frontend-shared && \
     yarn install --frozen-lockfile && \
-    yarn build-pattern-lib
+    NODE_ENV=production yarn build-pattern-lib
```
acelaya, line 6: We cannot set `ENV NODE_ENV=production` for the whole docker stage, because then `yarn install` does not install dev dependencies, which are needed for bundling.

gulpfile.js
```diff
@@ -60,3 +61,29 @@ gulp.task(
...
+
+  // Replace references to the generic bundle names with the resolved ones
+  const indexContent = fs
+    .readFileSync('templates/index.html')
+    .toString()
+    .replace('pattern-library.bundle.js', jsBundleName)
+    .replace('pattern-library.css', cssBundleName);
+
+  // eslint-disable-next-line no-undef
+  fs.writeFileSync('templates/index.html', Buffer.from(indexContent));
```
robertknight, line 88: Here `templates/index.html` has become an actual template instead of just a static file. If that is the case then I think we should make that obvious by rendering it with a template renderer when served or at build time. Assuming we're continuing to serve this as static content, then the output should go into the `build/` directory rather than modifying the source file.

robertknight: The dumbest possible template renderer is to use some obvious placeholder like `{{ js_bundle }}` and do a string search and replace all. We might as well use something like Mustache though and at least get proper escaping.

lyzadanger:
Hi, @acelaya ! Can you help me understand if this PR has any relationship or overlaps with https://github.com/hypothesis/frontend-shared/pull/1077 ?

gulpfile.js
lyzadanger: > Here templates/index.html has become an actual template instead of just a static file. If that is the case then I think we should make that obvious by rendering it with a template renderer when served or at build time. Assuming we're continuing to serve this as static content, then the output should go into the build/ directory rather than modifying the source file.

We recently renamed this file to change the mustache-specific extension to `.html` because it was static. If it's a "template again," we should once again change the source file name, methinks.

acelaya:
I have realized this can be achieved by using the existing `generateManifest` function from `frontend-build`, which brings hashing capabilities without affecting how we bundle CSS or JS, so I'm preparing another proposal which supersedes this one.

Issue #1075, opened 2023-06-01
Add reusable toast messages component


lyzadanger:
We have an upcoming need for toast messages in a new application. The `client` application already has toast messages. Extract and add a toast message component to this package.

lyzadanger:
As of now, I'm not entirely convinced we need the full toast-message functionality for this application. Perhaps. But the new `Callout` component released in v6.4.0 might do the trick.

acelaya:
Since on first pass we are creating the component directly in via, I'll remove the "video annotations" label from this ticket.

Once https://github.com/hypothesis/via/pull/1125 is merged, I will do further testing on how to use the component in `client`, and adapt where needed.

Once both projects can use this component, we can probably get back to this issue and move the component to frontend-shared.

PR #1116, opened 2023-06-21
Add `Callout` presentational component

lyzadanger:
This PR adds a new component, `Callout`, which we can use to style alert banners and toast messages. This is like toast messages without any behavior. We'll need toast messages for the video annotations app, so this is part of #1075.

Pattern-library documentation at http://localhost:4001/feedback-callout

The component styling echoes that of toast messages in the `client`. We should be able to use this component in place of [`ToastMessageItem`](https://github.com/hypothesis/client/blob/5baf673b8ec1bbdc1a5fac2b95d2749f74852d01/src/shared/components/BaseToastMessages.tsx#L29) in the `client`.

image

### DIFF
--- a/src/components/feedback/Callout.tsx
+++ b/src/components/feedback/Callout.tsx
@@ -1,0 +1,122 @@
+import classnames from 'classnames';
+import type { JSX } from 'preact';
+
+import type { PresentationalProps } from '../../types';
+import type { IconComponent } from '../../types';
+import { downcastRef } from '../../util/typing';
+import { CautionIcon, CheckIcon, CancelIcon } from '../icons';
+
+type ComponentProps = {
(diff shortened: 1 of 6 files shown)


src/components/feedback/Callout.tsx
```diff
@@ -0,0 +1,124 @@
...
+    switch (status) {
+      case 'success':
+        StatusIcon = CheckIcon;
+        break;
+      case 'error':
+        StatusIcon = CancelIcon;
+        break;
+      /* istanbul ignore next */
+      default:
+      case 'notice':
```
acelaya, line 56: What's the reason to add `case 'notice'` if we are also capturing `default` case?

acelaya (approved):
Looking good!

PR #1117, opened 2023-06-22
Allow closed state to be passed to Dialog for proper transition handling

acelaya:
This PR introduces some changes on how the `Dialog` handles closed state, so that it can be passed to it for proper transition handling, but in a way that we can dynamically render the `Dialog` itself from consumer code, if a `TransitionComponent` is not provided.

> **Note**
> Most of the code changes here are because of the introduction of a prototype page to experiment with the new capabilities in `Dialog`.

### How to use it

Up until now, we had to do something like this:

```tsx
const [closed, setClosed] = useState(true);

return (
  <>
    {!closed && (
      <Dialog
        title="Dialog"
        onClose={() => setClosed(true)}
        transitionComponent={Slider}
      >
        Hello world
      </Dialog>
    )}
  </>
);
```

This works, as long as `setClosed(true)` is not called from outside the `Dialog` itself.

[01 - broken.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/bbc123c1-994d-46b0-a333-49edd032de3e)

In order to fix that, a new `closed` prop is introduced, letting the `Dialog` itself both wrap `onClose`, but also react to changes to `closed`, and properly handle transitions in both cases:

```tsx
const [closed, setClosed] = useState(true);

return (
  <Dialog
    title="Dialog"
    closed={closed}
    onClose={() => setClosed(true)}
    transitionComponent={Slider}
  >
    Hello world
  </Dialog>
);
```

[02 - fixed.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/1f5f8818-8a60-45d9-9bd9-555f9def8926)

When no `transitionComponent` is provided, then both options above are valid.

[03 - no-transition.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/d211fa2f-c2ec-4748-9854-e71337b9352d)

> **Note**
> All examples can be tested by checking out this branch and going to http://localhost:4001/transition-components-experiments.

### State change flow

The way this works is currently a bit messy (I want to put some extra thought on it), as it requires deriving two pieces of state from the new `closed` prop.

The reason we cannot use the prop "directly" is that the prop might change sooner, but we want to delay the state to actually change, if there's a transition involved.

> These are the three pieces of state involved:
>
> * `closed`: The prop that is passed from consuming code.
> * `transitionComponentVisible`: Triggers in/out transition animations if a `TransitionComponent` is provided.
> * `isClosed`: Prevents children to be rendered when `true`. Set to `true` after `closed` is set to `true`, but with a "delay" in case of transition.

This explains how the different combinations of prop and state flows work:

#### With `TransitionComponent`

```
`closed` prop set to `true` 
└── immediately set `transitionComponentVisible` to `false`
    └── `out` animation triggers
        └── after animation ends (see `onTransitionEnd`), set `isClosed` to `true`
            └── children are not rendered anymore

`closed` prop set to `false`
└── immediately set `isClosed` to `false`
    ├── children are rendered
    └── immediately set `transitionComponentVisible` to `true`
        └── `in` animation triggers
```

#### Without `TransitionComponent`

```
`closed` prop set to `true`
└── immediately set `isClosed` to `true`
    └── children are not rendered anymore

`closed` prop set to `false`
└── immediately set `isClosed` to `false`
    └── children are rendered
```

The conclusion is that from the 4 possible combinations, only the first one justifies the need to "derive" `isClosed` state from `closed` prop.

### Considerations / thoughts / next steps

* `isClosed` feels like a too vague name for the internal state. It also might be confused with the `closed` prop.
* `close` and `isClosed` are true when not visible, while `transitionComponentVisible` is true when visible. Perhaps we should make the three consistent, maybe renaming to `transitionComponentHidden`.
* ~Removing the `if (isClosed) return null` logic simplifies the rest of the interactions, but leaves some DOM nodes for preact to handle forever.~ Scratch that. Without that check, then the `close` prop does nothing when no `transitionComponent` is provided, which would make it inconsistent.
* I think there should be a way to combine `isClosed` and `transitionComponentVisible`, but haven't nailed it yet.
* If we are planning to support transition components somewhere other than the `Dialog`, we may want to extract some of the logic, but I would start by copy-pasting once, and then it will be easier to identify what needs to be reused.

### DIFF
--- a/src/components/feedback/test/Dialog-test.js
+++ b/src/components/feedback/test/Dialog-test.js
@@ -1,5 +1,6 @@
 import { mount } from 'enzyme';
 import { createRef } from 'preact';
+import { useEffect } from 'preact/hooks';
 
 import { delay } from '../../../test-util/wait.js';
 import { testPresentationalComponent } from '../../test/common-tests';
@@ -19,7 +20,14 @@ const createComponent = (Component, props = {}) => {
  */
 const ComponentWithTransition = ({ children, direction, onTransitionEnd }) => {
   // Fake a 50ms transition time
-  setTimeout(() => onTransitionEnd?.(direction), 50);
+  useEffect(() => {
+    const timeout = setTimeout(() => onTransitionEnd?.(direction), 50);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [direction, onTransitionEnd]);
+
   return <div>{children}</div>;
 };
 
@@ -178,26 +186,37 @@ describe('Dialog', () => {
       assert.equal(document.activeElement, document.body);
     });
 
-    it('should restore focus when unmounted when `restoreFocus` set', () => {
-      const button = document.getElementById('focus-button');
+    [
+      {
+        action: 'closed',
+        updateWrapper: wrapper => wrapper.setProps({ closed: true }),
+      },
+      {
+        action: 'unmounted',
+        updateWrapper: wrapper => wrapper.unmount(),
+      },
+    ].forEach(({ action, updateWrapper }) => {
+      it(`should restore focus when ${action} when \`restoreFocus\` set`, () => {
+        const button = document.getElementById('focus-button');
 
-      // Start with focus on a button
-      button.focus();
-      assert.equal(document.activeElement, button);
+        // Start with focus on a button
+        button.focus();
+        assert.equal(document.activeElement, button);
 
-      const wrapper = mount(
-        <Dialog id="focus-dialog" title="My dialog" restoreFocus />,
-        {
-          attachTo: container,
-        },
-      );
-      const dialogElement = document.getElementById('focus-dialog');
-      // Focus moves to dialog by default when mounted
-      assert.equal(document.activeElement, dialogElement);
+        const wrapper = mount(
+          <Dialog id="focus-dialog" title="My dialog" restoreFocus />,
+          {
+            attachTo: container,
+          },
+        );
+        const dialogElement = document.getElementById('focus-dialog');
+        // Focus moves to dialog by default when mounted
+        assert.equal(document.activeElement, dialogElement);
 
-      // Unmount cleanup should restore focus to the button
-      wrapper.unmount();
-      assert.equal(document.activeElement, button);
+        // Updating should restore focus to the button
+        updateWrapper(wrapper);
+        assert.equal(document.activeElement, button);
+      });
     });
   });
 
@@ -267,9 +286,7 @@ describe('Dialog', () => {
             onClose={onClose}
             transitionComponent={ComponentWithTransition}
           />,
-          {
-            attachTo: container,
-          },
+          { attachTo: container },
         );
 
         // We simulate closing the Dialog's Panel
@@ -281,7 +298,7 @@ describe('Dialog', () => {
         // The onClose callback is not immediately invoked
         assert.notCalled(onClose);
         // Once the transition has ended, the callback should have been called
-        await delay(60); // Transition finishes after 50ms
+        await delay(70); // Transition finishes after 50ms
         assert.called(onClose);
       });
     });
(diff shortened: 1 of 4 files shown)


src/pattern-library/components/patterns/prototype/TrabsitionComponentExperimentsPage.tsx
```diff
@@ -0,0 +1,139 @@
+import { useState } from 'preact/hooks';
+
+import { Dialog } from '../../../../components/feedback';
+import { Button } from '../../../../components/input';
+import { Slider } from '../../../../components/transition';
```
lyzadanger, line 5: Minor nit, but I typically import from the package entry point in pattern library page component modules.

lyzadanger:
Thanks for putting so much thought into this! I think I generally grok where you're coming from.

> If we are planning to support transition components somewhere other than the Dialog

We definitely are, and very soon, in fact. If it helps to consider another use case: we're going to want transitions on toast messages. The new `Callout` component can be thought of as a (rough) replacement for `ToastMessageItem` in https://github.com/hypothesis/client/blob/3bf184abfe1ed828c6cc7a903969b9f1d67d0362/src/shared/components/BaseToastMessages.tsx and I am considering modeling a shared toast-messages component after the `BaseToastMessages` component here. Right now the `client` `BaseToastMessages` component applies transitions but not using a `TransitionComponent`. Might want to give this a think. Not even certain at this point whether the transition belongs on the individual toast messages or the toast-message container. Would love some thoughts and input!

acelaya:
Thanks for the feedback @lyzadanger 

@robertknight made an interesting suggestion on trying to improve the internal sync of the three pieces of state:

> Could we model the internal state as a flag indicating whether a transition is active or not?
> ```
> isClosed = closed && !transitionActive
> transitionComponentVisible = !closed || transitionActive
> ```

This was in line with one of my TODOs above: `I think there should be a way to combine isClosed and transitionComponentVisible, but haven't nailed it yet.`, but every attempt ends up breaking at least one of the combinations, so I think I will postpone this.

I'm adding a comment in the code to make sure we get reminded from time to time.

I may also try to play a bit with extracting the transition component handling into a hook that can be reused in other components:

```tsx
const { closeHandler, Wrapper } = useTransitionComponent(transitionComponent, onClose, closed, ...);
```

I think isolating that from the rest of the logic in `Dialog` will be helpful to identify how to simplify the internal state management, while we get ready for other components that need transitions.

src/components/feedback/test/Dialog-test.js
```diff
@@ -278,7 +284,7 @@ describe('Dialog', () => {
         // The onClose callback is not immediately invoked
         assert.notCalled(onClose);
         // Once the transition has ended, the callback should have been called
-        await delay(60); // Transition finishes after 50ms
+        await delay(70); // Transition finishes after 50ms
```
acelaya, line 287: This is a bit random, but the test was passing locally and failing in CI. I guess there's a chance there's a timing issue/race condition here, as we are talking about 10ms.

src/components/feedback/test/Dialog-test.js
```diff
@@ -19,7 +20,14 @@ const createComponent = (Component, props = {}) => {
...
 const ComponentWithTransition = ({ children, direction, onTransitionEnd }) => {
   // Fake a 50ms transition time
-  setTimeout(() => onTransitionEnd?.(direction), 50);
+  useEffect(() => {
+    const timeout = setTimeout(() => onTransitionEnd?.(direction), 50);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [direction, onTransitionEnd]);
```
acelaya, line 29: This stub `TransitionComponent` was triggering this timeout on every render, which was not relevant for a component that was previously mounted and unmounted every time.

With the new logic to open/close, the component can render multiple times, triggering the setTimeout an incorrect amount of times. Hence, the `useEffect`.

acelaya:
> I think there should be a way to combine isClosed and transitionComponentVisible, but haven't nailed it yet

After multiple attempts, I came to realize this is not possible, because the different state changes are not "linear", or "instant" when a transition component is provided. There are potential "delays" between the change of one piece of state and the next one.

I have slightly simplified it though, removing nested conditionals, so I think it's ready for a review again.

acelaya:
After giving this a lot of thought, we have come to the conclusion that this approach introduces too much complexity to solve a [niche use-case](https://github.com/hypothesis/client/issues/5520).

We'll probably try something else. Once of the options is using [`useImperativeHandle`](https://react.dev/reference/react/useImperativeHandle) to expose the "wrapped" close handler, so that consumers can close the dialog with the transition.

This approach would expose a bit the transition handling implementation details, but make it much simpler to maintain in the long term.

PR #1125, opened 2023-06-30
Clean up Callout exports

lyzadanger:
This tiny PR fixes a cleanup oversight when adding the new `Callout` component: clean up the exports.


acelaya (approved):
👍🏼

PR #1131, opened 2023-07-04
Extract transition component handling to useTransitionComponent hook

acelaya:
This PR extracts all the logic to manage `TransitionComponents` from `Dialog` to a reusable hook, opening the door to use it in other places with components that can expose a similar props API, while they internally use this hook.

As a good side effect, we can extract all the complexity to internally sync state to properly manage transitions, from the `Dialog` component, making it simpler.

> **Note**
> This PR is easier to review ignoring whitespaces.

### DIFF
--- a/src/hooks/use-transition-component.tsx
+++ b/src/hooks/use-transition-component.tsx
@@ -1,0 +1,102 @@
+import type { ComponentChildren } from 'preact';
+import type { JSX } from 'preact';
+import { useCallback, useEffect, useState } from 'preact/hooks';
+
+import type { TransitionComponent } from '../types';
+
+type TransitionComponentOptions = {
+  closed: boolean;
+  transitionComponent?: TransitionComponent;
+  onClose?: () => void;
+  onOpen?: () => void;
+};
+
+type TransitionComponentResult = {
+  closeHandler: () => void;
+  isClosed: boolean;
+  wrapWithTransition: (children: ComponentChildren) => JSX.Element;
+};
+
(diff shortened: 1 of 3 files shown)


src/hooks/use-transition-component.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <TransitionComp
+          direction={transitionComponentVisible ? 'in' : 'out'}
+          onTransitionEnd={onTransitionEnd}
+        >
+          {children}
+        </TransitionComp>
+      );
+    },
+    [transitionComponent, transitionComponentVisible, onTransitionEnd]
+  );
```
acelaya, line 97: My initial idea to hide the logic of conditionally wrapping the consuming component in a `TransitionComponent` or a `Fragment` was to expose an actual component:

```tsx
const Wrapper = ({ children }) => {
  if (!transitionComponent) {
    return <>{children}</>;
  }

  const TransitionComp = transitionComponent;
  return <TransitionComp ... />;
};
```

And then use it like this in the consuming component:

```tsx
const { Wrapper, ... } = useTransitionComponent(...);

return (
  <Wrapper>
    <div>...</div>
  </Wrapper>
);
```

However, that introduces a component which is defined within a component, and the effects didn't work as expected.

That's why I ended up creating a callback that does the wrapping.

acelaya:
Closing as per https://github.com/hypothesis/frontend-shared/pull/1117#issuecomment-1662100305

PR #1143, opened 2023-07-12
Update uses of `rounded` tailwind utilities to use default value correctly

lyzadanger:
This is the final PR of a set of PRs to update the use of border-radius styles in our apps.

See https://github.com/hypothesis/client/pull/5607

This PR:

* Replaces uses of unqualified `rounded` rules with their intent (e.g. `rounded-[4px]`)
* Replaces uses of `rounded-sm` with `rounded` (where it is used to imply the default rounding) and set a `DEFAULT` border-radius value in the tailwind configuration
* Clean up a few nits spotted when tweaking the `DEFAULT` border-radius setting to other values

After this lands, we can:

* Change the default border-radius of all package components by tweaking the `DEFAULT` border-radius value in the tailwind preset in this package
* Consuming apps control the default border radius for their local components AND shared components within their own local tailwind configurations, which now all have a `DEFAULT` border-radius value (which will override this package's setting).

### DIFF
--- a/src/components/feedback/Callout.tsx
+++ b/src/components/feedback/Callout.tsx
@@ -68,7 +68,7 @@ const Callout = function Callout({
       className={classnames(
         styled && 'flex items-center border',
         themed && {
-          'rounded-sm border': true,
+          'rounded border': true,
           'shadow hover:shadow-md cursor-pointer': variant === 'raised',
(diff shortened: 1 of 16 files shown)


acelaya (approved):
👍🏼

PR #1155, opened 2023-07-19
Add `tab` variant to `Tab` and support styling API

lyzadanger:
This PR adds some support we pretty much know we'll need for the import/export project by adding a new variant to `Tab` (called `tab`) and extending `Tab` to support the styling API. 

See the local [Tabs pattern library page](http://localhost:4001/navigation-tab) for details.

image

Part of #1141

Issue #1141, opened 2023-07-10
Establish tabbed-dialog pattern for import/export project


lyzadanger:
[UX mockups for the import/export project](https://www.figma.com/file/a42IFcLnKAJrpvwoMTPVME/Hypothesis---Import%2FExport?type=design&node-id=0-1&mode=design) include the notion of a tabbed dialog.

Detail here:

![Image](https://user-images.githubusercontent.com/439947/252425765-93cb7f74-946f-4526-bf08-eb4878ae1cef.png)

We need to extend `Dialog` to be able to make this happen, and other composed components used by `Dialog`.

lyzadanger:
The pattern is established. It now needs to be implemented in the `client`.

PR #1156, opened 2023-07-19
Establish import/export UI working area and prototype

lyzadanger:
This PR adds a prototype area for building out the UI for the import/export feature.

It uses a local version of the `Dialog` component with some adaptations. The idea is that we'll continue to figure out "what we need" from the `Dialog` component to support tabbed dialogs, and then eventually update the shared `Dialog` component accordingly. But it's premature to do so during this early phase.

Early prototype work can be found in [the Pattern Library](http://localhost:4001/tabbed-share-dialog)

image

Depends on #1155 
Part of #1141

### DIFF
--- a/src/components/navigation/Tab.tsx
+++ b/src/components/navigation/Tab.tsx
@@ -14,7 +14,11 @@ type ComponentProps = {
    */
   textContent?: string;
   selected?: boolean;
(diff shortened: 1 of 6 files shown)


acelaya (approved):
👍🏼

PR #1157, opened 2023-07-19
Fix slider transition event

acelaya:
This PR fixes an oversight on how the `Slider` triggers `onTransitionEnd`, which was currently being called for any CSS transition event, either on it or any children (via event bubbling).

With these changes it now only invokes `onTransitionEnd` when the event was triggered on the Slider container itself, and it happened on the `height` property (the only one animated when "opening" or "closing").

### DIFF
--- a/src/components/test/common-tests.js
+++ b/src/components/test/common-tests.js
@@ -336,7 +336,8 @@ export function testTransitionComponent(
 
         // Default to the main container as the `target` for the TransitionEvent,
         // as that would be the actual behavior at runtime.
-        // Callers can still override it in case that's not desired.
+        // If that's not desired, callers can still override it by providing
+        // event.target explicitly.
         const container = wrapper.find('div');
         container.prop('ontransitionend')({
           target: container.getDOMNode(),
--- a/src/components/test/common-tests.js
+++ b/src/components/test/common-tests.js
@@ -312,11 +312,11 @@ export function testSimpleComponent(Component) {
  * Common tests for simple design components
  *
  * @param {TransitionComponent} Component
- * @param {CommonTestOpts} opts
+ * @param {CommonTestOpts & { event?: Partial<TransitionEvent> }} opts
  */
 export function testTransitionComponent(
   Component,
-  { componentName, createContent = createComponent } = {}
+  { componentName, createContent = createComponent, event = {} } = {}
 ) {
   const displayName = componentName ?? Component.displayName ?? Component.name;
 
@@ -334,7 +334,14 @@ export function testTransitionComponent(
           onTransitionEnd,
         });
 
-        wrapper.find('div').prop('ontransitionend')();
+        // Default to the main container as the `target` for the TransitionEvent,
+        // as that would be the actual behavior at runtime.
+        // Callers can still override it in case that's not desired.
+        const container = wrapper.find('div');
+        container.prop('ontransitionend')({
+          target: container.getDOMNode(),
+          ...event,
+        });
 
         assert.calledWith(onTransitionEnd, direction);
       });
--- a/src/components/transition/Slider.tsx
+++ b/src/components/transition/Slider.tsx
@@ -53,17 +53,26 @@ const Slider: TransitionComponent = ({
     }
   }, [containerHeight, visible]);
 
-  const handleTransitionEnd = useCallback(() => {
-    if (visible) {
-      setContainerHeight('auto');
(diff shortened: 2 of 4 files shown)


src/components/test/common-tests.js
```diff
@@ -334,7 +334,10 @@ export function testTransitionComponent(
           onTransitionEnd,
         });
 
-        wrapper.find('div').prop('ontransitionend')();
+        wrapper.find('div').prop('ontransitionend')({
+          target: wrapper.find('div').getDOMNode(),
+          ...event,
+        });
```
lyzadanger, line 340: This is a little hard to human-parse. Can you help me understand the change here?

src/components/transition/Slider.tsx
```diff
@@ -1,7 +1,17 @@
...
 
+function isTransitionRelevant(
+  e: TransitionEvent,
+  containerRef: MutableRef<HTMLDivElement | null>
+): boolean {
+  // Only transitions on the actual component and for the "height" property are
+  // relevant "in"/"out" transitions.
+  return e.target === containerRef.current && e.propertyName === 'height';
+}
+
```
lyzadanger, line 14: Are you checking on the `height` property to be thorough, or was there a specific case that you can conceive of where this would matter? Are you thinking of a world where we'd apply more than one Transition to a component that takes a `TransitionComponent`? I can think of some _possible_ cases where checking the property might matter, but I'm not clear on whether it's meaningfully applicable to our real life.

I'm wondering if this might be simplified down to check for the `e.target` (only) and a simpler (commented) conditional within `handleTransitionEnd` (versus being its own function).

acelaya: > Are you checking on the height property to be thorough, or was there a specific case that you can conceive of where this would matter?

To be more explicit on what properties are relevant for the Slider's in->out or out->in change. We might at some point add other transitions, like maybe we want to change the background color on hover, and that should not trigger `onTransitionEnd`.

> I'm wondering if this might be simplified [...] a simpler (commented) conditional within handleTransitionEnd (versus being its own function).

That was the first iteration of this, and I'm thinking it probably makes sense.

I wrapped it on its own function to give it a descriptive name, but the name is actually not very descriptive on the why, and the comment inside is actually in the wrong place and should be in `handleTransitionEnd` instead.

src/components/test/common-tests.js
acelaya: Basically, `ontransitionend` now expects an event object to be passed, and will return unless the target is the wrapper's first children, and the `propertyName` is `'height'`.

First I thought I could just pass the event here, as in `wrapper.find('div').prop('ontransitionend')(event)`, but determining the target from the outside of the function call requires knowing this function implementation. That's why I `target`'s default value is `wrapper.find('div').getDOMNode()`, but you can still override it.

I could perhaps document this better with a/some comment.

src/components/test/common-tests.js
```diff
@@ -334,7 +334,14 @@ export function testTransitionComponent(
           onTransitionEnd,
         });
 
-        wrapper.find('div').prop('ontransitionend')();
+        // Default to the main container as the `target` for the TransitionEvent,
+        // as that would be the actual behavior at runtime.
+        // Callers can still override it in case that's not desired.
```
lyzadanger, line 339: How can the caller override? It looks like the next line explicitly looks for a `div`. Also, this test might break if more than on `div` is found?

acelaya: > How can the caller override?

By explicitly providing `target` on the event. Since it's unpacked afterwards, it would override it.

```ts
container.prop('ontransitionend')({
  target: container.getDOMNode(),
  ...event, // <- If `event` contains a `target` property, it will override the value from previous line
});
```

> It looks like the next line explicitly looks for a div. Also, this test might break if more than on div is found?

Yeah, I also noticed this, but that was already a problem before these changes, so I decided not to tackle that here. Instead I'll give it another thought afterwards.

What I think happened here is that we tried to provide a generic way to test TransitionComponents when we only had one implementation, so perhaps it was too soon.

lyzadanger: Can you add a comment that is explicit that you can provide an `event.target`?

lyzadanger (approved):
👍🏻

src/components/test/common-tests.js
acelaya: Yep!

Issue #1158, opened 2023-07-19
Part of tabbed-dialog pattern: tablist tab-key navigation behavior


lyzadanger:
The `Tablist` and `Tab` components in the shared package have some of the required WAI-ARIA goodies we need, but there's one piece of behavior that needs to be added.

`TabList` already supports arrow-key navigation between `Tab`s as per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role and WAI-ARIA authoring guidelines. That is, if focus is on `TabList` or its contents, left/right arrow keys (assuming the tab list is laid out in its default horizontal aspect) will move between non-disabled `Tab`s and `Enter` will select a `Tab`.

But we also need tab-key handling in `TabList`: pressing <kbd>Tab</kbd> while within a `TabList` should move focus to the `tabpanel` associated with the currently-selected `tab` in the `tablist` (which might not be the one that has focus). `tabpanel`s should not be in the tab sequence (i.e. `tabindex={-1}`) except in this case. `Tab`s should have an `aria-controls` attribute pointing at their associated `tabpanel` by ID.

lyzadanger:
I believe I have found a solution to this that doesn't require work and instead relies on the "natural" tab sequence in the document, but will leave it open until I land the changes that satisfy this.

lyzadanger:
I think I found an alternate solution to this tab sequence and focus behavior that doesn't require extra work in `TabList`.

From MDN, what should happen when <kbd>Tab</kbd> is pressed when focus is within a `tablist`:

> ...moves focus to the next element in the page tab sequence outside the tablist, which is the [active] tabpanel unless the first element containing meaningful content inside the tabpanel is focusable.

That's hard to parse. What it's saying is that focus should move to:

* The active `tabpanel` itself, if there is nothing good to focus on _inside_ of the `tabpanel`, or
* The first meaningful focusable element in the active `tabpanel`.

In our applied case, each one of the `tabpanel`s has a focusable interactive element. (If we had a `tabpanel` that didn't have a good first-focusable element, we can set `tabIndex={0}` on that `tabpanel`).

That, in combination with the structure of the tabbed-dialog I'm building, results in this just _being the natural tab sequence_ of the document, without need for JS intervention at this time. 

Press tab while in `TabList`, and you'll go to the right thing: either the active `tabpanel` or a focusable element inside of it.

[...]

PR #1160, opened 2023-07-20
Refine tabbed-dialog prototype and define component composition

lyzadanger:
This PR extends the import/export project's tabbed-dialog prototype page and starts coalescing the composition of components. It also adjusts tab sequence/navigation in a way that might address https://github.com/hypothesis/frontend-shared/issues/1158 without the need for extra work.

The prototype is still loose, but these constituent components are bubbling up:

* `TabListHeader`: Bakes the design pattern layout of a horizontal set of tabs at the top of a panel/card
* `TabPanel`: At present, combines both the design pattern of tab-panel content within a card/panel and some a11y details.
* `CloseButton`: Proposed, a styled button that I'd like to make context-aware for its click handler.
* `Divider`: Ultra-simple component, cribbed from MUI, that I'd like to use in place of border-underlines and whatnot in components in the future. 

The needed changes to `Dialog` are also solidifying:

* `variant` prop, default to `panel` (supports current usage) with a `custom` option (so that `Dialog` is not concerned with the presentation of its contents). Inelegant, but seems to allow us what we need to do here without breaking the current API. The long-term direction would be to migrate usage to `custom` across apps and then get rid of `variant`. Still percolating on this.
* Proposed: Add a new `CloseableContext` or similar that could be used by `Dialog` and potentially other components that can be "closed". Put a close handler (that takes transitions into account as needed) into this context. Context-aware things such as the proposed `CloseButton` could take this into account.

This PR doesn't need deep review (beyond what the reviewer would like to review), as it's still all prototype.

Next steps: Start building extracted components, update `Dialog`.

Part of #1141

### DIFF
--- a/src/pattern-library/components/patterns/prototype/TabbedDialogPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/TabbedDialogPage.tsx
@@ -1,23 +1,26 @@
 //import type { ComponentChildren } from 'preact';
 import classnames from 'classnames';
-import type { JSX } from 'preact';
+import type { ComponentChildren, JSX } from 'preact';
 import { useState } from 'preact/hooks';
 
 import {
   Button,
-  CopyIcon,
-  IconButton,
-  Input,
-  InputGroup,
-  TabList,
-  Tab,
-  CancelIcon,
   Card,
   CardActions,
   CardTitle,
+  IconButton,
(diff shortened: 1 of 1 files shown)


acelaya (approved):
I think your thoughts and considerations make sense in general.

PR #1162, opened 2023-07-20
Update `Dialog` and allies for import/export UI needs

lyzadanger:
This PR:

* Extends `Dialog` to allow for custom content layout (adds a `variant` prop)
* Adds a `CloseableContext` to set a close handler that can be used by descendants
* Extracts a `CloseButton` that is `CloseableContext`-aware, and uses it where applicable
* Updates import/export prototypes to use updated shared components and removes local prototype `Dialog` component

The immediate use case for `CloseButton` and the `CloseableContext` is to allow for Dialogs to have custom content that includes a close button, and ensure that that close button invokes the correct close handler for the ancestor `Dialog`. Specifically, this ensures that transitions will work. You can see some examples of this on the `Dialog` and import/export-prototype pattern-library pages.

To clarify, `CloseableContext` will need to work in _concert_ with the other change proposed for making sure transitions work when closed from "above" the `Dialog`, i.e. it doesn't _replace_ that concept.

A fair amount of the diff is pattern-library docs, but if this PR feels too big I can break it up. Suggest viewing diff without whitespace.

Next steps: Determine whether any more changes should be made to shared components (or additions) versus implementing the remaining components needed for the tabbed dialog in the `client` first, for better ergonomics.

## Testing

I suggest interacting with the demos in the pattern library pages for Dialogs, CloseButton and the import/export prototype to ensure that layouts look right and open/close works and transitions work where applicable. 

Part of #1141


src/pattern-library/components/patterns/prototype/TabbedDialogPage.tsx
```diff
@@ -207,6 +206,7 @@ function TabbedSharePanel() {
             variant="custom"
             title="Share annotations"
             onClose={() => setPanelOpen(false)}
+            transitionComponent={Slider}
```
lyzadanger, line 209: OK, so now transitions work, but adding this introduces a regression with the keyboard navigation vis-a-vis the tabs. I'll try to debug a little further.

lyzadanger, line 209: Without the fix in #1157 this introduces a regression with tab sequence and focus, but with the fix, it seems to work fine!

src/components/layout/CardHeader.tsx
```diff
@@ -45,6 +46,10 @@ const CardHeader = function CardHeader({
 
   ...htmlAttributes
 }: CardHeaderProps) {
+  const closeableContext = useContext(CloseableContext);
+  // Provided `onClose` is prioritized, but also check to see if there is
+  // a close handler provided by a `CloseableContext`
+  const closeHandler = onClose ?? closeableContext?.onClose;
```
acelaya, line 52: Considering we are now adding this logic in more than one place, and we could even have more in the future, maybe we can have a custom hook which gets the main optional `onClose`, and internally checks for the context as well.

Something in these lines.

```ts
function useOnClose(onClose?: () => void) {
  const closeableContext = useContext(CloseableContext);
  // Provided `onClose` is prioritized, but also check to see if there is
  // a close handler provided by a `CloseableContext`
  return onClose ?? closeableContext?.onClose;
}
```

And then it's used like this:

```ts
const closeHandler = useOnClose(onClose);
```

Maybe it's a bit overkill, and it also has the problem that it has a less obvious purpose. On the other hand, it would avoid explaining/commenting every time why we do that, while also avoiding potential typos in future uses where we could do `closeableContext?.onClose ?? onClose;` by mistake.

I'll leave this up to your judgement.

acelaya (approved):
I think the general approach makes sense.

I just left a couple minor comments that I'm fine if you choose to ignore.

src/pattern-library/components/patterns/prototype/TabbedDialogPage.tsx
acelaya: The fix is now merged, so feel free to rebase 🙂

src/components/layout/CardHeader.tsx
lyzadanger: I think this is worth considering. Do you think we could perhaps extend on this current work (i.e. later) to add that hook? I already have a TODO to track some follow-up, because it's likely we'd want to make `Panel` "closeable" (not just `Dialog`).

acelaya: Yep, absolutely.

PR #1163, opened 2023-07-21
Make testTransitionComponent helper more resilient and future-proof

acelaya:
During the review of https://github.com/hypothesis/frontend-shared/pull/1157 [it was spotted](https://github.com/hypothesis/frontend-shared/pull/1157#discussion_r1269352825) that `testTransitionComponent` helper was a bit coupled with our only existing `TransitionComponent`, the `Slider`.

The problem was that, in order to test the `onTransitionEnd` callback is invoked, we were looking for a `div` element, which is ok for the `Slider`, but there's no way to know if that will be true for future implementations:

* There could be no divs at all.
* There could be more than one div.
* There could be one div, but without an `ontransitionend` prop.

This PR tries to be more resilient, and use the `[ontransitionend]` selector when looking for an element with the `ontransitionend` prop.

There's still a a chance that some future `TransitionComponent` implementation handles transitions in a different way, but it's hard to know when we only have one implementation at the moment.

At least this approach makes less assumptions about the internal DOM structure of the component, making it a bit more resilient.


lyzadanger (approved):
Might I suggest an approach that builds on this a little?

`CommonTestOpts` accept an `elementSelector` that is used to select the component's "main" element if it is isn't the (default) wrapping element. You might consider extending the options accepted by transition tests to include a `transitionElementSelector`.

The default you have proposed here is a fine default (select for `[transitionend]` unless `transitionElementSelector` is provided). So you could do this separately.

acelaya:
Aha! Good idea.

src/components/test/common-tests.js
```diff
@@ -334,12 +339,12 @@ export function testTransitionComponent(
...
-        // as that would be the actual behavior at runtime.
-        // If that's not desired, callers can still override it by providing
-        // event.target explicitly.
-        const container = wrapper.find('div');
+        const container = wrapper.find(elementSelector);
         container.prop('ontransitionend')({
+          // Default to the main container as the `target` for the TransitionEvent,
+          // as that would be the actual behavior at runtime.
+          // If that's not desired, callers can still override it by providing
+          // event.target explicitly.
```
acelaya, line 347: This comment was misplaced. It references the setting of the `target` prop, so I have moved it down here.

acelaya:
> CommonTestOpts accept an elementSelector that is used to select the component's "main" element if it is isn't the (default) wrapping element. You might consider extending the options accepted by transition tests to include a transitionElementSelector.

@lyzadanger the `testTransitionComponent` already uses `CommonTestOpts`, so I have just used `elementSelector`. I think it makes it more consistent with other test anyway.

lyzadanger (approved):
LGTM!

PR #1169, opened 2023-07-24
Ensure `CloseButton` always rendered 16px square in panel/dialog headers

lyzadanger:
The recent changes that included `CloseButton` caused a small visual regression in `SidebarPanel`s in the client, to wit, they are too small:

image

This is because `IconButton`s size their icons at 1em square, and the local font size here in the client is only `13px` (versus the pattern library's more sane 16px base font size).

Right now, the _intent_ is that any time we render a close button in a header (for a panel or dialog), it should always be exactly 16x16, and should not adapt to local font size.

I've made this the case, and also added a little example for `CardHeader` that would help us spot future regressions before they happen.

After these changes, `client` against build of this package:

image

### DIFF
--- a/src/components/layout/CardHeader.tsx
+++ b/src/components/layout/CardHeader.tsx
@@ -74,7 +74,11 @@ const CardHeader = function CardHeader({
           classes={classnames(
             // Pull button right such that its icon right-aligns with the
             // header's bottom border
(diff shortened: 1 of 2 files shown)


acelaya (approved):
👍🏼

PR #1196, opened 2023-08-16
Add feedback prop to input to support dynamic styling

acelaya:
This PR adds a new `feedback` prop to `Input`, that can have values `error` or `warning`, and is used to dynamically style the input to provide extra feedback.

> **Note**
> I named it `feedback`, after checking how other UI libs treat this. Feedback is how it's named in bootstrap, and it sounds intuitive enough, and allows for more values to be eventually added (`success`, `info`, etc)

image

It also deprecates and replaces `hasError`, recommending `feedback"error"` as a direct replacement. 

image

> **Note**
> The `Select` was left out of this PR and will be added afterwards, once we have discussed and agreed on the right contract.

### DIFF
--- a/src/components/input/InputRoot.tsx
+++ b/src/components/input/InputRoot.tsx
@@ -7,6 +7,11 @@ import { inputGroupStyles } from './InputGroup';
 
 type RootComponentProps = {
   element?: 'input' | 'select';
+  feedback?: 'error' | 'warning';
+
+  /**
+   * @deprecated Use feedback="error" instead
+   */
   hasError?: boolean;
 };
 
@@ -27,9 +32,14 @@ const InputRoot = function InputRoot({
   classes,
   elementRef,
   hasError,
+  feedback,
 
   ...htmlAttributes
 }: InputRootProps) {
+  if (feedback === undefined && hasError) {
+    feedback = 'error';
+  }
+
   if (!htmlAttributes.id && !htmlAttributes['aria-label']) {
     console.warn(
       '`Input` component should have either an `id` or an `aria-label` attribute',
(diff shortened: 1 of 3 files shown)


src/components/input/InputRoot.tsx
```diff
@@ -46,7 +52,11 @@ const InputRoot = function InputRoot({
         'focus-visible-ring ring-inset border rounded w-full p-2',
         'bg-grey-0 focus:bg-white disabled:bg-grey-1',
         'placeholder:text-color-grey-5 disabled:placeholder:color-grey-6',
-        { 'ring-inset ring-2 ring-red-error': hasError },
+        {
+          'ring-inset ring-2': !!feedback || hasError,
+          'ring-red-error': feedback === 'error' || hasError,
+          'ring-yellow-notice': feedback === 'warning',
```
acelaya, line 58: `yellow-notice` is the same color we use for notice toast messages.

src/components/input/InputRoot.tsx
```diff
@@ -27,9 +32,12 @@ const InputRoot = function InputRoot({
   classes,
   elementRef,
   hasError,
+  feedback,
 
   ...htmlAttributes
 }: InputRootProps) {
+  const inputFeedback = feedback ?? (hasError ? 'error' : undefined);
```
robertknight, line 39: You could also just re-assign `feedback` here since arguments are mutable.

src/pattern-library/components/patterns/input/InputPage.tsx
```diff
@@ -69,11 +69,44 @@ export default function InputPage() {
...
+                  aria-label="Input with error"
+                  feedback="error"
+                  value="something invalid"
+                />
+              </div>
+              <div className="w-[350px]">
+                <Input
+                  aria-label="Input with warning"
+                  feedback="warning"
+                  value="something invalid"
```
robertknight, line 100:
```suggestion
                  value="might be a problem"
```

If the input was outright invalid, the "error" state should be used. The warning state is used when the input might cause a problem, but we're leaving it up to the user to decide to continue or change it.

robertknight:
> I thought feedback was more concise than color, and also prefer to have just one way to do something.

If we only support displaying one feedback state at a time, then indeed a single prop seems a better way to model this.

robertknight (approved):

src/components/input/InputRoot.tsx
acelaya: Hah, I didn't think about that. I'm too used to not reassigning stuff 😅

PR #1216, opened 2023-08-29
Update package version to 6.6.0

acelaya:


src/components/input/InputRoot.tsx
```diff
@@ -37,6 +37,7 @@ const InputRoot = function InputRoot({
 
   ...htmlAttributes
 }: InputRootProps) {
+  /* istanbul ignore next */
```
acelaya, line 40: Ignoring for now, but we need to discuss how to test components which props only apply classes dynamically, without coupling tests with tailwind.

PR #1222, opened 2023-09-07
Prompts confirm

acelaya:
Bringing the `confirm` helper function from client to this library, as we have use cases for it in other projects.

I brought it almost verbatim, with the only addition of a customizable cancel action text.

I also added a simple documentation piece in the patterns library, under `Dialogs`.

Captura desde 2023-09-07 10-30-45

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,6 +5,9 @@ export { useFocusAway } from './hooks/use-focus-away';
 export { useKeyPress } from './hooks/use-key-press';
 export { useSyncedRef } from './hooks/use-synced-ref';
 
+// Utils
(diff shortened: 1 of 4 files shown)


robertknight (approved):
LGTM

acelaya:
> Just a note on PR titles, I suggest to word them in the imperative mood as is recommended for the first line of git commit messages. eg. "Add `confirm` function to show modal prompt". This makes it easier to see what the PR is about at a glance.

I almost always send single-commit PRs, which means the PR title is generated from the commit message.

That made me forget to verify it here before submitting. My bad 😅

PR #1224, opened 2023-09-08
Allow initial focus to be provided to confirm prompt

acelaya:
This PR extends the options for `confirm`, including one to allow defining which of the two buttons should be initially focused.

By default it keeps focusing the "cancel" button, as that's probably the non-destructive option in most of the cases.

### DIFF
--- a/src/util/prompts.tsx
+++ b/src/util/prompts.tsx
@@ -1,6 +1,5 @@
 import { render } from 'preact';
 import { createRef } from 'preact';
-import type { RefObject } from 'preact';
 import type { ComponentChildren } from 'preact';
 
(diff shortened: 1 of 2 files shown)


acelaya:
I'm checking the failing test. I have reproduced it locally by running it a couple of times, so I'm looking for a different way to test it.

robertknight:
> EDIT 2: I found what I think is a more consistent way to wait for the button to be focused by adding a focus event listener to it.

If there is a specific event that you can wait for rather than a timeout, that is almost always preferable.

robertknight (approved):
LGTM

PR #1237, opened 2023-09-12
Create enhanced select prototype page

acelaya:
Experimentation to build an accessible select that supports custom UI on options dropdown.

Source of inspiration: https://headlessui.com/react/listbox

### TODO

- [ ] Decide on the best name (`EnhancedSelect`, `Listbox`...)
- [x] Handle arrow key navigation
- [x] Close on click away
- [ ] Define the right roles on elements (looks like a combination of `aria-haspopup="listbox"` + `role="listbox"` on the options container + `role="option"` on every option should do the trick. Using `Button` for the items also provides a lot of accessibility out of the box)
- [ ] Decide if `Option` should be an independent component, or one that can be used as `<Select.Option />`. The former needs to be reviewed for TS to understand it.
- [ ] Make the dropdown drop up when there's more available space.
- [ ] Decide on the best way to style a "selected" option.


robertknight:
> Decide on the best name 

Will we actually want to use the unstyled `<select>` anywhere after this? If the majority of uses are going to be the new control, then we could just call it `Select`. If we need an option not to have any custom styling, perhaps that could just be a `variant`? Internally it might make sense to switch between different implementation components depending on the variant.

robertknight:
The video in the description appears corrupted to me, in Safari and Chrome.

acelaya:
> Will we actually want to use the unstyled <select> anywhere after this? If the majority of uses are going to be the new control, then we could just call it Select. If we need an option not to have any custom styling, perhaps that could just be a variant? Internally it might make sense to switch between different implementation components depending on the variant.

That's actually not a bad idea

> The video in the description appears corrupted to me, in Safari and Chrome.

Just uploaded another one

acelaya:
> Will we actually want to use the unstyled `<select>` anywhere after this? If the majority of uses are going to be the new control, then we could just call it `Select`. If we need an option not to have any custom styling, perhaps that could just be a `variant`? Internally it might make sense to switch between different implementation components depending on the variant.

That's actually not a bad idea

> The video in the description appears corrupted to me, in Safari and Chrome.

Just uploaded another one

acelaya:
Based on the suggestion above, one way to support filtering items would be introducing a new `<Select.Filterable />` wrapper for `<Select.Option />` items, which allows children to be filtered via context.

```tsx
function App() {
  const [selected, setSelected] = useState<(typeof items)[number]>();

  return (
    <Select
      selected={selected}
      onChange={setSelected}
      label={
        selected ? (
          <>{item.name} ({item.id})</>
        ) : (
          <>Select one...</>
        )
      }
    >
      <Select.Filterable onFilter={(query: string, value: T) => boolean}>
        {items.map(item => (
          <Select.Option value={item} key={item.id}>
            {({ disabled, isSelected }) => <>{item.name} ({item.id})</>}
          </Select.Option>
        ))}
      </Select.Filterable>
    </Select>
  )
}
```

Then we can expose the `onFilter` callback via context, down to the `Option` entries, which can decide if they should render or not, based on that.

The `Filterable` component could be something in these lines:

```tsx
const FilterableContext = createContext<{ shouldRender: (value: T) => boolean } | null>(null);

function Filterable<T>({ onFilter, children }) {
  const [query, setQuery] = useState('');
  const shouldRender = useCallback(
    (value: T) => onFilter(query, value), 
    [onFilter, query],
  );

  return (
    <FilterableContext.Provider value={{ shouldRender }}>
      {children}
    </FilterableContext.Provider>
  );
}
```

robertknight:
I left some general notes on the visual design in the Hypothesis Reading group, copied below.

[2023-09-20-Sketches_ Enhanced select_ Hypothesis Component Li-Hypothesis-Reading.json.zip](https://github.com/hypothesis/frontend-shared/files/12670088/2023-09-20-Sketches_.Enhanced.select_.Hypothesis.Component.Li-Hypothesis-Reading.json.zip)

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,182 @@
...
+    },
+    [closeDropdown, onChange],
+  );
+
+  // When clicking away or pressing `Esc`, close the dropdown
+  useClickAway(wrapperRef, closeDropdown);
+  useKeyPress(['Escape'], closeDropdown);
+
+  // Vertical arrow key for options in the dropdown
+  useArrowKeyNavigation(wrapperRef, { horizontal: false });
```
robertknight, line 55: In the native select, keyboard arrows only loop over items within the popup, not the popup + current value display. Also in the native select, keyboard selection doesn't loop when you get to the top/bottom. I'd check the WAI-ARIA specs to see what they say about this.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  // Focus button after closing dropdown
+  useLayoutEffect(() => {
+    if (!isDropdownOpen) {
+      buttonRef.current!.focus();
+    }
+  }, [isDropdownOpen]);
+
+  if (isDropdownOpen) {
+    const viewportHeight = window.innerHeight;
```
robertknight, line 65: This reads the DOM, so belongs in a layout effect.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const { top: buttonDistanceToTop, height } =
+      buttonRef.current!.getBoundingClientRect();
```
robertknight, line 67: Reading the ref here happens to work because the `isDropdownOpen` state happens to always be false on the initial render, but it would be easier to break this assumption with future changes.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const buttonDistanceToBottom =
+      viewportHeight - (buttonDistanceToTop + height);
+
+    // The listbox should drop up only if there's not enough space below it for
+    // the listbox max height, and there's also more space above
+    setShouldDropUp(
+      // 320px is tailwind's h-80. e should try not to couple with this
+      buttonDistanceToBottom < 320 &&
```
robertknight, line 75: Where does the 320px (or `h-80` equivalent) come from? In my testing I found that the dropdown was dropping up even in cases where it looked like there was enough room that it should drop down.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,182 @@
...
+        id={buttonId}
+        variant="custom"
+        classes={classnames(
+          'w-full flex border',
+          'bg-grey-0 disabled:bg-grey-1 disabled:text-grey-6',
+        )}
+        expanded={isDropdownOpen}
+        pressed={isDropdownOpen}
+        disabled={disabled}
+        aria-haspopup="listbox"
```
robertknight, line 92: For me VoiceOver is reading the button as a toggle button, rather than a toggle button that controls a listbox. This might be because the listbox is not rendered, rather than merely hidden, until the dropdown is opened.

src/pattern-library/components/patterns/prototype/EnhancedSelectPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+      }
+    >
+      <Library.Section>
+        <Library.Pattern title="Select">
+          <Select_ />
+        </Library.Pattern>
+
+        <Library.Pattern title="Select with many options">
+          <Select_
+            theItems={[
```
robertknight: Can you use distinct items for this example. Just appending a number to the end would be enough. Having the exact same values repeated is a bit confusing.

src/pattern-library/components/patterns/prototype/EnhancedSelectPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+              ...items,
+              ...items,
+              ...items,
+              ...items,
+              ...items,
+            ]}
+          />
+        </Library.Pattern>
+
+        <Library.Pattern title="Plain text Select">
```
robertknight: I would call the plain text select just "Select" and the example with badges "Select (custom items)" or similar.

robertknight (commented):
I took a look over this and left some initial notes. A few high level comments:

- General component structure / use of context etc. looks fine
- The listbox toggle button is read as a plain toggle button rather than as one that controls a listbox in VoiceOver. I suspect this may be because the listbox is not rendered when hidden?
- Keyboard interactions didn't quite work the way I expected. See notes.
- The code to choose whether to drop up or drop down was dropping up even when there seemed to be quite a lot of space free at the bottom of the screen

General code notes:

- `selected` sounds like a boolean. I'd use `value` (or maybe `currentValue` internally if you need to disambiguate) for the currently selected value

src/components/input/SelectNext.tsx
acelaya: Yeah, there's room for improvement here.

Since the dropdown is not even rendered, it is not possible to infer how much size it is going to need, so what I did is use the maximum space it could take as reference (notice the listbox has `max-h-80`).

I tried to see how other libraries do this, but since they support other use cases, their code is more complex, and it's not easy to extract the piece solving this specific problem.

But yeah, the main problem with this is that it calculates based on the maximum size, which means, if the listbox is actually smaller, it can end up dropping up even if it would have been able to fit down.

I'll put a bit more thought on this logic. Perhaps moving the logic to `useLayoutEffect` as you pointed out in other comments, ends up solving this as well, but I want to avoid a listbox that "blinks" down and is immediately repositioned after. I'm trying to make it directly appear either up or down.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const selectContext = useContext(SelectContext);
+  if (!selectContext) {
+    throw new Error('Select.Option can only be used as Select child');
+  }
+
+  const { selectValue, selected } = selectContext;
+  const isSelected = !disabled && selected === value;
+
+  return (
+    <Button
```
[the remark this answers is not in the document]
acelaya: Mainly, that you get click, focus and keyboard interaction working out of the box.

Other libraries, like reactstrap, use this approach as well when rendering items that you need to be able to interact with.

For example, their dropdown https://reactstrap.github.io/?path=/docs/components-dropdown--dropdown. If you inspect the page you'll see both the button itself and all the menu items are `<button />` tags.

src/components/input/SelectNext.tsx
acelaya: I can confirm that using `useLayoutEffect` + toggling via `display: none` instead of via unmount, allows for the actual height of the listbox to be calculated, making this more accurate.

Also, hidding via CSS should also help addressing https://github.com/hypothesis/frontend-shared/pull/1237#discussion_r1331205890, but this needs to be confirmed.

src/components/input/SelectNext.tsx
acelaya: > in the native select, keyboard selection doesn't loop when you get to the top/bottom. I'd check the WAI-ARIA specs to see what they say about this.

This is the default behavior on `useArrowKeyNavigation`. I think it should be possible to add a new optional `loop` option to determine its behavior, but I would like to tackle that separately and not as part of this.

I will try to address the point in which the select itself is not part of the loop.

acelaya: > In the native select, keyboard arrows only loop over items within the popup, not the popup + current value display

If I exclude the toggle button, then `useArrowKeyNavigation` has a weird behavior that makes the second item to be the first one to be focused when pressing arrow down.

I have realized all existing use cases for `useArrowKeyNavigation` are working on lists of elements that are visible and mounted, so it's possible we need to tweak it for this to work the right way.

For now I will keep the toggle in the list of items to iterate, and create two separated issues to address the optional infinite loop and making sure the first list item is the one selected when pressing arrow down for the first time.

src/components/input/SelectNext.tsx
acelaya: I can confirm orca has the same behavior, identifying the button as a toggle.

If I change the listbox to be hidden instead of unmounted, then it identifies it properly as a listbox handler.

acelaya:
@robertknight I have addressed most of the comments and resolved the conversations above.

I only left two conversations open as they need more feedback.

Since most of the decisions have been aligned, I will go ahead and start writing a proper documentation page that will replace the prototype one, and create new follow-up issues for things that need to be addressed afterwards.

It's also very likely I close this PR and create a cleaner new one referencing this one, before merging anything.

acelaya:
Replaced by https://github.com/hypothesis/frontend-shared/pull/1247

Issue #1223, opened 2023-09-07
Build a custom select that allows for UI customization in the dropdown


acelaya:
As part of https://github.com/hypothesis/client/issues/5775, we have been presented with the need to have a `Select` component which dropdown is not a native list of `<option />` elements, but something more customizable (see https://github.com/hypothesis/client/issues/5775#issuecomment-1698491323).

The component needs to allow custom UI per option, as well as eventually support filtering, limit the amount of filtered results, etc.

It should also be as accessible as possible (keyboard navigation via Tab and arrow keys, proper option expand/contract and focusing, etc).

PR #1247, opened 2023-09-21
Create SelectNext component replacing Select

acelaya:
Create a new `SelectNext` component, which is meant to replace existing `Select` one.

The benefits it presents include being able to customize the listbox UI, and supporting complex objects as values.

`SelectNext` is a controlled component designed to be used like this:

```tsx
import { SelectNext } from '@hypothesis/frontend-shared';

const items = [
  { id: 1, name: 'All students' },
  { id: 2, name: 'Albert Banana' },
  { id: 3, name: 'Bernard California' },
  { id: 4, name: 'Cecelia Davenport' },
  { id: 5, name: 'Doris Evanescence' },
];

function App() {
  const [value, setValue] = useState<(typeof items)[number]>();

  return (
    <SelectNext
      selected={value}
      onChange={setValue}
      label={
        value ? (
          <>{value.name} ({value.id})</>
        ) : (
          <>Select one...</>
        )
      }
    >
      {items.map(item => (
        <SelectNext.Option value={item} key={item.id}>
          {({ disabled, isSelected }) => <>{item.name} ({item.id})</>}
        </SelectNext.Option>
      ))}
    </SelectNext>
  )
}
```

Existing `Select` is now marked as deprecated, and the plan is to remove it on next major release, while we also rename `SelectNext` to plain `Select`.

### TODO

- [x] Add test
- [x] Fix a11y issues highlighted when testing Select

### Follow up improvements

There are a couple of things we can improve, but it would be better to tackle separately to avoid a PR bigger than it already is:

- Handle too long content in options or select (overflow ellipsis?).
- Prevent looping back to top/bottom in arrow keys navigation when reaching the end of the list. [See comment](https://github.com/hypothesis/frontend-shared/pull/1237#discussion_r1331767946).
- Exclude the select toggle button from the arrow key navigation sequence. [See comment](https://github.com/hypothesis/frontend-shared/pull/1237#discussion_r1332634012).
- Close listbox when focus moves to some other component (make sure the toggle button is not focused in this case).

Nice to have:

- Support searching/filtering: See technical proposal https://github.com/hypothesis/frontend-shared/pull/1237#issuecomment-1723436418
- Support `feedback="error"` and `feedback="warning"` like the existing Select does.

> A big chunk of this PR is the new pattern library documentation page that you can see by checking out this branch and going to http://localhost:4001/input-select-next

Closes https://github.com/hypothesis/frontend-shared/issues/1223

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -1,0 +1,198 @@
+import classnames from 'classnames';
+import type { ComponentChildren } from 'preact';
+import {
+  useCallback,
+  useContext,
+  useId,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'preact/hooks';
+
+import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
+import { useClickAway } from '../../hooks/use-click-away';
+import { useKeyPress } from '../../hooks/use-key-press';
+import { useSyncedRef } from '../../hooks/use-synced-ref';
+import type { PresentationalProps } from '../../types';
+import { MenuCollapseIcon, MenuExpandIcon } from '../icons';
+import Button from './Button';
+import SelectContext from './SelectContext';
+
+export type SelectOptionStatus = {
+  selected: boolean;
+  disabled: boolean;
+};
+
+export type SelectOptionProps<T> = {
+  value: T;
+  disabled?: boolean;
+  children: (status: SelectOptionStatus) => ComponentChildren;
+  classes?: string | string[];
+};
+
+function SelectOption<T>({
+  value,
+  children,
+  disabled = false,
+  classes,
+}: SelectOptionProps<T>) {
+  const selectContext = useContext(SelectContext);
+  if (!selectContext) {
+    throw new Error('Select.Option can only be used as Select child');
+  }
+
+  const { selectValue, value: currentValue } = selectContext;
+  const selected = !disabled && currentValue === value;
+
+  return (
+    <Button
+      variant="custom"
+      classes={classnames(
+        'w-full ring-inset rounded-none !p-0',
+        'border-t first:border-t-0 bg-transparent',
+        { 'hover:bg-grey-1': !disabled },
+        classes,
+      )}
+      onClick={() => selectValue(value)}
+      role="option"
+      disabled={disabled}
+      aria-selected={selected}
+      // This is intended to be focused with arrow keys
+      tabIndex={-1}
+    >
+      <div
+        className={classnames('flex w-full p-1.5 border-l-4', {
+          'border-l-transparent': !selected,
+          'border-l-brand font-medium': selected,
+        })}
+      >
+        {children({ selected, disabled })}
+      </div>
+    </Button>
+  );
+}
+
+export type SelectProps<T> = PresentationalProps & {
+  value: T;
+  onChange: (newValue: T) => void;
+  label: ComponentChildren;
+  disabled?: boolean;
+};
+
+function SelectMain<T>({
+  label,
+  value,
+  onChange,
+  children,
+  disabled,
+  classes,
+  elementRef,
+}: SelectProps<T>) {
+  const [listboxOpen, setListboxOpen] = useState(false);
+  const [shouldListboxDropUp, setShouldListboxDropUp] = useState(false);
+  const closeListbox = useCallback(() => setListboxOpen(false), []);
+  const wrapperRef = useRef<HTMLDivElement | null>(null);
+  const listboxRef = useRef<HTMLDivElement | null>(null);
+  const listboxId = useId();
+  const buttonRef = useSyncedRef(elementRef);
+  const buttonId = useId();
+
+  const selectValue = useCallback(
+    (newValue: unknown) => {
+      onChange(newValue as T);
+      closeListbox();
+    },
+    [closeListbox, onChange],
+  );
+
+  // When clicking away or pressing `Esc`, close the listbox
+  useClickAway(wrapperRef, closeListbox);
+  useKeyPress(['Escape'], closeListbox);
+
+  // Vertical arrow key for options in the listbox
+  useArrowKeyNavigation(wrapperRef, { horizontal: false });
+
+  useLayoutEffect(() => {
+    if (!listboxOpen) {
+      // Focus button after closing listbox
+      buttonRef.current!.focus();
+      // Reset shouldDropUp so that it does not affect calculations next time
+      // it opens
+      setShouldListboxDropUp(false);
+    } else {
+      const viewportHeight = window.innerHeight;
+      const { top: buttonDistanceToTop, bottom: buttonBottom } =
+        buttonRef.current!.getBoundingClientRect();
+      const buttonDistanceToBottom = viewportHeight - buttonBottom;
+      const { bottom: listboxBottom } =
+        listboxRef.current!.getBoundingClientRect();
+      const listboxDistanceToBottom = viewportHeight - listboxBottom;
+
+      // The listbox should drop up only if there's not enough space below to
+      // fit it, and there's also more absolute space above than below
+      setShouldListboxDropUp(
+        listboxDistanceToBottom < 0 &&
+          buttonDistanceToTop > buttonDistanceToBottom,
+      );
+    }
+  }, [buttonRef, listboxOpen]);
+
+  return (
+    <div className="relative" ref={wrapperRef}>
+      <Button
+        id={buttonId}
+        variant="custom"
+        classes={classnames(
+          'w-full flex border',
+          'bg-grey-0 disabled:bg-grey-1 disabled:text-grey-6',
+          classes,
+        )}
+        expanded={listboxOpen}
+        disabled={disabled}
+        aria-haspopup="listbox"
+        aria-controls={listboxId}
+        elementRef={buttonRef}
+        onClick={() => setListboxOpen(prev => !prev)}
+        onKeyDown={e => {
+          if (e.key === 'ArrowDown' && !listboxOpen) {
+            setListboxOpen(true);
+          }
+        }}
+        data-testid="select-toggle-button"
+      >
+        {label}
+        <div className="grow" />
+        {listboxOpen ? <MenuCollapseIcon /> : <MenuExpandIcon />}
+      </Button>
+      <SelectContext.Provider value={{ selectValue, value }}>
+        <div
+          className={classnames(
+            'absolute z-5 w-full max-h-80 overflow-y-auto',
+            'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
+            {
+              'top-full mt-1': !shouldListboxDropUp,
+              'bottom-full mb-1': shouldListboxDropUp,
+
+              // Hiding instead of unmounting to
+              // * Ensure screen readers detect button as a listbox handler
+              // * Listbox size can be computed to correctly drop up or down
+              hidden: !listboxOpen,
+            },
+          )}
+          role="listbox"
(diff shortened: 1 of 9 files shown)


src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,195 @@
...
+      </Button>
+      <SelectContext.Provider value={{ selectValue, value }}>
+        <div
+          className={classnames(
+            'absolute z-5 w-full max-h-80 overflow-y-auto',
+            'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
+            {
+              'top-full mt-1': !shouldListboxDropUp,
+              'bottom-full mb-1': shouldListboxDropUp,
+              hidden: !isListboxOpen,
```
acelaya, line 176: There are two main reasons to hide the listbox rather than unmounting it:

* Screen readers can identify the toggle button as a listbox handler (see `aria-haspopup="listbox"` and `aria-labelledby={buttonId}` on the listbox itself)
* It allows to properly calculate the height the listbox is going to need, in order to determine if it should drop down or drop up.

robertknight:
Some accessibility things I noticed in testing with VoiceOver:

1. The toggle button is not identified as a control that shows a listbox, but just as an ordinary toggle button.
2. When pressing tab while the listbox is open, focus moves elsewhere in the document, but the popover remains visible

I think the general pattern that we want to follow here is the [Select-Only combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/). In addition to the two items mentioned above, there are other details of keyboard handling etc. in that document.

In my testing, setting `role=combobox` on the toggle button was sufficient to resolve item (1).

This PR is quite large now so I suggest we resolve that, and then create a follow-up issue for the remaining items.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,195 @@
...
+function SelectMain<T>({
+  label,
+  value,
+  onChange,
+  children,
+  disabled,
+  classes,
+  elementRef,
+}: SelectProps<T>) {
+  const [isListboxOpen, setIsListboxOpen] = useState(false);
```
robertknight, line 91:
```suggestion
  const [listboxOpen, setListboxOpen] = useState(false);
```

I normally only add "is" in the names of the boolean-ness is not obvious enough from the name otherwise.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const selectValue = useCallback(
+    (newValue: unknown) => {
+      onChange(newValue as T);
+      closeListbox();
+    },
+    [closeListbox, onChange],
+  );
+
+  // When clicking away or pressing `Esc`, close the listbox
```
robertknight, line 108: The listbox should also be closed if tabbing away. See notes in https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,195 @@
...
+  useClickAway(wrapperRef, closeListbox);
+  useKeyPress(['Escape'], closeListbox);
+
+  // Vertical arrow key for options in the listbox
+  useArrowKeyNavigation(wrapperRef, { horizontal: false });
+
+  useLayoutEffect(() => {
+    if (!isListboxOpen) {
+      // Focus button after closing listbox
+      buttonRef.current!.focus();
```
robertknight, line 118: Note that we shouldn't focus the button if closing the listbox happened as a result of tabbing away.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,195 @@
...
+          if (e.key === 'ArrowDown' && !isListboxOpen) {
+            setIsListboxOpen(true);
+          }
+        }}
+        data-testid="select-toggle-button"
+      >
+        {label}
+        <div className="grow" />
+        {isListboxOpen ? <MenuCollapseIcon /> : <MenuExpandIcon />}
+      </Button>
```
robertknight, line 167: As noted in comments, this button isn't identified as controlling a popup listbox in VoiceOver at least (not sure about other screen readers). Marking the button as being `role=combobox` resolved this.

src/components/input/SelectNext.tsx
robertknight: This makes sense. Perhaps add a short version of this code comment in the code itself.

src/components/input/SelectNext.tsx
```diff
@@ -0,0 +1,195 @@
...
+              'top-full mt-1': !shouldListboxDropUp,
+              'bottom-full mb-1': shouldListboxDropUp,
+              hidden: !isListboxOpen,
+            },
+          )}
+          role="listbox"
+          ref={listboxRef}
+          id={listboxId}
+          aria-labelledby={buttonId}
+          aria-orientation="vertical"
```
robertknight, line 183: Do we need an explicit `aria-orientation` here, or is vertical the default?

src/components/input/test/SelectNext-test.js
```diff
@@ -0,0 +1,220 @@
...
+
+    assert.isFalse(isListboxClosed(wrapper));
+  });
+
+  [
+    { windowHeight: 10000, containerPaddingTop: 0, shouldDropUp: false },
+    { windowHeight: 200, containerPaddingTop: 1000, shouldDropUp: true },
+  ].forEach(({ windowHeight, containerPaddingTop, shouldDropUp }) => {
+    it('makes listbox drop up or down based on available space below', () => {
+      window.innerHeight = windowHeight;
```
robertknight, line 183: Huh, I'm slightly surprised that setting `window.innerHeight` works. This could potentially impact other tests, so we might want to reset it afterwards, or add a test seam.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -0,0 +1,344 @@
...
+  { id: '2', name: 'Albert Banana' },
+  { id: '3', name: 'Bernard California' },
+  { id: '4', name: 'Cecelia Davenport' },
+  { id: '5', name: 'Doris Evanescence' },
+];
+
+function Select_({
+  disabled,
+  textOnly,
+  theItems = items,
```
robertknight, line 20: I think this might work without the `theItems` name here, since the left side is a property name and the right side is a different context. In any case, I would probably prefer to give the global variable the more distinct name than the local variable to avoid a conflict.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+const items = [
+  { id: '1', name: 'All students' },
+  { id: '2', name: 'Albert Banana' },
+  { id: '3', name: 'Bernard California' },
+  { id: '4', name: 'Cecelia Davenport' },
+  { id: '5', name: 'Doris Evanescence' },
+];
+
+function Select_({
```
robertknight, line 17: Maybe call this `SelectExample` or something, to make the difference with the imported `Select` more obvious.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -0,0 +1,344 @@
...
+              </>
+            )
+          }
+        </Select.Option>
+      ))}
+    </Select>
+  );
+}
+
+function InputGroupSelect_() {
```
robertknight, line 71: Maybe `InputGroupSelectExample`?

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -0,0 +1,344 @@
...
+  const [selected, setSelected] = useState<(typeof items)[number]>();
+  const selectedIndex = useMemo(
+    () => (!selected ? -1 : items.findIndex(item => item === selected)),
+    [selected],
+  );
+  const next = useCallback(() => {
+    const newIndex = selectedIndex + 1;
+    setSelected(
+      (items[newIndex]?.id === '4' ? items[newIndex + 1] : items[newIndex]) ??
```
robertknight, line 80: This test is because item 4 is disabled, right? That wasn't immediately obvious. Do we have a use case for disabling specific students in the list currently? If not I'd simplify the example by taking this out.

robertknight (commented):
This is generally looking in good shape. There are a few more things to address around keyboard input and accessibility. If we're agreed on following the "select-only combobox" pattern for this, then I think setting `role="combobox"` is enough to get this to a state where we can merge this version, and then a follow-up issue can be opened for the remaining items.

src/components/input/SelectNext.tsx
acelaya: The button was identified as a toggle becase I added `aria-pressed` attribute by mistake. I'll remove it.

I think I'm not going to go in the combobox direction for now, because then a11y rules require it to have a label or aria-label, whereas its own content is enough for screen readers while it has role button.

I considered inferring the `aria-label` attribute from the toggle content, but that would require a plain-text version of it (it allows ComponentChildren at the moment).

I'm not completely sure what's the right approach, but I see most UI libraries not using the combobox approach.

* https://mui.com/material-ui/react-select/
* https://headlessui.com/react/listbox
* https://getbootstrap.com/docs/5.3/components/dropdowns/#single-button

src/components/input/SelectNext.tsx
acelaya: > Do we need an explicit `aria-orientation` here

Yes. Default value is `horizontal` https://www.maxability.co.in/2015/07/29/aria-orientation-property

src/components/input/test/SelectNext-test.js
acelaya: Actually, I forgot I wanted to reset it to its previous value afterwards, otherwise it will keep the modified value for all tests coming after.

I'm not sure if a test seam would work here, because the bounding client rects from the other elements are affected by the window size and where are they rendered inside of it, but I will try just in case.

robertknight (approved):
Listbox popup button is now announced as expected, including collapsed / expanded state.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: Sounds reasonable. I can probably call the global one `defaultItems` or something like that.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: Yep. I initially used `Select_` so that we could show code in the examples, but I ended up manually writing the example because the part about providing props was more important.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: Yep, it's a bit hacky as it takes for granted things happening out of this. I'll simplify it.

src/components/input/SelectNext.tsx
acelaya: We can use `useFocusAway` for that. I'll address it on a follow-up PR

src/components/input/SelectNext.tsx
acelaya: I'll address this on a follow-up PR together with https://github.com/hypothesis/frontend-shared/pull/1247#discussion_r1334198485

src/components/input/test/SelectNext-test.js
acelaya: I realized adding padding to the container is enough. There's no need to change the window height during the test.

PR #1248, opened 2023-09-22
Allow type=number and password on Input component

acelaya:


src/components/input/Input.tsx
```diff
@@ -5,7 +5,7 @@ import { downcastRef } from '../../util/typing';
 import InputRoot from './InputRoot';
 
 type ComponentProps = {
-  type?: 'email' | 'search' | 'text' | 'url';
+  type?: 'email' | 'search' | 'text' | 'url' | 'number';
```
robertknight, line 8: I suggest to sort these values to make the list easier to scan.

robertknight (approved):

src/components/input/Input.tsx
robertknight: Also, are there other values for `type` that we should be enabling? See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types for a list.

acelaya: > I suggest to sort these values to make the list easier to scan.

Sort how?

acelaya: > Also, are there other values for type that we should be enabling? See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types for a list.

Perhaps it's worth having `password` as well.

acelaya: > > I suggest to sort these values to make the list easier to scan.
> 
> Sort how?

I have rearranged them a bit, putting text first, as it's the most common and default value.

PR #1256, opened 2023-09-25
Make Dialog API more strict depending on variation

acelaya:
This PR has been motivated by current requirement to pass some title (like `title=""`) when using `variant="custom"` in `Dialog`, even though that title is then ignored.

### Context

When the `Dialog` component was created, it was designed to wrap a `Panel`, and therefore, it extended some props from it that were later forwarded to the wrapped `Panel`.

Later, there was a use case to render a simpler `Dialog` where having to deal with a panel was inconvenient, and we introduced the `variant` prop, which could have values `'panel' | 'custom'`, being `panel` its default, and making provided children be rendered verbatim (not wrapped in a `Panel`) when value `custom` was provided.

However, the rest of the props API was the same, making some of the required `Panel` props still be required for `Dialog`, even when there's no `Panel` involved if `variant="custom"`.

### Actions taken

This PR updates the `Dialog` props so that its defined as a union type, where the `variant` prop is used to determine if the `Panel` props are required or not.

This way, we can continue defining panel dialogs, if no `variant` is provided, or it is provided with value `panel`:

```tsx
<Dialog title="Some dialog" buttons={...} icon={...} />

<Dialog variant="panel" title="Some dialog" buttons={...} icon={...} />
```

On the other hand, we can define custom dialogs where the panel props won't be required (in fact, they won't be allowed)

```tsx
<Dialog variant="custom" closeOnEscape restoreFocus />
```

### Considerations

The solution proposed here is the more resilient for consumers, as you will always be forced to provide a title as long as variation is `panel`, while you will be reminded no title can be used for `custom` dialogs.

However, it makes the `Dialog` implementation a bit complex, due to the mix of Panel props, Dialog props and HTML attributes, which requires a couple of type guard checks and multiple prop unpacking.

A more pragmatic solution would be to just define `title` as optional, and set a default fallback value when passing it down to the Panel. 

However, this solution is more prone to mistakes on consumers.

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -26,6 +26,7 @@ type ComponentProps = {
   closeOnClickAway?: boolean;
   closeOnEscape?: boolean;
   closeOnFocusAway?: boolean;
+
   /**
    * Dialog _should_ be provided with a close handler. We have a few edge use
    * cases, however, in which we need to render a "non-closeable" modal dialog.
@@ -48,19 +49,30 @@ type ComponentProps = {
    * Optional TransitionComponent for open (mount) and close (unmount) transitions
    */
   transitionComponent?: TransitionComponent;
-
-  /**
-   * Dialog lays out its content in a Panel by default, but `custom` allows any
-   * layout of Dialog children
-   */
-  variant?: 'panel' | 'custom';
 };
 
 // This component forwards a number of props on to `Panel` but always sets the
 // `fullWidthHeader` prop to `true`.
-export type DialogProps = PresentationalProps &
+export type PanelDialogProps = PresentationalProps &
   ComponentProps &
-  Omit<PanelProps, 'fullWidthHeader'>;
+  Omit<PanelProps, 'fullWidthHeader'> & {
+    /** Dialog lays out its content in a Panel by default */
+    variant?: 'panel';
+  };
+
+type CustomDialogProps = PresentationalProps &
+  ComponentProps & {
+    /** `custom` allows any layout of Dialog children */
+    variant: 'custom';
+  };
+
+export type DialogProps = PanelDialogProps | CustomDialogProps;
+
+function isPanelProps(
+  props: DialogProps,
+): props is Omit<PanelProps, 'fullWidthHeader'> {
+  return props.variant !== 'custom';
+}
 
 /**
  * Show a dialog
@@ -73,21 +85,32 @@ export default function Dialog({
   initialFocus = 'auto',
   restoreFocus = false,
   transitionComponent: TransitionComponent,
-  variant = 'panel',
 
   classes,
   elementRef,
-
-  // Forwarded to Panel
-  buttons,
-  icon,
   onClose,
-  paddingSize = 'md',
-  scrollable = true,
-  title,
 
-  ...htmlAttributes
+  ...rest
 }: DialogProps) {
+  const isPanel = isPanelProps(rest);
+  const {
+    buttons,
+    icon,
+    paddingSize = 'md',
+    title = '',
+    scrollable = true,
+    ...htmlAttributes
+  } = isPanel
+    ? rest
+    : {
+        buttons: undefined,
+        icon: undefined,
+        paddingSize: undefined,
+        title: undefined,
+        scrollable: undefined,
+        ...rest,
+      };
+
   const modalRef = useSyncedRef(elementRef);
   const restoreFocusEl = useRef<HTMLElement | null>(
     document.activeElement as HTMLElement | null,
@@ -225,7 +248,7 @@ export default function Dialog({
           )}
           ref={downcastRef(modalRef)}
         >
-          {variant === 'panel' && (
+          {isPanel ? (
             <Panel
               buttons={buttons}
               fullWidthHeader={true}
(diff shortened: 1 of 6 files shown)


src/components/feedback/Dialog.tsx
```diff
@@ -48,46 +49,77 @@ type ComponentProps = {
...
+    variant?: 'panel';
+  };
+
+type NonPanelDialogProps = PresentationalProps &
+  ComponentProps & {
+    /** `custom` allows any layout of Dialog children */
+    variant: 'custom';
+
+    /** @deprecated Title is ignored in custom dialogs */
+    title?: string;
```
acelaya, line 69: I've added this only for backwards compatibility, otherwise any consumer which was using `variant="custom"` will now get an error, as the title was mandatory up until now. But I feel tempted to just remove it.

src/components/feedback/Dialog.tsx
```diff
@@ -48,46 +49,77 @@ type ComponentProps = {
...
+  } = isPanelProps(rest)
+    ? rest
+    : {
+        buttons: undefined,
+        icon: undefined,
+        paddingSize: undefined,
+        title: undefined,
+        scrollable: undefined,
+        ...rest,
+      };
```
acelaya, line 121: This makes sure the panel props are stripped out of the `htmlAttributes`, avoiding type check errors when passing them to the `div` element.

src/components/feedback/Dialog.tsx
acelaya: I just noticed `title` gets marked as deprecated even when using `variant="panel"'`, so this might not be the best idea in the end.

src/components/feedback/Dialog.tsx
```diff
@@ -48,19 +49,33 @@ type ComponentProps = {
...
+    /** @deprecated Title is ignored in custom dialogs */
+    title?: string;
+  };
+
+export type DialogProps = PanelDialogProps | NonPanelDialogProps;
+
+function isPanelProps(
+  props: DialogProps,
+): props is Omit<PanelProps, 'fullWidthHeader'> {
+  return props.variant !== 'custom' && 'title' in props;
```
robertknight, line 77: I think that checking whether this is a panel should be based exclusively on the `variant` prop. Otherwise if, for any reason, the title attribute is incorrectly missing at runtime (eg. due to something that slips by TypeScript), the code will confusingly go down the wrong path.

src/components/feedback/Dialog.tsx
robertknight: > But I feel tempted to just remove it.

I would just do this. If we were being strict about semver we'd just bump the major version.

src/components/feedback/Dialog.tsx
```diff
@@ -48,19 +49,33 @@ type ComponentProps = {
...
-export type DialogProps = PresentationalProps &
+export type PanelDialogProps = PresentationalProps &
   ComponentProps &
-  Omit<PanelProps, 'fullWidthHeader'>;
+  Omit<PanelProps, 'fullWidthHeader'> & {
+    /** Dialog lays out its content in a Panel by default */
+    variant?: 'panel';
+  };
+
+type NonPanelDialogProps = PresentationalProps &
```
robertknight, line 63: Maybe name this `CustomDialogProps` to match the `variant` value?

src/components/feedback/Dialog.tsx
```diff
@@ -73,21 +88,32 @@ export default function Dialog({
...
 }: DialogProps) {
+  const isPanel = isPanelProps(rest);
+  const {
+    buttons,
+    icon,
+    paddingSize = 'md',
+    title = '',
+    scrollable = true,
+    ...htmlAttributes
+  } = isPanel
```
robertknight, line 106: An alternative approach to forwarding all the props would be to put all the panel props under a specific key. This would be a little less ergonomic to use in the common case that the dialog is a panel though.

src/components/feedback/Dialog.tsx
```diff
@@ -236,8 +262,9 @@ export default function Dialog({
             >
               {children}
             </Panel>
+          ) : (
+            <>{children}</>
```
robertknight, line 266: An alternative approach would be to make consumers responsible for composing the dialog and panel. It looks to me like `Dialog` is mostly just forwarding props here, plus setting a few additions and defaults (`fullWidthHeader`). This would be a more disruptive change to downstream consumers though.

robertknight (approved):
These changes effectively document the actual the current `Dialog` API better, so that's a plus.

Some alternative approaches that we could use to simplify the `Dialog` API in future might be:

1. Make the consumer responsible for composing `Dialog` and `Panel`. `Dialog` does pass some defaults along to the `Panel`, that could be done internally by using `cloneElement` though.
2. Since the dialog-with-panel is an extension of the custom dialog, you could create a separate component for it (`<PanelDialog ...>` instead of `<Dialog variant="custom">`)

These are all more disruptive changes to downstream consumers though, so this PR is a useful intermediate step.

acelaya:
> 2\. Since the dialog-with-panel is an extension of the custom dialog, you could create a separate component for it (`<PanelDialog ...>` instead of `<Dialog variant="custom">`)

Yep, this is probably how the Dialog should have been designed in the first place, but it was not easy to tell at first. It has become more obvious now that we have something in front of us.

I even remember discussing this with Lyza.

src/components/feedback/Dialog.tsx
acelaya: Makes sense

src/components/feedback/Dialog.tsx
acelaya: I'll remove it then. At the end of the day, we are most probably the only ones using this library.

PR #1267, opened 2023-09-26
Use grey-6 color for SelectNext chevron

### LINKED ISSUE #1250: SelectNext: chevron color is too dark

The color of the chevron is too dark in most cases (and also, it inherits its color). Set a fixed and lighter color that matches the one used for the original `Select` component.

acelaya:
Closes #1250 

While experimenting with `SelectNext` in client, I realized the chevron didn't have a fixed color, but instead it was inheriting default page color.

That was less obvious in the pattern library, as it uses a lighter color, but in client it was rendering on a too dark color.

This PR sets a fixed `text-grey-6` on the chevron for a more predictable result.

Before:

image

After:

image

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -162,7 +162,9 @@ function SelectMain<T>({
       >
         {label}
         <div className="grow" />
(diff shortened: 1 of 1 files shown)


robertknight (approved):
LGTM, although you can set class names on icons directly. See various examples in the client.

PR #1268, opened 2023-09-26
Add arrow key navigation `loop` option and use it in `SelectNext`

### LINKED ISSUE #1251: SelectNext: Prevent looping back to top/bottom in arrow keys navigation when reaching the end of the list

This component uses `useArrowKeyNavigation`, which has looping turned on by default.

Let's add an option so that it can be turned off by consumers.

acelaya:
Closes #1251 

Add a new `loop` option to `useArrowKeyNavigation` hook, which allows to check if focus should loop back to the first element once the end of the list is reached. It defaults to `true`, as that was its previous behavior.

Additionally, this PR disables `loop` in `SelectNext`, so that it behaves closer to native selects.

### Testing steps

1. Check out this branch and start dev server (`make dev`)
2. Go to http://localhost:4001/input-select-next
3. Arrow key navigation on the select dropdown should stop once reaching the first or last items.

On `main` branch, the navigation would loop back to the first item once the end is reached.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -110,7 +110,7 @@ function SelectMain<T>({
   useKeyPress(['Escape'], closeListbox);
 
   // Vertical arrow key for options in the listbox
-  useArrowKeyNavigation(wrapperRef, { horizontal: false });
+  useArrowKeyNavigation(wrapperRef, { horizontal: false, loop: false });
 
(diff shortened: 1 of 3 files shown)


robertknight (approved):
LGTM. The default makes sense for migration, although we might want to revisit it depending on whether looping is more commonly on/off in recommended keyboard handling for common widgets.

PR #1269, opened 2023-09-26
Exclude irrelevant mutations from useArrowKeyNavigation's MutationObserver

### LINKED ISSUE #1252: SelectNext: Exclude the toggle button from the arrow key navigation sequence

This component uses `useArrowKeyNavigation` to allow focusing the listbox items in sequence.

However, this hook works better with components which are always mounted, otherwise, it seems to skip the first item of the list.

This was worked around by including the select toggle button in the sequence, but we should change that.

acelaya:
Closes #1252

This PR excludes `SelectNext`'s main toggle button from arrow key navigation, so that once the listbox is open, you can navigate through its options exclusively.

This has been achieved by providing `selector: '[role="option"]` to `useArrowKeyNavigation`.

However, this was already attempted, but resulted in the second option in the listbox to be the first focused item.

The reason was that the chevron toggling (from chevron pointing down to chevron pointing up) was triggering a mutation from `useArrowKeyNavigation`'s internal MutationObserver, which resulted in the side effect being triggered twice when opened.

This PR also updates that so that the MutationObserver triggers the side effect only if any of the mutations happened on items matching `useArrowKeyNavigation`'s `selector`, or containing children that match `selector`.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -112,7 +112,11 @@ function SelectMain<T>({
   useKeyPress(['Escape'], closeListbox);
 
   // Vertical arrow key for options in the listbox
-  useArrowKeyNavigation(wrapperRef, { horizontal: false, loop: false });
+  useArrowKeyNavigation(wrapperRef, {
+    horizontal: false,
+    loop: false,
+    selector: '[role="option"]',
+  });
 
   useLayoutEffect(() => {
     if (!listboxOpen) {
--- a/src/hooks/test/use-arrow-key-navigation-test.js
+++ b/src/hooks/test/use-arrow-key-navigation-test.js
@@ -12,6 +12,9 @@ function Toolbar({ navigationOptions = {} }) {
 
   return (
     <div ref={containerRef} data-testid="toolbar" tabIndex={-1}>
+      <div>
+        <div data-testid="non-focusable-item" />
+      </div>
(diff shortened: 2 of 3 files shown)


src/components/input/SelectNext.tsx
```diff
@@ -110,7 +110,11 @@ function SelectMain<T>({
   useKeyPress(['Escape'], closeListbox);
 
   // Vertical arrow key for options in the listbox
-  useArrowKeyNavigation(wrapperRef, { horizontal: false, loop: false });
+  useArrowKeyNavigation(wrapperRef, {
+    horizontal: false,
+    loop: false,
+    selector: '[role="option"]',
+  });
```
acelaya, line 117: My initial intention was to change the containerRef, and use one for the listbox. This way the toggle button gets automatically excluded without having to change the selector.

However, this causes no items in the listbox to ever get focused, until you click on it. Then it becomes tab-focusable and then you can use arrow-key navigation.

I wasn't able to figure out why, so this is like the second best option.

We can leave this in the oven for a day and see if I get some inspiration to figure the other issue out and not having to mess-up with the hook internals.

acelaya:
Converting to draft while I try to get a better understanding of `useArrowKeyNavigation`, in order to find a better solution.

acelaya:
Replaced by https://github.com/hypothesis/frontend-shared/pull/1277

PR #1272, opened 2023-09-27
Add reusable ToastMessages component

### LINKED ISSUE #1075: Add reusable toast messages component

acelaya:
Closes #1075 

This PR brings the `ToastMessages` component from client/via, and adds the corresponding documentation to the pattern library.

image

Once this is merged, we can start using it in frontend apps and remove duplicated code.

### Testing steps

1. Check out this branch.
2. Start dev server (`make dev`)
3. Go to http://localhost:4001/feedback-toast-messages

### DIFF
--- a/src/components/feedback/ToastMessages.tsx
+++ b/src/components/feedback/ToastMessages.tsx
@@ -1,0 +1,221 @@
+import classnames from 'classnames';
+import type {
+  ComponentChildren,
+  ComponentProps,
+  FunctionComponent,
+} from 'preact';
+import {
+  useCallback,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'preact/hooks';
+
+import type { TransitionComponent } from '../../types';
+import Callout from './Callout';
+
+export type ToastMessage = {
+  id: string;
+  type: 'error' | 'success' | 'notice';
+  message: ComponentChildren;
+
+  /**
+   * Visually hidden messages are announced to screen readers but not visible.
+   * Defaults to false.
+   */
+  visuallyHidden?: boolean;
+
+  /**
+   * Determines if the toast message should be auto-dismissed.
+   * Defaults to true.
+   */
+  autoDismiss?: boolean;
+};
+
+export type ToastMessageTransitionClasses = {
+  /** Classes to apply to a toast message when appended. Defaults to 'animate-fade-in' */
+  transitionIn?: string;
+  /** Classes to apply to a toast message being dismissed. Defaults to 'animate-fade-out' */
+  transitionOut?: string;
+};
+
+type ToastMessageItemProps = {
+  message: ToastMessage;
+  onDismiss: (id: string) => void;
+};
+
+/**
+ * An individual toast message: a brief and transient success or error message.
+ * The message may be dismissed by clicking on it. `visuallyHidden` toast
+ * messages will not be visible but are still available to screen readers.
+ */
+function ToastMessageItem({ message, onDismiss }: ToastMessageItemProps) {
+  // Capitalize the message type for prepending; Don't prepend a message
+  // type for "notice" messages
+  const prefix =
+    message.type !== 'notice'
+      ? `${message.type.charAt(0).toUpperCase() + message.type.slice(1)}: `
+      : '';
+
+  return (
+    <Callout
+      classes={classnames({
+        'sr-only': message.visuallyHidden,
+      })}
+      status={message.type}
+      onClick={() => onDismiss(message.id)}
+      variant="raised"
+    >
+      <strong>{prefix}</strong>
+      {message.message}
+    </Callout>
+  );
+}
+
+type ToastMessageTransitionType = FunctionComponent<
+  ComponentProps<TransitionComponent> & {
+    transitionClasses?: ToastMessageTransitionClasses;
+  }
+>;
+
+const ToastMessageTransition: ToastMessageTransitionType = ({
+  direction,
+  onTransitionEnd,
+  children,
+  transitionClasses = {},
+}) => {
+  const isDismissed = direction === 'out';
+  const containerRef = useRef<HTMLDivElement>(null);
+  const handleAnimation = (e: AnimationEvent) => {
+    // Ignore animations happening on child elements
+    if (e.target !== containerRef.current) {
+      return;
+    }
+
+    onTransitionEnd?.(direction ?? 'in');
+  };
+  const classes = useMemo(() => {
+    const {
+      transitionIn = 'animate-fade-in',
+      transitionOut = 'animate-fade-out',
+    } = transitionClasses;
+
+    return {
+      [transitionIn]: !isDismissed,
+      [transitionOut]: isDismissed,
+    };
+  }, [isDismissed, transitionClasses]);
+
+  return (
+    <div
+      data-testid="animation-container"
+      onAnimationEnd={handleAnimation}
+      ref={containerRef}
+      className={classnames('relative w-full container', classes)}
+    >
+      {children}
+    </div>
+  );
+};
+
+export type ToastMessagesProps = {
+  messages: ToastMessage[];
+  onMessageDismiss: (id: string) => void;
+  transitionClasses?: ToastMessageTransitionClasses;
+  setTimeout_?: typeof setTimeout;
+};
+
+type TimeoutId = number;
+
+/**
+ * A collection of toast messages. These are rendered within an `aria-live`
+ * region for accessibility with screen readers.
+ */
+export default function ToastMessages({
+  messages,
+  onMessageDismiss,
+  transitionClasses,
+  /* istanbul ignore next - test seam */
+  setTimeout_ = setTimeout,
+}: ToastMessagesProps) {
+  // List of IDs of toast messages that have been dismissed and have an
+  // in-progress 'out' transition
+  const [dismissedMessages, setDismissedMessages] = useState<string[]>([]);
+  // Tracks not finished timeouts for auto-dismiss toast messages
+  const messageSchedules = useRef(new Map<string, TimeoutId>());
+
+  const dismissMessage = useCallback(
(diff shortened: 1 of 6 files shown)


src/components/feedback/ToastMessages.tsx
acelaya, line 1: Brought verbatim from client https://github.com/hypothesis/client/blob/2fa444e05d806893606a9181423d9d4f35bfbc7f/src/shared/components/ToastMessages.tsx

src/components/feedback/test/ToastMessages-test.js
acelaya, line 1: Brought verbatim from client https://github.com/hypothesis/client/blob/2fa444e05d806893606a9181423d9d4f35bfbc7f/src/shared/components/test/ToastMessages-test.js

I only added a small fix to make sure there are no components left in the DOM that have side effects in other tests.

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,206 @@
...
+        return;
+      }
+
+      // Track that this message has been scheduled to be dismissed. After a
+      // period of time, actually dismiss it
+      scheduledMessages.current.add(id);
+      setTimeout_(() => {
+        dismissMessage(id);
+        scheduledMessages.current.delete(id);
+      }, 5000);
```
robertknight, line 153: Minor issue - this timeout doesn't get cleared if the component is unmounted.

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,206 @@
...
+    },
+    [dismissMessage, setTimeout_],
+  );
+
+  const onTransitionEnd = useCallback(
+    (direction: 'in' | 'out', message: ToastMessage) => {
+      const autoDismiss = message.autoDismiss ?? true;
```
robertknight, line 160: We might want to consider making the default value of `autoDismiss` depend upon the message type in future, eg. by disabling it for errors.

src/components/feedback/test/ToastMessages-test.js
```diff
@@ -0,0 +1,168 @@
...
+      callback => callback(),
+    );
+
+    // Trigger "in" animation for all messages, which will schedule dismiss for
+    // appropriate messages
+    messages.forEach((_, index) => {
+      triggerAnimationEnd(wrapper, index, 'in');
+    });
+
+    // Trigger "out" animation on components which "direction" prop is currently
```
robertknight, line 102:
```suggestion
    // Trigger "out" animation on components whose "direction" prop is currently
```

src/components/feedback/test/ToastMessages-test.js
```diff
@@ -0,0 +1,168 @@
...
+        if (transitionComponent.prop('direction') === 'out') {
+          triggerAnimationEnd(wrapper, index);
+        }
+      });
+
+    // Only one toast message will remain, as it was marked as `autoDismiss: false`
+    assert.equal(fakeOnMessageDismiss.callCount, 3);
+  });
+
+  it('schedules dismiss only once per message', async () => {
```
robertknight, line 116: This test is marked async, but contains no `await`. Is it missing something?

robertknight (approved):
LGTM with some minor comments. I would leave the ones about changing the arguments to the `onDismiss` callback for a subsequent revision.

src/pattern-library/components/patterns/feedback/ToastMessagesPage.tsx
```diff
@@ -0,0 +1,229 @@
...
+        )}
+      </div>
+    </div>
+  );
+}
+
+export default function ToastMessagesPage() {
+  return (
+    <Library.Page
+      title="Callout"
```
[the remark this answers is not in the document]
acelaya: OMG, and it's huge in the page. I don't know how I missed that 😅

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,206 @@
...
+      onClick={() => onDismiss(message.id)}
+      variant="raised"
+    >
+      <strong>{prefix}</strong>
+      {message.message}
+    </Callout>
+  );
+}
+
+type BaseToastMessageTransitionType = FunctionComponent<
```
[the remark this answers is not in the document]
acelaya: I think we used `Base` initially with the intention that at some point consumers would be able to customize this. However, `Default` would have probably been a better prefix for that purpose.

Since we don't have this capability for now, and we have moved in a direction in which we allow customizing classes, I will rename this just `ToastMessageTransition`.

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,206 @@
...
+      className={classnames('relative w-full container', classes)}
+    >
+      {children}
+    </div>
+  );
+};
+
+export type ToastMessagesProps = {
+  messages: ToastMessage[];
+  onMessageDismiss: (id: string) => void;
```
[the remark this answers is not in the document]
acelaya: I'll create a follow-up issue to change this for next major version

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,206 @@
...
+ * region for accessibility with screen readers.
+ */
+export default function ToastMessages({
+  messages,
+  onMessageDismiss,
+  transitionClasses,
+  /* istanbul ignore next - test seam */
+  setTimeout_ = setTimeout,
+}: ToastMessagesProps) {
+  const [dismissedMessages, setDismissedMessages] = useState<string[]>([]);
```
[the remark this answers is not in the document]
acelaya: `dismissedMessages` are those that have been dismissed, but which are still performing an exit transition, but it includes both user-dismissed ones, and auto-dismissed ones.

`scheduledMessages` only tracks those that had `autoDismiss` true when the `in` transition ended (which is effectively when they were appended). The initial intention was to be able to cancel their timeout if they were manually dismissed before scheduled, but in the end that was never done. We just let that trigger eventually and do nothing.

I'm starting to think if `scheduledMessages` is actually needed. We also check that the same message is not scheduled to dismiss more than once, but I don't know if that's even possible or if that would also have any effect that really justifies the extra checks.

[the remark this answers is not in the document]
acelaya: Yeah, this is not needed.

All tests pass even without that, and some manual testing and console.logs prove that is never called more than once for the same message, so it's an unnecessary extra check.

I'll remove `scheduledMessages`.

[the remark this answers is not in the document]
acelaya: I'll also add a comment to `dismissedMessages` explaining what it is.

src/components/feedback/test/ToastMessages-test.js
acelaya: Nop, I think it's just an overlook after refactoring the test. It works the same without `async`, so I've removed it.

src/pattern-library/routes.ts
```diff
@@ -142,6 +143,12 @@ const routes: PlaygroundRoute[] = [
     component: CalloutPage,
     route: '/feedback-callout',
   },
+  {
+    title: 'ToastMessages',
+    group: 'feedback',
+    component: ToastMessagesPage,
+    route: '/feedback-toast-messages',
+  },
```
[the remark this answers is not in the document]
acelaya: Oh, I didn't notice this was ordered alphabetically and I thought it made sense to keep Callout and ToastMessages together.

Will fix it.

src/components/feedback/ToastMessages.tsx
acelaya: I have added a Map ref to track in-progress timeots, and a `useEffect` that returns the callback to clear them.

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+      if (direction === 'out') {
+        onMessageDismiss(message.id);
+        setDismissedMessages(ids => ids.filter(id => id !== message.id));
+      }
+    },
+    [scheduleMessageDismiss, onMessageDismiss],
+  );
+
+  useEffect(() => {
```
robertknight, line 176: Probably want to use a layout effect for this so that it happens immediately on unmount, rather than on the next animation frame.

src/pattern-library/components/patterns/feedback/ToastMessagesPage.tsx
```diff
@@ -0,0 +1,249 @@
...
+                onMessageDismiss={() => {}}
+              />
+            </Library.Demo>
+          </Library.Example>
+          <Library.Example title="onMessageDismiss">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Callback invoked with the toast message id, when it ends its
+                dismiss transition. The message with that id can then be removed
+                form the list.
```
robertknight, line 206:
```suggestion
                from the list.
```

robertknight (commented):
I had a few minor corrections and comments, otherwise looks good.

src/components/feedback/ToastMessages.tsx
```diff
@@ -0,0 +1,219 @@
...
+  messages,
+  onMessageDismiss,
+  transitionClasses,
+  /* istanbul ignore next - test seam */
+  setTimeout_ = setTimeout,
+}: ToastMessagesProps) {
+  // List of IDs of toast messages that have been dismissed and have an
+  // in-progress 'out' transition
+  const [dismissedMessages, setDismissedMessages] = useState<string[]>([]);
+  // Tracks not finished timeouts for auto-dismiss toast messages
```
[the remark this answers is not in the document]
acelaya: I like the type alias idea.

src/components/feedback/ToastMessages.tsx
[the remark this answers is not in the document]
acelaya: https://github.com/hypothesis/frontend-shared/issues/1273

PR #1274, opened 2023-09-27
Extract hook with logic to determine SelectNext listbox location

acelaya:
This creates a new `useShouldBePositionedAbove` hook that wraps the logic originally defined in `SelectNext` to determine if the listbox should be rendered above or below the toggle button.

There could be other use-cases for this hook, but for now I have opted to not expose it until we really see a case for it.

The main motivation to extract this is to simplify the code in `SelectNext` while trying to tackle https://github.com/hypothesis/frontend-shared/pull/1269 and is proving a bit confusing otherwise.

Due to the reasons explained above, I have also decided not to extract individual tests for this new hook for now. Its logic is covered by `SelectNext`'s hook anyway.

We can extract tests in a follow-up PR or if we decide to expose it and use it in other cases.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -13,6 +13,7 @@ import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
 import { useClickAway } from '../../hooks/use-click-away';
 import { useFocusAway } from '../../hooks/use-focus-away';
 import { useKeyPress } from '../../hooks/use-key-press';
+import { useShouldBePositionedAbove } from '../../hooks/use-should-be-positioned-above';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
 import type { PresentationalProps } from '../../types';
 import { MenuCollapseIcon, MenuExpandIcon } from '../icons';
@@ -90,13 +91,17 @@ function SelectMain<T>({
   elementRef,
 }: SelectProps<T>) {
   const [listboxOpen, setListboxOpen] = useState(false);
-  const [shouldListboxDropUp, setShouldListboxDropUp] = useState(false);
   const closeListbox = useCallback(() => setListboxOpen(false), []);
   const wrapperRef = useRef<HTMLDivElement | null>(null);
   const listboxRef = useRef<HTMLDivElement | null>(null);
   const listboxId = useId();
   const buttonRef = useSyncedRef(elementRef);
   const buttonId = useId();
+  const shouldListboxDropUp = useShouldBePositionedAbove(
+    buttonRef.current ?? null,
+    listboxRef.current,
+    listboxOpen,
+  );
 
   const selectValue = useCallback(
     (newValue: unknown) => {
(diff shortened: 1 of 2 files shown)


src/hooks/use-should-position-above.ts
```diff
@@ -0,0 +1,49 @@
...
+ * those two components, and the available viewport space above and below the
+ * reference component.
+ *
+ * By default, we prefer the popping component to be positioned below the
+ * reference component, and only if there's enough space above, and not enough
+ * space below, we will prefer it to be positioned above.
+ */
+export function useShouldPositionAbove(
+  referenceComponent: HTMLElement | null,
+  poppingComponent: HTMLElement | null,
```
acelaya, line 15: I don't love the `referenceComponent` and `poppingComponent` names, so I'm open to suggestions here.

src/hooks/use-should-be-positioned-above.ts
```diff
@@ -0,0 +1,49 @@
...
+ * those two components, and the available viewport space above and below the
+ * reference component.
+ *
+ * By default, we prefer the popping component to be positioned below the
+ * reference component, and only if there's enough space above, and not enough
+ * space below, we will prefer it to be positioned above.
+ */
+export function useShouldBePositionedAbove(
+  referenceComponent: HTMLElement | null,
+  poppingComponent: HTMLElement | null,
```
acelaya, line 15: I don't love the `referenceComponent` and `poppingComponent` names, so I'm open to suggestions here.

src/components/input/SelectNext.tsx
```diff
@@ -90,13 +91,17 @@ function SelectMain<T>({
...
 }: SelectProps<T>) {
   const [listboxOpen, setListboxOpen] = useState(false);
-  const [shouldListboxDropUp, setShouldListboxDropUp] = useState(false);
   const closeListbox = useCallback(() => setListboxOpen(false), []);
   const wrapperRef = useRef<HTMLDivElement | null>(null);
   const listboxRef = useRef<HTMLDivElement | null>(null);
   const listboxId = useId();
   const buttonRef = useSyncedRef(elementRef);
   const buttonId = useId();
+  const shouldListboxDropUp = useShouldBePositionedAbove(
```
robertknight, line 100: If this is an internal helper not used by other components, it could just go inside the same `SelectNext` module. Then it doesn't have to pretend to be generic.

src/hooks/use-should-be-positioned-above.ts
robertknight: `reference`/`popup` would be fine. An `Element` suffix could be added, although that information is already available in the types. Also these are not _components_ in the Preact sense, but DOM elements.

src/components/input/SelectNext.tsx
acelaya: That's probably the best approach at this point. We can keep the naming close to select/listbox and think on better names and extracting it if the need apperas.

acelaya:
Replaced by https://github.com/hypothesis/frontend-shared/pull/1276

PR #1276, opened 2023-09-28
Extract local useShouldDropUp hook from SelectNext

acelaya:
Extract the logic to determine if the listbox in `SelectNext` should drop down or up into a local hook, to improve readability.

> This PR replaces https://github.com/hypothesis/frontend-shared/pull/1274


src/components/input/SelectNext.tsx
```diff
@@ -90,13 +123,17 @@ function SelectMain<T>({
...
   const wrapperRef = useRef<HTMLDivElement | null>(null);
   const listboxRef = useRef<HTMLDivElement | null>(null);
   const listboxId = useId();
   const buttonRef = useSyncedRef(elementRef);
   const buttonId = useId();
+  const shouldListboxDropUp = useShouldDropUp(
+    buttonRef.current ?? null,
+    listboxRef.current,
+    listboxOpen,
+  );
```
robertknight, line 136: You need to pass in the refs here, rather than the current value. If, for any reason, `listboxOpen` was true on the first render, the layout effect wouldn't get called.

robertknight (commented):
To ensure the helper is working with the current values of the refs, you need to pass it the refs rather than the value at the point when the render happens.

src/components/input/SelectNext.tsx
acelaya: Ugh, you are right

robertknight (approved):

PR #1277, opened 2023-09-28
Allow useArrowKeyNavigation consumers tell if container is visible

### LINKED ISSUE #1252: SelectNext: Exclude the toggle button from the arrow key navigation sequence

acelaya:
Closes #1252
Depends on https://github.com/hypothesis/frontend-shared/pull/1276

This PR adds a new `containerVisible` flag option to `useArrowKeyNavigation` that can be used by consumers to more granularly control when to re-compute the hook's side effects.

This is needed if the arrow-key navigation is used on components which are initially not visible.

The PR also updates the logic in `SelectNext` to use this new option, allowing it to use the listbox as the container of `useArrowKeyNavigation`, and therefore preventing the toggle button to be included in the focus sequence.

### DIFF
--- a/src/hooks/use-arrow-key-navigation.ts
+++ b/src/hooks/use-arrow-key-navigation.ts
@@ -33,6 +33,15 @@ export type UseArrowKeyNavigationOptions = {
    * CSS selector which specifies the elements that navigation moves between
    */
   selector?: string;
+
+  /**
+   * Indicates if the container element is currently visible.
+   * This information is used to focus the current element when the container
+   * transitions from hidden to visible.
+   *
+   * Defaults to `true`.
+   */
+  containerVisible?: boolean;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/hooks/use-arrow-key-navigation.ts
```diff
@@ -33,6 +33,14 @@ export type UseArrowKeyNavigationOptions = {
...
    */
   selector?: string;
+
+  /**
+   * Tells if the container is visible, ensuring no elements on it are focused
+   * while it is not.
+   * It also lets consumers control when the side effects should be re-computed.
+   * Defaults to `true`.
+   */
+  containerVisible?: boolean;
```
robertknight, line 43: This description is a bit vague. I would suggest something like: "Indicates if the container element is currently visible. This information is used to focus the current element when the container transitions from hidden to visible."

robertknight (approved):
LGTM. Works as expected.

src/hooks/use-arrow-key-navigation.ts
acelaya: Writing a lot of words and not say anything is my specialty 😅.

Thanks for the suggestion. I quite like it.

Issue #1285, opened 2023-09-29
Fix issues found after using `SelectNext` in the open


acelaya:
Some bugs have appeared after using the `SelectNext` in a real world use case.

* [x] Toggle's `onKeyDown` should call `e.preventDefault()` when all conditions it checks are met (addressed by https://github.com/hypothesis/frontend-shared/pull/1288).
* [x] Add a `buttonId` prop that can be used to propagate as `id` to the toggle button (addressed by https://github.com/hypothesis/frontend-shared/pull/1288).
* [x] Arrow key navigation resets focus to the first element of the list every time it's open (probably related with this https://github.com/hypothesis/frontend-shared/pull/1277). It should keep the focused element when it closed (addressed by https://github.com/hypothesis/frontend-shared/pull/1290).
* [x] `SelectNext` and `SelectNext.Option` have non-intuitive `displayName`s. Explicitly set them (addressed by https://github.com/hypothesis/frontend-shared/pull/1290):
    * `SelectMain.displayName = 'SelectNext'`
    * `SelectOption.displayName = 'SelectNext.Option'`
* [x] The `label` prop represents the content of the toggle button, but its name suggests something else. Replace with a new `buttonContent` prop and deprecate `label` (addressed by https://github.com/hypothesis/frontend-shared/pull/1288).
* [ ] Clicking the toggle displays the listbox, sets `tabIndex=0` to the first option (via `useArrowKeyNavigation`) but does not focus first item. This means that pressing `ArrowDown` effectively selects the second option instead of the first one (addressed by https://github.com/hypothesis/frontend-shared/pull/1292).
* [ ] If keyboard and mouse are combined to subsequently select different options, latest focused item is not properly update (addressed by https://github.com/hypothesis/frontend-shared/pull/1292).

To be discussed:

* `SelectNext` should implement `HTMLButtonAttributes`, and pass them down to the toggle button, excluding those attributes internally handled, like `aria-haspopup` and `aria-controls`.

PR #1288, opened 2023-10-02
Improve SelectNext props API

acelaya:
Part of https://github.com/hypothesis/frontend-shared/issues/1285

This PR addresses some of the issues highlighted while using it in a real use-case:

* Make sure we prevent default behavior of the keyboard event dispatched when pressing `ArrowDown` on the select toggle, which is used to open the dropdown, and otherwise scrolls the page down a bit.
* Allow a custom `buttonId` to be provided, which is passed as the toggle button `id` if provided. We keep generating a fallback value with `useId`.
* Deprecate `label` and replace it with `buttonContent`, as `label` was suggesting a different purpose than the one it is actually used for.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -110,18 +110,29 @@ function useShouldDropUp(
 export type SelectProps<T> = PresentationalProps & {
   value: T;
   onChange: (newValue: T) => void;
-  label: ComponentChildren;
+  buttonContent?: ComponentChildren;
   disabled?: boolean;
+
+  /**
+   * `id` attribute for the toggle button. This is useful to associate a label
+   * with the control.
+   */
+  buttonId?: string;
+
+  /** @deprecated Use buttonContent instead */
+  label?: ComponentChildren;
 };
 
 function SelectMain<T>({
+  buttonContent,
   label,
   value,
   onChange,
   children,
   disabled,
-  classes,
   elementRef,
+  classes,
+  buttonId,
 }: SelectProps<T>) {
   const [listboxOpen, setListboxOpen] = useState(false);
   const closeListbox = useCallback(() => setListboxOpen(false), []);
@@ -129,7 +140,7 @@ function SelectMain<T>({
   const listboxRef = useRef<HTMLDivElement | null>(null);
   const listboxId = useId();
   const buttonRef = useSyncedRef(elementRef);
-  const buttonId = useId();
+  const defaultButtonId = useId();
   const shouldListboxDropUp = useShouldDropUp(
     buttonRef,
     listboxRef,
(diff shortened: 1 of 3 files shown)


robertknight:
> Extend the props API, allowing all props from HTMLButtonElement to be propagated down to the wrapped Button. 

Hmm, I think it is possibly unclear here which element the props will be applied to, since the select control has at least three elements:

- An outer container
- A button
- A listbox

The general convention in the package is that forwarded props get applied to the outermost element, although I do see at least one exception (`Checkbox`). A possible solution would be to have specific props to target those elements (eg. `buttonProps={...}`).

We could also defer this decision by only supporting setting an `id` prop for the button for the moment.

acelaya:
> > Extend the props API, allowing all props from HTMLButtonElement to be propagated down to the wrapped Button.
> 
> Hmm, I think it is possibly unclear here which element the props will be applied to, since the select control has at least three major elements:
> 
> * An outer container
> * A button
> * A listbox
> 
> The general convention in the package is that forwarded props get applied to the outermost element, although I do see at least one exception (`Checkbox`). A possible solution would be to have specific props to target those elements (eg. `buttonProps={...}`).
> 
> We could also defer this decision by only supporting setting an `id` prop for the button for the moment.

If we check the `elementRef` description, it says `Ref associated with component's outermost or primary element`.

In most of the components, the outermost and primary elements are the same, but there are exceptions, like the `Checkbox` you mentioned.

I think `SelectNext` is another exception. The outermost element is a container div, but the most relevant and primary element is the button.

With this in mind I think it makes more sense (and it is what I would personally expect) that the main props API we expose allows non-specific props to be forwarded to the button.

However, I think it would make sense to have `listboxProps={{ ... }}` if needed at some point.

I'm also open to start just with the `id` for now, and not try to solve a problem we don't have yet.

acelaya:
@robertknight in order to move things forward I have reverted the `HTMLButtonAttributes` change and added a `buttonId` prop for now.

Let's continue the other discussion when/id needed.

src/components/input/SelectNext.tsx
```diff
@@ -110,26 +110,37 @@ function useShouldDropUp(
 export type SelectProps<T> = PresentationalProps & {
   value: T;
   onChange: (newValue: T) => void;
-  label: ComponentChildren;
+  buttonContent?: ComponentChildren;
   disabled?: boolean;
+
+  /**
+   * Propagated as `id` attributed for the toggle button.
+   * A random one is generated by default
```
robertknight, line 118:
```suggestion
   * `id` attribute for the toggle button. This is useful to associate a label with the control.
```

A randomly generated ID is not useful from the outside, since the caller has no way to refer to it. If that randomly generated ID is used internally, that's just an implementation detail.

robertknight (approved):

src/components/input/SelectNext.tsx
acelaya: Good point.

PR #1289, opened 2023-10-02
Adjusted various UI elements `border-radius`

jaredpdesigns:
This PR increases the overall "roundness" of many elements in the UI, which rely on the `rounded` class from Tailwind. Primary changes made:
- Removed the overwrite of `DEFAULT` value for `border-radius` inside of `tailwind.preset.js` which reset the `rounded` class value to equal 4px.
- Adjusted a few elements which had behavior different than using `rounded` including `InputGroup`
- Increased the radius to 8px for `Card` and `Tab` using the `rounded-lg` class.
- Added roundness to `Table` which may need to reverted based on feedback.

> Part of https://github.com/hypothesis/frontend-shared/issues/1304

### DIFF
--- a/src/components/data/ScrollBox.tsx
+++ b/src/components/data/ScrollBox.tsx
@@ -6,6 +6,8 @@ import ScrollContent from './ScrollContent';
 export type ScrollBoxProps = CompositeProps & {
   /** Turn off borders on outer container */
   borderless?: boolean;
+  /** Add rounded corners to scrollable container */
+  rounded?: boolean;
 };
 
 /**
@@ -17,6 +19,7 @@ export default function ScrollBox({
   elementRef,
 
   borderless = false,
+  rounded = false,
 
   ...htmlAttributes
 }: ScrollBoxProps) {
@@ -25,6 +28,7 @@ export default function ScrollBox({
       data-composite-component="ScrollBox"
       {...htmlAttributes}
       borderless={borderless}
+      rounded={rounded}
       elementRef={elementRef}
     >
       <Scroll>
--- a/src/pattern-library/components/patterns/data/TablePage.tsx
+++ b/src/pattern-library/components/patterns/data/TablePage.tsx
@@ -285,7 +285,7 @@ export default function TablePage() {
             <Library.Info>
               <Library.InfoItem label="description">
                 Set this boolean prop if you want to remove the table{"'"}s
-                outer borders, in case it is rendered on a container with its
+                outer borders, in case it is rendered in a container with its
                 own borders.
               </Library.InfoItem>
               <Library.InfoItem label="type">
--- a/src/pattern-library/components/patterns/data/TablePage.tsx
+++ b/src/pattern-library/components/patterns/data/TablePage.tsx
@@ -1,4 +1,5 @@
 import {
+  ScrollContainer,
   Scroll,
   Table,
   TableBody,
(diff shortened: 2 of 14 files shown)


robertknight:
A few screenshots of what this branch looks like in the client:

Annotation cards:

Rounded corners

With these changes we now have three distinct levels of rounded-ness in the UI for the top bar and menus (square), left sidebar buttons (small rounding) and annotation cards (large rounding). We probably are going to need different levels of roundedness for different elements, but we should have a coherent story for when to use each.

Tabs. Note there is an issue with the connection between the tab list and the panels:

Tabs

Dialog:

Annotation card

Menus:

These are currently still square.

Menu roundedness

robertknight:
LMS assignment configuration front page:

LMS assignment details

There is an issue here with the corners which we'll need to fix.

LMS assignment picker URL entry dialog:

LMS enter URL

LMS Canvas files picker:

LMS canvas files picker

LMS YouTube video picker:

LMS YouTube video picker

jaredpdesigns:
@robertknight just a few comments:
- Regarding tabs, the best way to adjust this is `<Card classes={selectedTab === 'shared' ? 'rounded-tl-none' : undefined}>` where the class to remove rounding from the adjacent `Card` is conditional, I had implemented that in one of the demo pages in the Pattern Library.
- In the case of some of the LMS examples, specifically the Assignment Configuration page, adding an additional class of `overflow-hidden` is a heavy-handed way to confirm the roundness of the parent element masks any children.
- Menus would be another good thing to review, is this an element in the Pattern Library? In general agree with the 3 levels of roundness. In my opinion most everything that isn't a card should defer to using the `rounded` class, which defaults to 4px, and only element that are either the Card itself or "touching" elements like the tabs should take on the `rounded-lg` class which doubles the roundness to 8px.

acelaya:
> Menus would be another good thing to review, is this an element in the Pattern Library?

No, menus are currently defined in `client`. See [Menu](https://github.com/hypothesis/client/blob/38f8caa2331cca213d083c04b1b21e580add4866/src/sidebar/components/Menu.tsx), [MenuSection](https://github.com/hypothesis/client/blob/38f8caa2331cca213d083c04b1b21e580add4866/src/sidebar/components/MenuSection.tsx) and [MenuItem](https://github.com/hypothesis/client/blob/38f8caa2331cca213d083c04b1b21e580add4866/src/sidebar/components/MenuItem.tsx)

jaredpdesigns:
@acelaya I'd be happy to spin up a separate PR to look at stuff in the client that the Pattern Library doesn't cover. Looks like we are using Tailwind here as well. However, if it’s loading Tailwind via the `frontend-shared` repo we may need to release the changed there first then look at the client.

acelaya:
> However, if it’s loading Tailwind via the frontend-shared repo we may need to release the changes there first then look at the client.

We usually use `yalc` to test unreleased frontend-shared changes in the other projects.

You can see the instructions here https://github.com/hypothesis/frontend-shared/blob/f3ac2d1a3c7d804e0b6c77c42e472286dbab1c4b/docs/developing.md#testing-locally-with-other-projects

jaredpdesigns:
@robertknight @acelaya is there more here that needs to be done before considering merging it? Not sure what is remaining to consider. Does it just need some other kind of approval?

cc: @dwhly

acelaya:
Hey @jaredpdesigns.

The main consideration before merging this, is that the way it is now, it's a breaking change. That means it would block adoption of this library, and make it harder to release other feature or fixes here.

That's why we suggested testing it in every consuming project with `yalc`, before merging it.

Another option (probably preferable, now that I think about it) would be to try and make it non-breaking, by keeping the `borderRadius` configuration as it is, and letting downstream projects adopt rounded corners and overwrite that config at their own pace.

That way we can use this PR as a sandbox just to make every component here ready for rounded corners (the other changes you did), but without introducing side effects if downstream needs to update to newer frontend-shared versions for other unrelated reasons.

On a later stage, once every project has adopted rounded corners, we can remove the `borderRadius` option from the tailwind preset here.

There are a few grey areas though, which are the components moving here from `rounded` to `rounded-lg` and such, but I think that, considering the rounded corner size we have now, it is negligible. Worth double checking though.

robertknight:
I agree with @acelaya that it would probably be best for us to roll out the changes one project at a time, and then fold them back into the pattern library when we are done. We could do as @acelaya suggested and roll out the change in each individual project's config first, and then later fold it back into the pattern library. Or we could change the defaults here in a new major version and override them downstream until we are happy that the revised styles are working as desired.

I think it would be quite useful if we could feature flag style changes like this. Alejandro and I can talk about it and see how much effort that would require. Perhaps we could do something with CSS variables for example.

acelaya:
> Is there anything else I need to do at this point?

I think you are good 🙂 

During this sprint we'll do a couple more checks and then merge this with the needed adjustments.

acelaya:
I have continued testing this, and found a couple of edge cases that needed to be fixed. Having in mind the our roll-out plan, these are the things I have changed :

> I have pushed the changes in additional commits.

- `CardHeader` with `variant="secondary"` was rendering a background color "outside" the card.
  - I have fixed it by adding `rounded-t-[inherit]` to the `CardHeader`.
- I had mixed feelings about adding rounded corners to tables, and I have found some problems with that.
  1. Sticky headers were no longer working due to the `overflow-hidden` added to `table`.
  2. Tables inside scrollable containers looked odd, due to rounded corners being only visible when at the very top or very bottom.
  
  [not-working.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/cbae2e3a-4403-4a38-8174-fd1c29149f2d)
  - My proposed solution is to not make tables have rounded corners by default, and instead wrap them in a container with rounded corners when needed.
  - This required adding a new `borderless` property to tables to make them look good when wrapped in a container which renders its own border.
  
  [working.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/4dce108a-2d7e-4b05-a46c-31ce6c5ac388)
- I noticed our `Scroll` did not have rounded corners.
  - I kept this as the default behavior, but added a new `rounded` prop to `ScrollBox` and `ScrollContainer` that allows for scrollable containers to render rounded corners that are always visible.
  This is useful for cases like the above (tables) but can be useful in other cases.
- As discussed in previous comments, the changes in tailwind preset are not backward compatible.
  - I have reverted back to what we had before, with two changes:
    1. We not only define `DEFAULT`, but also `lg`, because I have noticed it was too obvious and still a breaking change.
    2. I have defined the config in a way that downstream projects can overwrite it via CSS vars, allowing for this to be dynamically enabled at runtime (for example with feature flags).
    ```
    borderRadius: {
      lg: 'var(--h-border-radius-lg, 0.125rem)',
      DEFAULT: 'var(--h-border-radius, 0.125rem)',
    }
    ```
    3. Next major version can drop this completely, but all downstream projects should update first.

###  TBD

There's a few things that still need testing before merging this PR.

- I would like Cards to automatically drop the top-left rounded corner when used together with tabs and the first tab is selected. Let's see if I can manage to get this done.
- These changes still need further testing with LMS, via and client. I'll be doing that today.

src/components/data/ScrollBox.tsx
```diff
@@ -6,6 +6,8 @@ import ScrollContent from './ScrollContent';
 export type ScrollBoxProps = CompositeProps & {
   /** Turn off borders on outer container */
   borderless?: boolean;
+  /** Add rounded corners to scrollable container */
+  rounded?: boolean;
```
acelaya, line 10: It's hard to imagine the different scenarios in which we use scrollable containers, so I'm adding rounded corners as an optional opt-in feature

src/tailwind.preset.js
```diff
@@ -18,8 +18,13 @@ export default /** @type {Partial<import('tailwindcss').Config>} */ ({
...
       borderRadius: {
-        // Equivalent to tailwind default `rounded-sm` size
-        DEFAULT: '0.125rem',
+        // Equivalent to tailwind default `rounded-sm` size.
+        // These should be overwritten by downstream projects in order to
+        // "enable" rounded corners.
+        // Next major version will remove this config, effectively enabling
+        // rounded corners.
+        lg: 'var(--h-border-radius-lg, 0.125rem)',
+        DEFAULT: 'var(--h-border-radius, 0.125rem)',
```
acelaya, line 27: Downstream projects should not see any changes when updating the library, but they can overwrite the rounded border values via `--h-border-radius` and `--h-border-radius-lg` vars.

We will need to find a way to define CSS vars based on feature flags.

styles/library.scss
```diff
@@ -2,6 +2,15 @@
  * This stylesheet contains styles for pattern-library components (only).
  */
 
+@layer base {
+  :root {
+    // This is the same as default tailwind values
+    // Once our overwrites have been removed from the preset, this can be dropped
+    --h-border-radius: 0.25rem;
+    --h-border-radius-lg: 0.5rem;
```
acelaya, line 10: I have enabled rounded corners in the pattern library.

acelaya:
Tested the changes in client. It looks good, and there's no changes in the UI.

I also verified we can add something like this in our styles:

```css
html[data-rounded-corners] {
  --h-border-radius: 0.25rem;
  --h-border-radius-lg: 0.5rem;
}
```

And then use a feature flag to conditionally add the `data-rounded-corners` attribute to the html node. Something in those lines should allow us to use a feature flag to enable the feature conditionally.

jaredpdesigns:
@acelaya this is all sounding really good. I did run into the issue with tables, because the outermost element is the only convenient way to target sub-children which aren't rounded by default, wrapping it in a rounded parent sounds like a most effective solution. Great work!

src/components/data/ScrollBox.tsx
robertknight: To confirm, the current default of `false` matches the existing look?

src/pattern-library/components/patterns/data/TablePage.tsx
```diff
@@ -155,71 +156,76 @@ export default function TablePage() {
                 <code>false</code>
               </Library.InfoItem>
             </Library.Info>
-            <Library.Demo title="Table with stickyHeader and Scroll" withSource>
+            <Library.Demo
+              title="Table with stickyHeader and rounded Scroll"
```
robertknight, line 160: What was the reason for picking this example to add rounding to - because it was already customizing the scroll container?

src/pattern-library/components/patterns/data/TablePage.tsx
```diff
@@ -275,6 +281,59 @@ export default function TablePage() {
               </Table>
             </Library.Demo>
           </Library.Example>
+          <Library.Example title="borderless">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Set this boolean prop if you want to remove the table{"'"}s
+                outer borders, in case it is rendered on a container with its
```
robertknight, line 288:
```suggestion
                outer borders, in case it is rendered in a container with its
```

src/pattern-library/components/patterns/prototype/TabbedDialogPage.tsx
```diff
@@ -141,7 +141,9 @@ function TabbedDialog() {
                 Three
               </Tab>
             </TabListHeader>
-            <Card>
+            <Card
+              classes={selectedTab === 'one' ? 'rounded-tl-none' : undefined}
```
robertknight, line 145: Unrelated to the changes here, this component is called "TabbedDialogPage" but that doesn't match the Import/Export label in the sidebar. We also might want to extract this pattern as a component in future to make it easier to instantiate?

robertknight (approved):
LGTM as far as I can see in the pattern library.

src/components/data/ScrollBox.tsx
acelaya: Yes. It will continue being non-rounded (current behavior) unless actively opting in.

src/pattern-library/components/patterns/data/TablePage.tsx
acelaya: No particular reason. Just to avoid having so many very similar examples.

I'm open to keep this one as is and add an extra example with a rounded scroll.

src/pattern-library/components/patterns/prototype/TabbedDialogPage.tsx
acelaya: > We also might want to extract this pattern as a component in future to make it easier to instantiate?

Agree.

It would also make it easier to transparently handle the rounded corners when the first tab is selected, so that you don't need to remember every time this pattern is used.

PR #1290, opened 2023-10-04
Improve SelectNext arrow key navigation

acelaya:
Part of https://github.com/hypothesis/frontend-shared/issues/1285
Depends on https://github.com/hypothesis/frontend-shared/pull/1288

This PR's primary focus is to improve arrow key navigation, making sure the active option is the focused one when the listbox is opened, falling back to the first option if none is active.

Additionally, we add explicit `displayName`s for `SelectNext` and `SelectOption`, which don't have the expected value by default due to the `Object.assign(...)` approach to expose the main component and its sub-components.

### Testing steps

1. Check out this branch and start dev server (`make dev`)
2. Go to http://localhost:4001/input-select-next
3. Focus one of the selects, and press <kbd>ArrowDown</kbd>. Notice the list is open and the first option is focused.
4. Select some other option. The list should close.
5. Repeat step 3. The active option should be the one selected this time.

In `main`, the first option is the one selected when the list is open, no matter if there's an active option or not.

### DIFF
--- a/src/hooks/use-arrow-key-navigation.ts
+++ b/src/hooks/use-arrow-key-navigation.ts
@@ -77,7 +77,6 @@ export type UseArrowKeyNavigationOptions = {
  * [2] https://www.w3.org/TR/wai-aria-practices/#keyboard
  *
  */
-
 export function useArrowKeyNavigation(
   containerRef: RefObject<HTMLElement | undefined>,
   {
@@ -92,7 +91,7 @@ export function useArrowKeyNavigation(
   // Keep track of the element that was last focused by this hook such that
   // navigation can be restored if focus moves outside the container and then
   // back to/into it.
-  const lastFocusedItem = useRef<HTMLOrSVGElement | null>(null);
+  const lastFocusedItem = useRef<HTMLElement | null>(null);
 
   useEffect(() => {
     if (!containerRef.current) {
@@ -192,7 +191,16 @@ export function useArrowKeyNavigation(
       event.stopPropagation();
     };
 
-    updateTabIndexes(getNavigableElements(), 0, containerVisible && autofocus);
+    const navigableElements = getNavigableElements();
+    // Start focus sequence with previously focused element, if any
+    const initialIndex = lastFocusedItem.current
+      ? navigableElements.indexOf(lastFocusedItem.current)
+      : 0;
+    updateTabIndexes(
+      navigableElements,
+      initialIndex,
+      containerVisible && autofocus,
+    );
 
     const listeners = new ListenerCollection();
 
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -73,6 +73,8 @@ function SelectOption<T>({
   );
 }
 
+SelectOption.displayName = 'SelectNext.Option';
+
 function useShouldDropUp(
   buttonRef: RefObject<HTMLElement | undefined>,
   listboxRef: RefObject<HTMLElement | null>,
@@ -234,6 +236,8 @@ function SelectMain<T>({
     </div>
   );
 }
+
+SelectMain.displayName = 'SelectNext';
 
 const SelectNext = Object.assign(SelectMain, { Option: SelectOption });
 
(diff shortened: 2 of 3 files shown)


src/hooks/use-arrow-key-navigation.ts
```diff
@@ -92,7 +91,7 @@ export function useArrowKeyNavigation(
   // Keep track of the element that was last focused by this hook such that
   // navigation can be restored if focus moves outside the container and then
   // back to/into it.
-  const lastFocusedItem = useRef<HTMLOrSVGElement | null>(null);
+  const lastFocusedItem = useRef<HTMLElement | null>(null);
```
robertknight, line 94: Is this because `getNavigableElements` returns `HTMLElement[]`?

src/hooks/use-arrow-key-navigation.ts
```diff
@@ -192,7 +191,16 @@ export function useArrowKeyNavigation(
...
+    const navigableElements = getNavigableElements();
+    // Start focus sequence with previously focused element, if any
+    const initialIndex = lastFocusedItem.current
+      ? navigableElements.indexOf(lastFocusedItem.current)
+      : 0;
+    updateTabIndexes(
+      navigableElements,
+      initialIndex,
+      containerVisible && autofocus,
+    );
```
robertknight, line 203: To recap: Every time this effect ran, it reset the selected item to the first one in the list. Now it will reset it to `lastFocusedItem.current`. This is an improvement over the way it worked before, although `lastFocusedItem` doesn't currently get updated if you focus an item a different way, eg using the mouse. In the context of `SelectNext`, this means that if you:

1. Open the list and select an item using the keyboard
2. Re-open the list and select a different item using the mouse
3. Re-open the list using the keyboard

Then the focused item after step (3) is the one that was focused in step (1) rather than step (2).

robertknight: Related to this, the docs for the `autofocus` prop say that focusing is applied when the component is mounted, but the current behavior is that auto-focusing is applied when:

- The component is mounted
- `containerVisible` changes
- Any of the general configuration props for the hook change (we expect this will not happen for the majority of uses)

robertknight:
Hmm, I'm seeing some odd behavior in Safari when testing in http://localhost:4001/input-select-next:

1. Click drop-down menu for the first select control in the page
2. Press the down arrow key, nothing is highlighted
3. Press it several more times. Eventually the bottom-most item ("Doris ...") is highlighted. None of the items before that are highlighted.

If you press Enter to select an item, the control behaves as if the expected item had been focused given the number of arrow presses, so it looks like this might be an issue with the focus ring.

robertknight (approved):
I did find an issue with the focus highlight when mixing mouse and keyboard input in Safari, and I did note an issue when mixing mouse and keyboard input in all browsers. Nevertheless this is an improvement over the previous behavior for the common case of using exclusively mouse or exclusively keyboard input.

src/hooks/use-arrow-key-navigation.ts
acelaya: In essence yes, but I could not find a reason for the ref to be typed as `HTMLOrSVGElement`.

I'm ok doing an inline `as HTMLElement` cast though. We are already doing it in a couple of other places for this hook.

src/hooks/use-arrow-key-navigation.ts
acelaya: > To recap: Every time this effect ran, it reset the selected item to the first one in the list. Now it will reset it to `lastFocusedItem.current`. This is an improvement over the way it worked before, although `lastFocusedItem` doesn't currently get updated if you focus an item a different way, eg using the mouse. In the context of `SelectNext`, this means that if you:
> 
> 1. Open the list and select an item using the keyboard
> 2. Re-open the list and select a different item using the mouse
> 3. Re-open the list using the keyboard
> 
> Then the focused item after step (3) is the one that was focused in step (1) rather than step (2).

Hmm, good catch.

There's another problem still to be solved, also related with combining mouse and keyboard (see last checkpoint in https://github.com/hypothesis/frontend-shared/issues/1285).

I'll try to address those together.

Also, I have checked that the behavior you describe does not happen with other arrow-key-navigable components, like `Tabs`. If you focus one tab via arrows, then click another tab, then focus away, then focus back, the tab that gets focused is the one you clicked. So at least the bug is limited to the Select.

PR #1292, opened 2023-10-04
Fix SelectNext focus inconsistencies when mixing mouse and keyboard interaction

acelaya:
Closes https://github.com/hypothesis/frontend-shared/issues/1285

This PR addresses the two problems left in `SelectNext`, both related with option focusing.

* Now, the focus ring is visible for the first option in the listbox when open. The option was previously focused already, but the style was applied to `:focus-visible` and not `:focus`, which prevents the UA from applying the styles. See first example in https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#examples
* When an option is clicked, we now make sure it is focused just before closing the listbox, becoming the last focused element. This ensures it is focused when the listbox is opened again, regardless of how the last option was selected (via mouse or keyboard).

In order to fix the first issue, options are no longer `Button` components, but simple `li` where we have to override less styles and we can apply focus rings to `:focus` instead of `:focus-visible`.

This also means the listbox is now a `ul` instead of a `div`.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -46,15 +46,20 @@ function SelectOption<T>({
   const selected = !disabled && currentValue === value;
 
   return (
-    <Button
-      variant="custom"
-      classes={classnames(
-        'w-full ring-inset rounded-none !p-0',
-        'border-t first:border-t-0 bg-transparent',
+    <li
+      className={classnames(
+        'w-full ring-inset focus:ring outline-none rounded-none cursor-pointer',
+        'border-t first:border-t-0 transition-colors whitespace-nowrap',
         { 'hover:bg-grey-1': !disabled },
         classes,
       )}
       onClick={() => selectValue(value)}
+      onKeyPress={e => {
+        if (e.code === 'Enter' || e.code === 'Space') {
+          e.preventDefault();
+          selectValue(value);
+        }
+      }}
       role="option"
       disabled={disabled}
       aria-selected={selected}
@@ -69,7 +74,7 @@ function SelectOption<T>({
       >
         {children({ selected, disabled })}
       </div>
-    </Button>
+    </li>
   );
 }
 
@@ -139,7 +144,7 @@ function SelectMain<T>({
   const [listboxOpen, setListboxOpen] = useState(false);
   const closeListbox = useCallback(() => setListboxOpen(false), []);
   const wrapperRef = useRef<HTMLDivElement | null>(null);
-  const listboxRef = useRef<HTMLDivElement | null>(null);
+  const listboxRef = useRef<HTMLUListElement | null>(null);
   const listboxId = useId();
   const buttonRef = useSyncedRef(elementRef);
   const defaultButtonId = useId();
@@ -168,6 +173,7 @@ function SelectMain<T>({
     loop: false,
     autofocus: true,
     containerVisible: listboxOpen,
+    selector: '[role="option"]',
   });
 
   useLayoutEffect(() => {
@@ -209,7 +215,7 @@ function SelectMain<T>({
         </div>
       </Button>
       <SelectContext.Provider value={{ selectValue, value }}>
-        <div
+        <ul
           className={classnames(
             'absolute z-5 w-full max-h-80 overflow-y-auto',
             'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
@@ -231,7 +237,7 @@ function SelectMain<T>({
           data-testid="select-listbox"
         >
           {children}
-        </div>
+        </ul>
       </SelectContext.Provider>
     </div>
   );
--- a/src/hooks/test/use-arrow-key-navigation-test.js
+++ b/src/hooks/test/use-arrow-key-navigation-test.js
@@ -380,21 +380,23 @@ describe('useArrowKeyNavigation', () => {
         container,
       ),
     );
-    const toggleToolbar = () => findElementByTestId('toggle').click();
+    const toggleToolbar = () =>
+      act(() => findElementByTestId('toggle').click());
 
     // No button should be initially focused
     assert.equal(document.activeElement, document.body);
 
     // Once we open the toolbar, the first item will be focused
-    toggleToolbar();
+    await toggleToolbar();
     await waitFor(() => document.activeElement === findElementByTestId('bold'));
 
     // If we then focus another toolbar item, then close the toolbar and open it
     // again, that same element should be focused again
     pressKey('ArrowDown'); // "italic" is focused
     pressKey('ArrowDown'); // "underline" is focused
-    toggleToolbar(); // Close toolbar
-    toggleToolbar(); // Open toolbar again
+    await toggleToolbar(); // Close toolbar
+    await toggleToolbar(); // Open toolbar again
+
     await waitFor(
       () => document.activeElement === findElementByTestId('underline'),
     );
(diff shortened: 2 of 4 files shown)


src/components/input/Button.tsx
```diff
@@ -96,8 +103,11 @@ export default function Button({
       {...htmlAttributes}
       className={classnames(
         {
-          'focus-visible-ring transition-colors whitespace-nowrap flex items-center':
-            styled,
+          'transition-colors whitespace-nowrap flex items-center': styled,
+        },
+        styled && {
+          'focus-visible-ring': focusRing === 'focus-visible',
+          'focus:ring outline-none': focusRing === 'focus',
```
acelaya, line 110: I don't love having to add a new property to the `Button`, but it's the only way to not affect other use cases.

Basically, the custom `focus-visible-ring` aims to style focused elements based on `:focus-visible`, instead of `:focus`.

Those two are equivalent for editable elements (like inputs or elements with `contenteditable`), but not for `buttons`.

A clicked button becomes focused, but does not become `focus-visible`, preventing the focus ring to render (See https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible for details and examples).

This effectively means that, if the `SelectNext` toggle is clicked, the listbox is opened, the first option is focused, but it does not render a focus ring, causing the second option to render a focus ring when pressing <kbd>ArrowDown</kbd> for the first time.

By using tailwind's standard `focus:ring`, we apply the ring to `:focus`, making sure the first option renders a focus ring when the listbox is opened via mouse, and then we can interact with the list via arrow keys with an intuitive behavior.

src/hooks/use-arrow-key-navigation.ts
```diff
@@ -94,6 +94,10 @@ export function useArrowKeyNavigation(
   const lastFocusedItem = useRef<HTMLElement | null>(null);
 
   useEffect(() => {
+    if (!containerVisible) {
+      return () => {};
+    }
```
acelaya, line 99: This is not strictly related with the other changes on this PR, but I realized we can safely skip any action when `containerVisible` is false, making the rest of the logic easier to reason about.

src/hooks/use-arrow-key-navigation.ts
```diff
@@ -215,10 +215,11 @@ export function useArrowKeyNavigation(
         lastFocusedItem.current.focus();
         return;
       }
+
       const elements = getNavigableElements();
       const targetIndex = elements.indexOf(event.target as HTMLElement);
       if (targetIndex >= 0) {
-        updateTabIndexes(elements, targetIndex);
+        updateTabIndexes(elements, targetIndex, autofocus);
```
acelaya, line 222: This is what was causing the options to not me set as `lastFocusedItem` when being clicked instead of selected with the keyboard.

src/components/input/Button.tsx
```diff
@@ -27,6 +27,12 @@ type ComponentProps = {
    */
   title?: string;
 
+  /**
+   * Whether the focus ring should be applied to `:focus` or `:focus-visible`.
+   * Defaults to focus-visible.
+   */
+  focusRing?: 'focus' | 'focus-visible';
```
acelaya, line 34: > See https://github.com/hypothesis/frontend-shared/pull/1292/files#r1345926735 for more details on why this has been introduced

I think `focusRing` sounds like a boolean property. Perhaps a better name would be one of `focusRingMode`, `focusRingType` or `focusRingVariation`.

src/hooks/test/use-arrow-key-navigation-test.js
```diff
@@ -30,15 +30,14 @@ function Toolbar({ navigationOptions = {} }) {
...
-  const toggleVisible = useCallback(() => setVisible(prev => !prev), []);
   const options = useMemo(
     () => ({ ...navigationOptions, containerVisible: visible }),
     [navigationOptions, visible],
   );
 
   return (
     <>
-      <button data-testid="toggle" onClick={toggleVisible}>
+      <button data-testid="toggle" onClick={() => setVisible(!visible)}>
```
acelaya, line 40: This was an attempt to fix an uncovered line, but didn't help. I'll remove it.

acelaya:
There's a line not covered, which should be when `containerVisible` changes and the effect is re-triggered. However, the test is not changing this and I'm trying to figure out why.

I'll put this in draft state for now, just in case this ends up having deeper implications.

acelaya:
> There's a line not covered, which should be when `containerVisible` changes and the effect is re-triggered. However, the test is not changing this and I'm trying to figure out why.
> 
> I'll put this in draft state for now, just in case this ends up having deeper implications.

Ok, I was missing some `act` calls to make sure side effects were triggered when clicking a button.

src/components/input/Button.tsx
robertknight: If we're using a `Button` but overriding basically all the key aspects of it (style, semantics, behavior), then we should consider not using `Button` at all here. The complexity added for non-buttony use cases is going to make it increasingly difficult to maintain this.

FWIW the Headless UI and React Spectrum Picker examples we've been referencing don't use buttons.

acelaya: I knew you were going to suggest this 😅

Yeah, I'll do the changes. At this point it probably makes sense to use something else

src/components/input/SelectNext.tsx
```diff
@@ -46,15 +46,19 @@ function SelectOption<T>({
...
   return (
-    <Button
-      variant="custom"
-      classes={classnames(
-        'w-full ring-inset rounded-none !p-0',
-        'border-t first:border-t-0 bg-transparent',
+    <li
+      className={classnames(
+        'w-full ring-inset focus:ring outline-none rounded-none',
+        'border-t first:border-t-0 transition-colors whitespace-nowrap',
```
acelaya, line 52: Moving for `Button` to `li` allows for less style overrides to be needed, and `focus:ring` to be used instead of `focus-visible-ring`

src/components/input/SelectNext.tsx
```diff
@@ -46,15 +46,19 @@ function SelectOption<T>({
...
         { 'hover:bg-grey-1': !disabled },
         classes,
       )}
       onClick={() => selectValue(value)}
+      onKeyPress={e => {
+        if (e.key === 'Enter') {
+          selectValue(value);
+        }
+      }}
```
acelaya, line 61: This was transparently handled by `Buttons`, but needs to be explicitly added for `li`s.

src/components/input/SelectNext.tsx
robertknight: One other style that came for free with buttons was setting `cursor: pointer` (or the equivalent TW style, that will need to be added here too).

src/components/input/SelectNext.tsx
```diff
@@ -46,15 +46,19 @@ function SelectOption<T>({
...
+    <li
+      className={classnames(
+        'w-full ring-inset focus:ring outline-none rounded-none',
+        'border-t first:border-t-0 transition-colors whitespace-nowrap',
         { 'hover:bg-grey-1': !disabled },
         classes,
       )}
       onClick={() => selectValue(value)}
+      onKeyPress={e => {
+        if (e.key === 'Enter') {
```
robertknight, line 58: Also the Space key should select the item.

robertknight (commented):
Just two minor things I spotted with the change to `<li>`:

- Space should select the item as well as Enter (I'm not sure if there is a more general way of getting the keys which have activation behavior)
- The cursor should be set to `pointer`

robertknight (approved):

Issue #1254, opened 2023-09-22
SelectNext: support searching/filtering options


acelaya:
Add some way to do text search inside the `SelectNext` options, as sketched here https://github.com/hypothesis/client/issues/5775#issuecomment-1698491323

Based on how the `SelectNext` component was architected, one way to support filtering items would be introducing a new `<Select.Filterable />` wrapper for `<Select.Option />` items, which allows children to be filtered via context.

```tsx
function App() {
  const [selected, setSelected] = useState<(typeof items)[number]>();

  return (
    <Select {...}>
      <Select.Option value={...}>Main</Select.Option>

      <Select.Filterable onFilter={(query: string, value: T) => boolean} placeholder={...}>
        {items.map(item => (
          <Select.Option value={item} key={item.id}>
            {({ disabled, isSelected }) => <>{item.name} ({item.id})</>}
          </Select.Option>
        ))}
      </Select.Filterable>
    </Select>
  )
}
```

Then we can expose the `onFilter` callback via context, down to the `Option` entries, which can decide if they should render or not, based on that.

We could even allow multiple Filterable blocks, or options outside any Filterable, that cannot be filtered. Ideally the search box should be position sticky, making sure it's visible while scrolling between the options it can filter.

The `Filterable` component could be something in these lines:

```tsx
const FilterableContext = createContext<{ shouldRender: (value: T) => boolean } | null>(null);

function Filterable<T>({ onFilter, placeholder, children }) {
  const [query, setQuery] = useState('');
  const shouldRender = useCallback(
    (value: T) => onFilter(query, value), 
    [onFilter, query],
  );

  return (
    <FilterableContext.Provider value={{ shouldRender }}>
      <input type="search" placeholder={placeholder} value={query} onInput={e => setQuery(e.target.value)} />
      {children}
    </FilterableContext.Provider>
  );
}
```

acelaya:
Closing in favor of https://github.com/hypothesis/frontend-shared/issues/1315

PR #1294, opened 2023-10-09
[POC] Add filtering capabilities to SelectNext

acelaya:
> Part of https://github.com/hypothesis/frontend-shared/issues/1254

This PR adds a new `SelectNext.Filterable` component, designed to wrap `SelectNext.Options` objects, and allow them to be filtered via text query.

[Grabación de pantalla desde 2023-10-09 11-18-50.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/71829a1c-87af-4efe-97c3-365038d8e6bc)

### Usage

Proposal was described in Part of https://github.com/hypothesis/frontend-shared/issues/1254

### TODO

- [ ] Currently changing the input content triggers a re-render, which makes every option call `shouldRender` as expected.
  However, this is not very obvious. There should be a more direct way to call `shouldRender` instead of relying on it being called as a side effect.
- [ ] Since Options are responsible for their own rendering, there's no easy way to tell if none are being rendered in order to display a "No results" placeholder.
- [ ] It might make sense to debounce the invokation to `onFilter`
- [ ] The original design includes a magnifying glass icon in the search box. We don't have one yet.
- [ ] We might want to customize the native "clear" button for `input[type="search"]`. Maybe using a text input with a custom clear button would be an option.

### Testing steps

This can be tested in http://localhost:4001/input-select-next, with the `Select with filtering` example.

### DIFF
--- a/src/components/input/SelectFilterableContext.ts
+++ b/src/components/input/SelectFilterableContext.ts
@@ -1,0 +1,10 @@
+import { createContext } from 'preact';
+
+export type SelectFilterableContextType<T = unknown> = {
+  shouldRender: (value: T) => boolean;
+};
+
+const SelectFilterableContext =
+  createContext<SelectFilterableContextType | null>(null);
+
+export default SelectFilterableContext;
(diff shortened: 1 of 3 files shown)


acelaya:
Closing this for now. It can serve as an example if we decide to come back to it at some point.

PR #1312, opened 2023-10-10
Deprecate useUniqueId and recommend preact's useId instead

### LINKED ISSUE #838: Deprecate `useUniqueId`, recommending the direct use of preact's `useId` hook
Internal hook `useUniqueId` was written before the existence of `useId` in preact. The custom hook does have a nice affordance in that it supports a prefix. Update it to use `useId` instead of a local reference for ID generation.

acelaya:
Closes #838

### DIFF
--- a/src/hooks/use-unique-id.ts
+++ b/src/hooks/use-unique-id.ts
@@ -6,6 +6,8 @@ let idCounter = 0;
  * instance.
  *
  * This avoids different instances of a component re-using the same ID.
+ *
+ * @deprecated Use Preact's `useId()` instead.
  */
 export function useUniqueId(prefix: string) {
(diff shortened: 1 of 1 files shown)


src/hooks/use-unique-id.ts
```diff
@@ -6,6 +6,8 @@ let idCounter = 0;
  * instance.
  *
  * This avoids different instances of a component re-using the same ID.
+ *
+ * @deprecated Use preact's `useId()` instead.
```
robertknight, line 10:
```suggestion
 * @deprecated Use Preact's `useId()` instead.
```

robertknight (approved):

PR #1314, opened 2023-10-10
Do not unmount SelectNext chevron when toggling listbox

acelaya:
This PR fixes a bug on `SelectNext`, where clicking exactly over the chevron would not open the list.

The reason is that we were mounting only one chevron icon depending if the list is open or closed. This was causing `onClickAway` to think a foreign element was clicked, as it is no longer in the DOM by the time the event bubbles.

In order to fix that behavior, this PR makes sure both chevrons are always mounted in the DOM, but hides the non-relevant one via CSS.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -215,7 +215,14 @@ function SelectMain<T>({
         {buttonContent ?? label}
         <div className="grow" />
         <div className="text-grey-6">
-          {listboxOpen ? <MenuCollapseIcon /> : <MenuExpandIcon />}
+          <MenuCollapseIcon
+            className={classnames({
+              // Hide instead of unmount to avoid `onClickAway` from thinking a
+              // foreign component was clicked and close the listbox
+              hidden: !listboxOpen,
+            })}
+          />
+          <MenuExpandIcon className={classnames({ hidden: listboxOpen })} />
         </div>
       </Button>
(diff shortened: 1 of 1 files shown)


robertknight (commented):
I wonder if we can handle this subtle issue in `useClickAway` itself, since it seems likely that it might occur in some other context.

Some possible options:

1. Use capture listeners, so that we check whether the clicked element is a descendant of the container before it has responded to the click. This is the easiest to implement, but has the downside that `event.stopPropagation()` will no longer be able to prevent the `useClickAway` callback from triggering.
2. Check [event.composedPath()](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath#browser_compatibility) to see whether the container is in the chain of objects (elements + the top-level `Window`) due to receive the click event. If it is, that means the click occurred inside the container. Otherwise the click occurred outside. This path is captured at the point when the click occurs AFAIK, so should reflect the DOM structure at that point.

acelaya:
> I wonder if we can handle this subtle issue in useClickAway itself, since it seems likely that it might occur in some other context.

I like that much better. I'll explore the options you suggest.

acelaya:
I can confirm using `event.composedPath().includes(container)` instead of `container.contains(event.target)` works like a charm, even with elements that were removed from the DOM.

I'll create the PR, and add an extra test case covering this, since all existing tests pass.

PR #1316, opened 2023-10-11
Use `event.composePath()` in `useClickAway` to determine if component is in container

acelaya:
This fixes a bug in which, if an element inside the container handled by `useClickAway` is clicked, and immediately removed from the DOM, the hook would think an element outside the container was clicked, and incorrectly call the callback.

In order to solve this, we check if the container is part of the event's `composePath()`, which is a list of the target element and all its parents at the moment of invoking the event. This will be true even if the target itself has been removed from the DOM before the event handler is invoked.

### DIFF
--- a/src/hooks/test/use-click-away-test.js
+++ b/src/hooks/test/use-click-away-test.js
@@ -6,8 +6,10 @@ import { useClickAway } from '../use-click-away';
 
 describe('useClickAway', () => {
   let handler;
+  let wrapper;
 
-  const events = [new Event('mousedown'), new Event('click')];
+  const event = name => new Event(name, { bubbles: true });
+  const events = [event('mousedown'), event('click')];
 
   // Create a fake component to mount in tests that uses the hook
   function FakeComponent({ enabled = true }) {
@@ -22,11 +24,19 @@ describe('useClickAway', () => {
   }
 
   function createComponent(props) {
-    return mount(<FakeComponent {...props} />);
+    const container = document.createElement('div');
+    document.body.append(container);
+
+    wrapper = mount(<FakeComponent {...props} />, { attachTo: container });
+    return wrapper;
   }
 
   beforeEach(() => {
     handler = sinon.stub();
+  });
+
+  afterEach(() => {
+    wrapper.unmount();
   });
 
   events.forEach(event => {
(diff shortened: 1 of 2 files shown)


src/hooks/use-click-away.ts
```diff
@@ -28,7 +28,7 @@ export function useClickAway(
     const handleAwayClick = (event: Event) => {
       if (
         container.current &&
-        !container.current.contains(event.target as Node)
+        !event.composedPath().includes(container.current)
```
robertknight, line 31: This definitely deserves a comment to explain why we're not using the more obvious `container.current.contains(event.target)` test. Something like:

```js
// We test the composed path here to handle the case where the clicked element is
// removed from the DOM (eg. by a re-render in a child component) before this callback is run.
// The composed path reflects the DOM hierarchy at the time the event was dispatched.
```

src/hooks/test/use-click-away-test.js
```diff
@@ -62,7 +60,23 @@ describe('useClickAway', () => {
       });
       wrapper.update();
 
-      assert.equal(handler.callCount, 0);
+      assert.notCalled(handler);
+    });
+
+    it(`should not invoke callback on elements removed from container (${event.type})`, () => {
```
robertknight, line 66:
```suggestion
    it(`should not invoke callback if clicked element is removed from container (${event.type})`, () => {
```

The previous wording to me suggested the callback being invoked with the element as an argument.

src/hooks/test/use-click-away-test.js
```diff
@@ -62,7 +60,23 @@ describe('useClickAway', () => {
...
+      const wrapper = createComponent();
+      const button = wrapper.find('button').getDOMNode();
+
+      act(() => {
+        button.addEventListener(event.type, () => {
+          // Remove the button from the DOM before next listeners are invoked
```
robertknight, line 72: I would suggest to note this simulates what happened if the clicked element was removed as a result of a state update + re-render in a child component.

src/hooks/test/use-click-away-test.js
```diff
@@ -62,7 +70,25 @@ describe('useClickAway', () => {
...
+          // Remove the button from the DOM before next listeners are invoked,
+          // to simulate what happens if the clicked element was removed as a
+          // result of a state update + re-render in a child component
+          button.remove();
+        });
+        button.dispatchEvent(event);
+      });
+      wrapper.update();
+
+      assert.notCalled(handler);
```
acelaya, line 91: This test does not pass with the previous logic

robertknight (approved):
Sweet 👍

PR #1317, opened 2023-10-11
Add inherited rounded corners to SelectNext toggle

acelaya:
While trying to fix the [SelectNext when used inside an InputGroup](https://github.com/hypothesis/frontend-shared/pull/1311), I introduced another bug, which made the toggle button itself to not have rounded corners, and overlap with the wrapper ones.

This was not directly obvious with the subtle rounded corners we have now, so I have increased their size to demonstrate the problem:

Before:

[Grabación de pantalla desde 2023-10-11 11-47-24.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/5da322e7-73c8-4e55-99c6-77e7bd9517e8)

After

[Grabación de pantalla desde 2023-10-11 11-46-50.webm](https://github.com/hypothesis/frontend-shared/assets/2719332/fe27e8b8-354a-4949-a98c-a2fa763b4184)

> You'll notice there's also a weird overlapping on the focus ring for first and last options when the listbox has larger rounded corners. I'll address that separately.


src/components/input/SelectNext.tsx
```diff
@@ -196,6 +196,11 @@ function SelectMain<T>({
         classes={classnames(
           'w-full flex',
           'bg-grey-0 disabled:bg-grey-1 disabled:text-grey-6',
+          // Adding inherited rounded corners so that the toggle is consistent
```
robertknight, line 199:
```suggestion
          // Add inherited rounded corners so that the toggle is consistent
```

robertknight (approved):

PR #1325, opened 2023-10-16
Handle overflow in SelectNext

### LINKED ISSUE #1249: SelectNext: Handle too long content in options or select

Implement some logic to handle too long content in the options or the select itself, but taking into consideration both support wrapping rich content (`ComponentChildren`).

It is also ok to provide the right tools to delegate it to consumers.

acelaya:
> Closes #1249 

This PR is trying to make `SelectNext` behave closer to the native `select` when the items overflow the size of the select itself.

For context, this is how a native select looks in that case:

Captura desde 2023-10-16 15-33-07

* The content in the select itself is cropped, to make sure it does not overflow its container.
* The listbox grows to fit all options without cropping them. This is important, as there has to be a place where you can see the full content for every option.

Taking that into consideration, the next changes have been made.

1. The listbox has changed from `w-full` to `min-w-full`. That way it keeps looking the same as before in most of the cases, but it grows if the content is too long.

    image
    image

2. The toggle button now truncates its content. That ensures there won't be content overflow, and displays ellipsis if the content is just plain text.
  When the content is not plain text, consumers might need to adapt their code in order to make it look nice. An example of this is included in the pattern library docs.

    image

### Testing steps.

You can check the new examples at the bottom of http://localhost:4001/input-select-next

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -194,7 +194,7 @@ function SelectMain<T>({
         id={buttonId ?? defaultButtonId}
         variant="custom"
         classes={classnames(
-          'w-full flex',
+          'w-full flex justify-between',
           'bg-grey-0 disabled:bg-grey-1 disabled:text-grey-6',
           // Add inherited rounded corners so that the toggle is consistent with
           // the wrapper, which is the element rendering borders.
@@ -217,8 +217,7 @@ function SelectMain<T>({
         }}
         data-testid="select-toggle-button"
       >
-        {buttonContent ?? label}
-        <div className="grow" />
+        <div className="truncate">{buttonContent ?? label}</div>
         <div className="text-grey-6">
           {listboxOpen ? <MenuCollapseIcon /> : <MenuExpandIcon />}
         </div>
@@ -226,7 +225,7 @@ function SelectMain<T>({
       <SelectContext.Provider value={{ selectValue, value }}>
         <ul
           className={classnames(
-            'absolute z-5 w-full max-h-80 overflow-y-auto',
+            'absolute z-5 min-w-full max-h-80 overflow-y-auto',
             'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
             {
               'top-full mt-1': !shouldListboxDropUp,
--- a/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
@@ -17,10 +17,12 @@ const defaultItems = [
 function SelectExample({
   disabled,
   textOnly,
+  classes,
   items = defaultItems,
 }: {
   disabled?: boolean;
   textOnly?: boolean;
+  classes?: string;
   items?: typeof defaultItems;
 }) {
   const [value, setValue] = useState<(typeof items)[number]>();
@@ -29,13 +31,17 @@ function SelectExample({
     <SelectNext
       value={value}
       onChange={setValue}
+      classes={classes}
       buttonContent={
         value ? (
           <>
-            {value.name}
+            {textOnly && value.name}
             {!textOnly && (
-              <div className="rounded px-2 bg-grey-7 text-white">
-                {value.id}
+              <div className="flex">
+                <div className="truncate">{value.name}</div>
+                <div className="rounded px-2 ml-2 bg-grey-7 text-white">
+                  {value.id}
+                </div>
               </div>
             )}
           </>
@@ -56,7 +62,7 @@ function SelectExample({
               <>
                 {item.name}
                 <div className="grow" />
-                <div className="rounded px-2 bg-grey-7 text-white">
+                <div className="rounded px-2 ml-2 bg-grey-7 text-white">
                   {item.id}
                 </div>
               </>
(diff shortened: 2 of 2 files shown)


src/components/input/SelectNext.tsx
```diff
@@ -194,7 +194,7 @@ function SelectMain<T>({
         id={buttonId ?? defaultButtonId}
         variant="custom"
         classes={classnames(
-          'w-full flex',
+          'w-full flex justify-between',
```
robertknight, line 197: What is the `justify-between` for here? To replace the `grow` spacer below?

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -216,11 +222,31 @@ export default function SelectNextPage() {
...
+
+          <Library.Example title="Select with long content">
+            <p>
+              <code>SelectNext</code> makes sure the content never overflows,
+              but you may need to tweak <code>buttonContent</code> or options
+              content in cases where more complex content needs to be rendered.
+            </p>
+
+            <Library.Demo title="Plain text">
+              <div className="mx-auto">
```
robertknight, line 239: The width and left position of the select control changes depending on the selected item. That looks weird to me, and it means that the dropdown arrow moves around. I think the widths of these controls is almost always fixed.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -216,11 +222,31 @@ export default function SelectNextPage() {
...
-              <div className="w-[350px] mx-auto">
+              <div className="w-96 mx-auto">
                 <SelectExample disabled />
               </div>
             </Library.Demo>
           </Library.Example>
+
+          <Library.Example title="Select with long content">
+            <p>
+              <code>SelectNext</code> makes sure the content never overflows,
```
robertknight, line 233: Are you referring to the button or listbox content here?

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -216,11 +222,31 @@ export default function SelectNextPage() {
...
+              but you may need to tweak <code>buttonContent</code> or options
+              content in cases where more complex content needs to be rendered.
```
robertknight, line 235: What kind of "tweak" do you mean?

src/components/input/SelectNext.tsx
acelaya: Yes. The `text-overflow: ellipsis` didn't work on a `grow` element.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: I think I wrote this at one point when I had a different approach in mind than what I ended up pushing 😅 

I'll re-write it to make it more clear.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: I'll try to clarify it.

Basically, the toggle button provides some affordances in the form of `text-overflow: ellipsis` and `overflow: hidden`.

While the later is always going to apply no matter the type of children provided to `buttonContent`, the former only applies when it is "plain text" (so to speak).

If you want a more complex set of children to rearrange differently, it's up to the consumer to handle it.

For example, I have included an example in the pattern library where the children have its own `truncate` wrapper to make sure the badge with the number is always displayed and only the text is hidden.

What I was trying to convey is that every use case might be different and it's up to downstream projects to customize it if the default `overflow: hidden` is not right for them.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: > The width and left position of the select control changes depending on the selected item

This is precisely what I was trying to prevent. I think I may have broken in some of my latest changes before pushing the branch.

It should behave like in the second screenshot I added to the PR description. I'll try to figure it out.

acelaya: Yeah, I see the problem.

In order for this to work, the dimensions need to be somehow constrained, either by the `SelectNext` itself (like here) or its container. Otherwise, it can potentially grow.

In this case, since I'm trying to constrain it through `SelectNext`'s own `classes` prop, it needs to be `!w-36`, not `w-36`, otherwise the `w-full` it already has will take precedence.

Another option would be to constrain it through the container (I think that's what I did initially, when I took the screenshots), but it's important the other option also works, in case we need to use it inside an `InputGroup`.

Having to add the `important` modifier is not ideal (it means you need to know the internal details), but we can think on a better way to address that later.

acelaya:
@robertknight I have pushed the next changes.

* I have tried to improve the explanation of the component's behavior when content would overflow, trying to clarify the behaviors of the toggle button vs the listbox.
* I have fixed the miss-behavior you detected, where the button size changed depending on selected option, rather than hiding the overflowing content.
* I have added one more example where an `InputGroup` is used.

robertknight (approved):
LGTM

PR #1328, opened 2023-10-20
Fixes and improvements for SelectNext.Option

acelaya:
This PR introduces a couple of improvements and one fix for the `SelectNext.Option` components:

* It ensures `<SelectNext.Option disabled />` occurrences cannot be interacted with the mouse or keyboard.
  This was a regression introduced when moving from `Button` to `li`, as it worked out of the box for `<Button disabled />`.
* It adds some default style for disabled `SelectNext.Option`s.  
  
  image

* It adds support for regular children.
  So far, children had to be a callback which receives the `disabled` and `selected` statuses, in case you need to apply some customization, but while using the `SelectNext` I have found that in most of the cases we are ignoring those, and being able to pass regular children is more convenient.
  So, instead of having to do this:
  ```tsx
  <SelectNext.Option>
    {() => 'Some value'}
  </SelectNext.Option>
  ```
  We can now also do this:
  ```tsx
  <SelectNext.Option>
    Some value
  </SelectNext.Option>
  ```
* It adds the missing documentation for `SelectNext.Option` props in the pattern library, as we were only documenting the main `SelectNext` props.
  
  image

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -64,9 +64,12 @@ function SelectOption<T>({
   return (
     <li
       className={classnames(
-        'w-full ring-inset focus:ring outline-none rounded-none',
+        'w-full ring-inset outline-none rounded-none',
         'border-t first:border-t-0 transition-colors whitespace-nowrap',
-        { 'cursor-pointer hover:bg-grey-1': !disabled },
+        {
+          'text-grey-4': disabled,
+          'cursor-pointer focus:ring hover:bg-grey-1': !disabled,
+        },
         classes,
       )}
       onClick={() => {
@@ -81,7 +84,7 @@ function SelectOption<T>({
         }
       }}
       role="option"
-      disabled={disabled}
+      aria-disabled={disabled}
       aria-selected={selected}
       // This is intended to be focused with arrow keys
       tabIndex={-1}
@@ -193,7 +196,7 @@ function SelectMain<T>({
     loop: false,
     autofocus: true,
     containerVisible: listboxOpen,
-    selector: '[role="option"]',
+    selector: '[role="option"]:not([aria-disabled="true"])',
   });
 
   useLayoutEffect(() => {
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -28,9 +28,24 @@ export type SelectOptionStatus = {
 export type SelectOptionProps<T> = {
   value: T;
   disabled?: boolean;
-  children: (status: SelectOptionStatus) => ComponentChildren;
+  children:
+    | ComponentChildren
+    | ((status: SelectOptionStatus) => ComponentChildren);
   classes?: string | string[];
 };
+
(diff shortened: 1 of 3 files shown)


robertknight (changes requested):
Can you add a demo to the `SelectNext` example page that shows what a disabled _item_ looks like, alongside non-disabled items. I also think that there should be some default disabled appearance, as otherwise disabled items look the same as non-disabled ones unless the user of `SelectNext` takes special care to implement a disabled state. For example by changing the inherited color:

```diff
diff --git a/src/components/input/SelectNext.tsx b/src/components/input/SelectNext.tsx
index 9a7cb7d..9a75cd4 100644
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -67,6 +67,7 @@ function SelectOption<T>({
         'w-full ring-inset focus:ring outline-none rounded-none',
         'border-t first:border-t-0 transition-colors whitespace-nowrap',
         { 'cursor-pointer hover:bg-grey-1': !disabled },
+        { 'text-grey-5': disabled },
         classes,
       )}
       onClick={() => {
```

acelaya:
> Can you add a demo to the `SelectNext` example page that shows what a disabled _item_ looks like, alongside non-disabled items. I also think that there should be some default disabled appearance, as otherwise disabled items look the same as non-disabled ones unless the user of `SelectNext` takes special care to implement a disabled state. For example by changing the inherited color:
> 
> ```diff
> diff --git a/src/components/input/SelectNext.tsx b/src/components/input/SelectNext.tsx
> index 9a7cb7d..9a75cd4 100644
> --- a/src/components/input/SelectNext.tsx
> +++ b/src/components/input/SelectNext.tsx
> @@ -67,6 +67,7 @@ function SelectOption<T>({
>          'w-full ring-inset focus:ring outline-none rounded-none',
>          'border-t first:border-t-0 transition-colors whitespace-nowrap',
>          { 'cursor-pointer hover:bg-grey-1': !disabled },
> +        { 'text-grey-5': disabled },
>          classes,
>        )}
>        onClick={() => {
> ```

Good catch. The disabled appearance was also something coming from `Button`.

src/components/input/SelectNext.tsx
```diff
@@ -49,20 +64,27 @@ function SelectOption<T>({
   return (
     <li
       className={classnames(
-        'w-full ring-inset focus:ring outline-none rounded-none cursor-pointer',
+        'w-full ring-inset outline-none rounded-none',
         'border-t first:border-t-0 transition-colors whitespace-nowrap',
-        { 'hover:bg-grey-1': !disabled },
+        {
+          'text-grey-4': disabled,
```
acelaya, line 70: I have used `text-grey-4` instead of `text-grey-5` as suggested, as this is the color we use for other components when disabled.

robertknight (approved):
Test coverage is missing for the non-function branch in `optionChildren` and I had a query about moving the disabled select examples under the corresponding prop documentation. Otherwise LGTM.

Issue #1331, opened 2023-10-23
SelectNext: Refine documentation


acelaya:
> Originally posted in https://github.com/hypothesis/frontend-shared/pull/1330#pullrequestreview-1692129430

- At the top the component is described as a presentational component. I don't think that's really accurate any more. It isn't just styling around a native `<select>`.
- There is an example near the top that shows what it looks like, but there is no code view
- Underneath that is a "Working with SelectNext" section that references `SelectNext.Option` but at this point of reading through the page, there has been no explanation or demonstration of what an `Option` is or what the code structure looks like.

robertknight:
Related to this, the switches to toggle between UI preview and code are missing from all the examples, as we have for other components.

Docs: https://patterns.hypothes.is/input-select-next

`SelectNext` example:

<img width="857" alt="SelectNext" src="https://github.com/hypothesis/frontend-shared/assets/2458/3d131964-569a-42ee-a9e1-6ec65b51e019">

Compare with the old `Select` (note switch button on the right):

<img width="851" alt="Select" src="https://github.com/hypothesis/frontend-shared/assets/2458/b654393b-7776-4583-8313-5dc61e38fdd1">

PR #1340, opened 2023-10-24
Add individual code examples for SelectNext

acelaya:
This PR is an attempt to improve SelectNext's documentation, and close https://github.com/hypothesis/frontend-shared/issues/1331

The primary focus is to illustrate the different ways in which the component can be used by adding smaller and individual code examples in different sections of the pages, and removing the single bigger existing code example at the bottom.

This helps referencing concepts later in the docs, making sure they have already been illustrated beforehand.

### Considerations

The way this has been achieved is not ideal though (and that's why I'm creating this a draft).

The `Library.Demo` component already supports displaying code snippets with syntax highlight, based on code examples that are verified and we know are correct.

However, this component is limited to only render code for JSX, and not javascript, which is a problem, since in most of the `SelectNext` examples we need to show how different elements interact via hooks or external props.

Because of that, I have kept the interactive examples using `Library.Demo` and used `Library.Code` to illustrate the usage.

This has some drawbacks:

- The code snippets can become outdated, as there's no static checks over them. It's even possible they have syntax errors.
- We have to "duplicate" the code for the interactive examples and the code examples, and keep them in sync.

### Testing

The changes can be seen by checking out this branch and going to http://localhost:4001/input-select-next

### DIFF
--- a/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
@@ -173,8 +173,7 @@ export default function SelectNextPage() {
       title="SelectNext"
       intro={
         <p>
-          <code>SelectNext</code> is a presentational component which behaves
-          like the native <code>{'<select>'}</code> element.
+          <code>SelectNext</code> is a component that behaves like a combobox.
         </p>
       }
     >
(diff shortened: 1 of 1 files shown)


acelaya:
Closing this, as we have discussed some potential improvements for the pattern library that would make this easier to maintain in the long term.

The changes go in the direction of embedding some kind of interactive editor (like [jsfiddle](https://jsfiddle.net/)) which we can use to load individual example files that can be used as code usage examples, but at the same time they can get statically checked and tanspiled to generate the interactive examples.

This will help making sure the code examples are actually correct and up to date, while it won't require maintaining separate "snippets" for interactive examples.

PR #1341, opened 2023-10-28
Adjust coloring of checked Checkbox

jaredpdesigns:
To support work related to email digest settings, this PR changes the `fill` and `stroke` values for the checked state of the Checkbox.

### DIFF
--- a/src/components/icons/CheckboxChecked.tsx
+++ b/src/components/icons/CheckboxChecked.tsx
@@ -13,7 +13,6 @@ export default function CheckboxCheckedIcon(props: CheckboxCheckedIconProps) {
       height="16"
       aria-hidden="true"
       viewBox="-4 -4 39 39"
-      className="text-grey-1"
       data-component="CheckboxCheckedIcon"
       {...props}
     >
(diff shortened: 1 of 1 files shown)


jaredpdesigns:
@acelaya no problem:

**Before:**
Screenshot 2023-10-30 at 6 59 18 AM

**After:**
Screenshot 2023-10-30 at 6 59 57 AM

src/components/icons/CheckboxChecked.tsx
```diff
@@ -13,6 +13,7 @@ export default function CheckboxCheckedIcon(props: CheckboxCheckedIconProps) {
       height="16"
       aria-hidden="true"
       viewBox="-4 -4 39 39"
+      className="text-grey-1"
```
acelaya, line 16: This class introduces an undesired change on the icon when used standalone.

image

However, I can remove this class and the icon still has the new look-and-feel when used in the `Checkbox` component.

Can you clarify the context of this change, just to make sure I'm not missing anything?

```suggestion
```

acelaya (changes requested):
I think the change makes sense for the `Checkbox` component, as the difference between checked and unchecked was too subtle before.

However, I noticed a small visual glitch in the `CheckboxChecked` icon caused by the `text-grey-1` class and would like to understand that change.

src/components/icons/CheckboxChecked.tsx
jaredpdesigns: @acelaya I removed this, originally I wanted to ensure the color of the checkbox would be consistent (i.e. dark gray) instead of taking whatever the parent's color was (i.e. `currentColor`), but I think it's preferable to be flexible.

acelaya (approved):
LGTM

PR #1353, opened 2023-11-07
Add `rounded` prop to `OptionButton` to make it optionally consistent with rounded corners

acelaya:
This was an overlook from https://github.com/hypothesis/frontend-shared/pull/1289

This PR adds a `rounded` property to `OptionButton`, so that we can optionally make it rounded.

The property defaults to `false` for backward compatibility, but is deprecated indicating the intention is to make all buttons rounded by default.

image


robertknight (approved):
LGTM, although this change is subtle enough that I think you could skip making it optional if you wanted to.

PR #1354, opened 2023-11-07
Propagate borderless prop from DataTable to Table

acelaya:
This was an overlook from https://github.com/hypothesis/frontend-shared/pull/1289

Expose a `borderless` prop from `DataTable` that can be propagated to the `Table` component.


src/components/data/DataTable.tsx
```diff
@@ -40,6 +40,9 @@ type ComponentProps<Row> = {
   /** Callback to render an individual table cell */
   renderItem?: (r: Row, field: keyof Row) => ComponentChildren;
   title: string;
+
+  /** Turn off outer table borders */
+  borderless?: boolean;
```
robertknight, line 45: I was confused that this prop was only added in the props type and no executable code changed. I realized that this works because this prop is captured by `htmlAttributes`. That rest property name is now misleading since we're including `Table` props too. Perhaps we _should_ continue to pass this prop and any forwarded non-HTML props explicitly to avoid confusion.

src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -488,6 +488,35 @@ export default function DataTablePage() {
             </Library.Demo>
           </Library.Example>
 
+          <Library.Example title="borderless">
```
robertknight, line 491: I think ideally we would just link to the `Table.borderless` prop documentation here, to avoid the risk of it getting out of sync. This is awkward right now because individual prop docs in the pattern library don't have their own fragment headers that you can conveniently link to.

robertknight (approved):
LGTM, but see comment about whether we should pass this prop explicitly rather than through the now-misnamed `htmlAttributes`.

src/components/data/DataTable.tsx
acelaya: Yep, makes sense.

PR #1357, opened 2023-11-09
Make it easier to prevent unnecessary `DataTable` re-renders

robertknight:
While adding some new UI to the VitalSource chapter/page selection dialog (see https://github.com/hypothesis/lms/issues/5808) I found that the UI got very slow when working with a book that had a large (> 1000) number of entries in the chapter list. Such a large chapter list could really use virtualization to limit rendering time, but memo-ing is much easier to add and still has a significant impact.

I tried memo-ing the `DataTable` but that initially had no effect, due to the parent `Scroll` triggering a re-render even if the `DataTable` was memoed. This PR fixes that issue, and adds a bunch of additional memo-ing inside `DataTable` to make it easier for downstream apps to prevent re-renders. They now need to just memo: 1) The rows, 2) the columns, 3) the selected item. For better results they can also memo the whole `DataTable`.

In the process I imported the [useStableCallback](https://github.com/hypothesis/via/pull/1129) hook from Via since it is generally useful for these kinds of optimizations and I expect it will be useful in other applications in future.

### DIFF
--- a/src/components/data/DataTable.tsx
+++ b/src/components/data/DataTable.tsx
@@ -1,7 +1,8 @@
 import type { ComponentChildren, JSX } from 'preact';
-import { useContext, useEffect } from 'preact/hooks';
+import { useCallback, useContext, useEffect, useMemo } from 'preact/hooks';
 
 import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
+import { useStableCallback } from '../../hooks/use-stable-callback';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
 import type { CompositeProps } from '../../types';
 import { downcastRef } from '../../util/typing';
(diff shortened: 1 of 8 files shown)


src/components/data/DataTable.tsx
```diff
@@ -49,6 +50,10 @@ export type DataTableProps<Row> = CompositeProps &
   ComponentProps<Row> &
   Omit<JSX.HTMLAttributes<HTMLElement>, 'size' | 'rows' | 'role' | 'loading'>;
 
+function defaultRenderItem<Row>(r: Row, field: keyof Row): ComponentChildren {
+  return r[field] as ComponentChildren;
+}
```
robertknight, line 55: Hoisted this default prop value to a top-level function so it doesn't change on each render.

src/components/data/DataTable.tsx
```diff
@@ -81,23 +86,25 @@ export default function DataTable<Row>({
...
-  function selectRow(row: Row) {
+  const selectRow = useStableCallback((row: Row) => {
     onSelectRow?.(row);
-  }
-
-  function confirmRow(row: Row) {
+  });
+  const confirmRow = useStableCallback((row: Row) => {
     onConfirmRow?.(row);
-  }
```
robertknight, line 92: Wrap callback props so we don't unnecessarily re-render even if consumers fail to `useCallback` these.

src/hooks/test/use-stable-callback-test.js
```diff
@@ -0,0 +1,39 @@
+// These tests use Preact directly, rather than via Enzyme, to simplify debugging.
```
robertknight, line 1: This file and `use-stable-callback.ts` were copied from Via.

acelaya (approved):
Good one 👍🏼

PR #1360, opened 2023-11-10
Allow user to select multiple rows in DataTable

robertknight:
Add `selectedRows` and `onSelectRows` props to `DataTable`. When `selectedRows` is passed, it enables multiple-selection for the table, allowing the user to select a contiguous range of rows via shift + click. Discontiguous selections via ctrl+click are not supported, as we don't want this capability for the initial use case.

Part of https://github.com/hypothesis/lms/issues/5808.


src/components/data/test/DataTable-test.js
```diff
@@ -140,20 +175,82 @@ describe('DataTable', () => {
   });
 
   describe('interacting with row data', () => {
-    ['click', 'focus'].forEach(event => {
```
robertknight, line 143: The "focus" test here was for testing what happens when the user alters the selection with arrow keys. I have replaced this with a more explicit test of arrow key interaction below.

src/components/data/test/DataTable-test.js
```diff
@@ -140,20 +175,82 @@ describe('DataTable', () => {
...
-        assert.calledWith(onSelectRow, fakeRows[0]);
+      wrapper.find('tbody').first().simulate('keydown', {
+        key: 'ArrowDown',
+        shiftKey: true,
       });
+      assert.calledWith(onSelectRows, [fakeRows[0], fakeRows[1]]);
     });
 
-    it('invokes provided `onConfirmRow` callback when row is double-clicked', () => {
+    it('invokes `onConfirmRow` callback when row is double-clicked', () => {
```
robertknight, line 253: The word "provided" is superfluous in these test descriptions.

acelaya (approved):
Looks good in general, and works as expected.

Just some thoughts:

* When selecting multiple rows via <kbd>shift</kbd> + <kbd>click</kbd>, it sometimes selects the text between those rows. Maybe we want to disallow selecting text in `DataTable`s, but perhaps this has other side effects.
* It would be good that the props API did not allow passing both `selectedRow` and `selectedRows`, but just one.
  However, this would require adding some type guards in order to read the props, so maybe it's better to just be pragmatic.

robertknight:
> When selecting multiple rows via shift + click, it sometimes selects the text between those rows. Maybe we want to disallow selecting text in DataTables, but perhaps this has other side effects.

Yes, I'd noticed this but hadn't settled how best to handle it yet. My initial thought was to allow downstream consumers to decide on a case-by-case basis.

> It would be good that the props API did not allow passing both selectedRow and selectedRows, but just one.

One possibility is to simply drop `selectedRow` in future and only have `selectedRows`, with a separate prop to control whether multi-selection is allowed or not. This would be a breaking change however.

acelaya:
> Yes, I'd noticed this but hadn't settled how best to handle it yet. My initial thought was to allow downstream consumers to decide on a case-by-case basis.

That's a good approach.

> One possibility is to simply drop `selectedRow` in future and only have `selectedRows`, with a separate prop to control whether multi-selection is allowed or not. This would be a breaking change however.

Let's do that. I'll create a ticket and add ti to v7 milestone.

PR #1369, opened 2023-11-14
Add props to SelectNext to granularly pass classess

### LINKED ISSUE #1319: SelectNext: Be able to pass classes to wrapper, button and listbox
Currently the components supports `classes`, and these are passed to the toggle button.

Sometimes you need to be able to customize either the wrapper (for sizing) or the listbox, so we should probably expose:

* `classes`: Goes to the wrapper element.
* `buttonClasses`: In line with `buttonId` and `buttonContent`. Goes to the toggle button.
* `listboxClasses`: Goes to the listbox.

acelaya:
Add `buttonClasses`, `containerClasses` and `listboxClasses` props to `SelectNext`, to have more control over how individual components are styled.

The `classes` props is now deprecated in favor of `buttonClasses`.

We have discussed this approach in previous PRs, but it was always possible to work around it by wrapping the `SelectNext` in other elements. However, the need for this is more obvious the more we use it.

Captura desde 2023-11-14 11-01-01

Closes #1319

### DIFF
--- a/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
@@ -28,7 +28,12 @@ function SelectExample({
   ...rest
 }: Pick<
   SelectNextProps<ItemType>,
-  'aria-label' | 'aria-labelledby' | 'classes' | 'disabled'
+  | 'aria-label'
+  | 'aria-labelledby'
+  | 'buttonClasses'
+  | 'containerClasses'
+  | 'listboxClasses'
+  | 'disabled'
 > & {
   textOnly?: boolean;
   items?: ItemType[];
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -173,8 +182,8 @@ export default function SelectNextPage() {
       title="SelectNext"
       intro={
         <p>
-          <code>SelectNext</code> is a presentational component which behaves
-          like the native <code>{'<select>'}</code> element.
+          <code>SelectNext</code> is a composite component which behaves like
+          the native <code>{'<select>'}</code> element.
```
robertknight, line 186: This part could do with an expansion at some point to explain how it differs from a `<select>`, as well as linking to the accessibility patterns it implements, specifically the [Select-only Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/).

robertknight (approved):
LGTM

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: There's an issue around improving this documentation. I'll add a reference to this comment to make sure it's not forgotten.

PR #1372, opened 2023-11-15
Add `right` prop to SelectNext

### LINKED ISSUE #1371: SelectNext: Allow listbox to be aligned to the right side when it's bigger than the toggle button
The listbox is designed to be at least as big as the toggle button, but it can be bigger if its contents are too long.

By default, the listbox aligns to the left in those cases, but if the select itself is right-aligned, the listbox should also align to the right.

right

acelaya:
Closes #1371

The first time we had to implement a SelectNext that is rendered at the very end of the viewport, we realized the listbox aligns to the left side of the select, making it look odd (see screenshot from #1371).

This PR adds a new `right` boolean prop that makes the listbox align to the right side when it grows bigger than the toggle button.

image

### Context

Initially, I wanted to find a way to make this work transparently, but it requires a lot of custom logic that's not easy to maintain.

Material UI does this, but through the popper.js library, which renders popover-like elements individually, and calculates their position at runtime.

I checked how other solved this problem:

* Headless UI does not deal with this. The listbox is always as big or small as the toggle button, and the extra content is cropped.
* [Reactstrap](https://reactstrap.github.io/?path=/docs/components-dropdown--dropdown) has a similar `end` option. In their case it's more obvious, because menus from dropdowns don't have a minimum width matching the button.

I decided to go with the `right` prop because it's very simple.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -156,6 +156,14 @@ export type SelectProps<T> = CompositeProps & {
   /** Additional classes to pass to listbox */
   listboxClasses?: string | string[];
 
+  /**
+   * Align the listbox to the right.
+   * Useful when the listbox is bigger than the toggle button and this component
+   * is rendered next to the right side of the page/container.
+   * Defaults to false.
+   */
+  right?: boolean;
+
   'aria-label'?: string;
   'aria-labelledby'?: string;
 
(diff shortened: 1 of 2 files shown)


robertknight (approved):
> I decided to go with the right prop because it's very simple.

👍

PR #1405, opened 2023-12-13
Create useToastMessages hook

acelaya:
This PR provides a convenient `useToastMessages` hook that can handle the state for `<ToastMessages />` component.

Currently both via and lms define very similar hooks, so this is a way to extract the logic from them.

Client has more complex needs, and handles toast messages through a shared redux store, but for simple use cases, it comes handy.

### DIFF
--- a/src/hooks/test/use-toast-messages-test.js
+++ b/src/hooks/test/use-toast-messages-test.js
@@ -1,0 +1,105 @@
+import { mount } from 'enzyme';
+
+import { useToastMessages } from '../use-toast-messages';
+
+describe('useToastMessages', () => {
+  function FakeToastMessagesContainer({ initialToastMessages }) {
+    const { toastMessages, appendToastMessage, dismissToastMessage } =
+      useToastMessages(initialToastMessages);
+
+    return (
+      <div>
+        {!toastMessages.length && (
+          <span data-testid="no-toast-messages">
+            There are no toast messages
+          </span>
+        )}
+        {toastMessages.map(({ message }, index) => (
+          <span key={`${message}_${index}`} className="toast-message">
+            {message}
+          </span>
+        ))}
+        <button
+          data-testid="append-button"
+          onClick={() =>
+            appendToastMessage({
+              type: 'success',
+              message: 'Toast message',
+            })
+          }
+        >
+          Append
+        </button>
+        <button
+          data-testid="dismiss-button"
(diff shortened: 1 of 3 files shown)


src/hooks/test/use-toast-messages-test.js
acelaya, line 1: Extracted from via, with the addition of a test that verifies initial toast messages can be provided.

src/hooks/use-toast-messages.ts
acelaya, line 1: Extracted from via, with the capability to provide initial toast messages that lms requires.

src/hooks/use-toast-messages.ts
```diff
@@ -0,0 +1,52 @@
+import { useCallback, useState } from 'preact/hooks';
+
+import type { ToastMessage } from '../components/feedback';
+
+export type ToastMessageData = Omit<ToastMessage, 'id'>;
+
+export type ToastMessageAppender = (toastMessageData: ToastMessageData) => void;
```
robertknight, line 7: What was the rationale for having this type alias, given that the signature is pretty simple?

src/hooks/use-toast-messages.ts
```diff
@@ -0,0 +1,52 @@
...
+
+function dataToToastMessage(toastMessageData: ToastMessageData): ToastMessage {
+  toastMessageCounter++;
+  const id = `${toastMessageCounter}`;
+  return { ...toastMessageData, id };
+}
+
+/**
+ * Hook providing a simple way to handle state for <ToastMessages />
+ * {@link ToastMessages}
```
robertknight, line 26:
```suggestion
 * Hook providing a simple way to handle state for {@link ToastMessages}.
```

robertknight (approved):

src/hooks/use-toast-messages.ts
acelaya: I don't remember 😅 

I don't really see a point for it, to be honest. We can get rid of it.

acelaya: Oh, I remembered. This type was being imported in one of via's components. This makes me realize that:

1. We need to make the types be exposed by the library's entry point.
2. I think we still don't need an alias for this type, as we can type as `ToastMessages['appendToastMessage']`. We have used this pattern in other places, and it gives a clear intention that those two types need to align.

PR #1433, opened 2024-01-17
Add a "minus" icon

robertknight:
Add a "minus" icon which matches our existing "plus" icon, for use in the client's filter controls. See https://github.com/hypothesis/client/pull/6106. See `MinusIcon` in http://localhost:4001/data-icons for a preview.

As noted in the commit message, the icon was created by getting ChatGPT to edit the `plus.svg` markup.

In the process I fixed the icon generation script to work with the latest version of Prettier and regenerated icons. This made some small edits to existing icons due to SVGO updates.


acelaya:
Is it expected that all icons were affected by changes? 🤔. It looks like they all "lost" the `Z` or `z` char somewhere.

robertknight:
> Is it expected that all icons were affected by changes? 🤔. It looks like they all "lost" the Z or z char somewhere.

Yes, since there have been updates to the SVG optimizer (SVGO) that we use since the icons were last generated. From a quick visual skim the icons look the same as before, so it looks like some redundant parts were removed.

acelaya:
> > Is it expected that all icons were affected by changes? 🤔. It looks like they all "lost" the Z or z char somewhere.
> 
> Yes, since there have been updates to the SVG optimizer (SVGO) that we use since the icons were last generated. From a quick visual skim the icons look the same as before, so it looks like some redundant parts were removed.

Gotcha

acelaya (approved):

PR #1449, opened 2024-02-08
Mark inputs as invalid if `feedback="error"`

acelaya:
Part of https://github.com/hypothesis/lms/issues/6012

This PR updates the `Input` component so that it is accessibly marked as invalid when `feedback` prop is provided with `"error"` value.

This requires [HTMLObjectElement.setCustomValidity()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) to be supported by the browser. Since not all browsers we support do it yet, the logic also checks the method is available, and behaves as before otherwise.

### Testing steps

1. Go to http://localhost:4001/input-input
2. Run a screen reader
3. Scroll to the part where the `feedback` prop is documented.
4. Focus the input which has `feedback="error"` -> The screen reader should announce it as "invalid input".

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -1,6 +1,8 @@
 import classnames from 'classnames';
-import type { JSX } from 'preact';
+import type { JSX, RefObject } from 'preact';
+import { useEffect } from 'preact/hooks';
 
+import { useSyncedRef } from '../../hooks/use-synced-ref';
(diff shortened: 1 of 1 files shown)


acelaya:
The same result can be achieved in a more portable way using `aria-invalid="true"`. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid

Closing this.

PR #1452, opened 2024-02-12
Do not render prefix for visually hidden toast messages

acelaya:
Historically, all `success` and `error` toast messages have been rendered with the **Success** or **Error** prefixes. However, this doesn't usually make sense for hidden toast messages, as the message itself is usually composed taking into consideration that it is going to be announced by a screen reader, and therefore it should be as specific as possible.

This PR removes prefixes for any message when it has `visuallyHidden: true`.

### DIFF
--- a/src/components/feedback/ToastMessages.tsx
+++ b/src/components/feedback/ToastMessages.tsx
@@ -52,9 +52,9 @@ type ToastMessageItemProps = {
  */
 function ToastMessageItem({ message, onDismiss }: ToastMessageItemProps) {
   // Capitalize the message type for prepending; Don't prepend a message
-  // type for "notice" messages
+  // type for "notice" or hidden messages
   const prefix =
-    message.type !== 'notice'
+    message.type !== 'notice' && !message.visuallyHidden
       ? `${message.type.charAt(0).toUpperCase() + message.type.slice(1)}: `
       : '';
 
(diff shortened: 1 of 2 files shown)


acelaya:
As an alternative to this PR, we could consider removing prefixes all-together, because for sighted users, success messages are green and error messages are red, which is probably clear enough.

src/components/feedback/ToastMessages.tsx
```diff
@@ -52,9 +52,9 @@ type ToastMessageItemProps = {
  */
 function ToastMessageItem({ message, onDismiss }: ToastMessageItemProps) {
   // Capitalize the message type for prepending; Don't prepend a message
-  // type for "notice" messages
+  // type for "notice" or hidden messages
```
robertknight, line 55: This comment explains what the code does, but not why. The prefix is omitted for notices because that's the default, and for hidden messages because the type is assumed to be obvious from the context.

The main problem with the prefix is just that it takes up time (when read) or space (on screen) that feels unnecessary. This is especially true because all the callouts have an icon to indicate their type.

robertknight (commented):
On further reflection, I think we should just remove the prefix entirely. It is redundant visually, because the callous have an icon and color that depends on their status, and for screen readers, we assume that the message is written in such a way that it makes the status clear (eg. "Text copied", "Annotation saved", "Saving failed").

acelaya:
> On further reflection, I think we should just remove the prefix entirely. It is redundant visually, because the callous have an icon and color that depends on their status, and for screen readers, we assume that the message is written in such a way that it makes the status clear (eg. "Text copied", "Annotation saved", "Saving failed").

Fully agree

PR #1453, opened 2024-02-12
Update to latest GitHub actions version which use node 20

acelaya:
This addresses some warnings that show when running different workflows, due to "old" actions versions using node 16 instead of node 20.

The versions used in this PR are all using node 20.


robertknight (approved):
LGTM. Please feel free to merge the same change in other repos.

Issue #1461, opened 2024-02-13
New `Dialog` component wrapping a native `dialog`


acelaya:
The support for the native `dialog` is reaching all the browsers we support. This is a good opportunity to build a new `Dialog` component that makes use of it, and allows us to drop some of the custom logic.

We could just refactor the existing `Dialog` and `ModalDialog` components, but creating a new one instead, would allow us to apply some architectural changes and not having to keep a 100% compatible API, while we slowly adopt it in downstream projects.

#### Handle its open/closed state via an `open` prop, instead of via mounting/unmounting.

This will better fit the side effects we need to trigger in order to show/hide the native dialog.

It will also help solving an existing problem with current components when transition components are passed. We tried to solve that problem this way some time ago, and it proved challenging, but with a simplified component it will probably be easier to handle.

Ultimately, most of the component libraries we usually get inspiration from, use some kind of prop handled by consumer code to determine if the component should show or not.

#### Do not implicitly wrap a `Panel`.

Existing `Dialog` started as a wrapper around `Panel`, which became a problem later when we tried to use it for other components providing its own "layout".

Instead, we should have a simpler Dialog, and if needed, a higher level `PanelDialog` wrapping the new Dialog and implicitly rendering a `Panel` or `Card`.

### Single `Dialog` component for modal and non-modal contexts.

Due to the amount of custom logic in existing implementations, we have two separate components, `Dialog` and `ModalDialog`.

Thanks to the amount of logic we will be able to get rid of when using the native `dialog`, maybe we can have a single component with a `modal={true|false}` prop.

### Use cases

An inspiration for this component can be found in the client, where we recently used the native `dialog` with a very small fraction of the API we might want to support, to render the notebook and the user profile: https://github.com/hypothesis/client/pull/6207

PR #1464, opened 2024-02-16
New dialog poc

acelaya:
Part of https://github.com/hypothesis/frontend-shared/issues/1461

This PR is a proof of concept of what could be a new component to render native `dialog` elements.

It covers the basic functionality, with an `open` prop that triggers the `dialog.show()` side effect when true, or `dialog.close()` when false.

It also takes into consideration how regular dialogs and modal dialogs should be displayed.

Finally, it does not make assumptions about how children should be rendered, rendering what's provided verbatim.

For cases where we want to render the dialog as a panel/card, we can consider adding a new component that wraps children in a panel and then nests that into the new dialog.

### Open questions

* I reused the `modal` prop for modal sizes, because it was convenient, and the resulting API seems intuitive: `modal` for an auto-size modal dialog, and `modal="lg"` for a large-size modal dialog.
  Should we have a `size` prop that can be used with non-modal dialogs as well? Should we have a separate `ModalDialog` that handles sizes?
* Should we have a fallback for browsers not supporting `dialog` elements? Perhaps we could define a component wrapping the new dialog and the old one and orchestrating which one to render based on browser support.
* Should we drop support for the `closeOnFocusAway` prop? It is false by default and is not being redefined in any downstream project.
  Also, you cannot focus away when opening the native dialog as modal, and when it is not a modal you will likely want to keep it open.
* Should we drop support for `closeOnEscape`? It is true by default for modal dialogs and false for regular dialogs.
  Native dialogs implement this natively when opened as modal, which matches that behavior.
  There's only one use case in which we are currently using a non-modal dialog which can be closed with `Esc`, which is the grade comment popover. In this case we are using the dialog because it's convenient, but it is actually implementing the popover pattern, which is different.
* Should we drop support for `closeOnClickAway`? It's false by default and like in previous point, we are redefining this behavior only for the grade comment popover.

### TODO

* [ ] Decide if we want to support `closeOnClickAway`, and implement it if so
* [ ] Decide if we want to support `closeOnFocusAway`, and implement it if so
* [ ] Decide if we want to support `closeOnEscape`, and implement it if so
* [ ] Implement `initialFocus`
* [ ] Implement `restoreFocus`
* [ ] Implement `transitionComponent`
* [ ] Test the component with real use cases in downstream projects

### DIFF
--- a/src/components/feedback/DialogNext.tsx
+++ b/src/components/feedback/DialogNext.tsx
@@ -7,11 +7,26 @@ import type { PresentationalProps } from '../../types';
 import type { CloseableInfo } from '../CloseableContext';
 import CloseableContext from '../CloseableContext';
 
+export type ModalSize = 'sm' | 'md' | 'lg';
+
 type ComponentProps = {
   open?: boolean;
   onClose?: () => void;
   closeTitle?: string;
-  modal?: boolean;
+
+  /**
+   * Determines if this dialog should be opened as a modal, ensuring proper
+   * focus trap.
+   * The modal size can be optionally provided too, otherwise it will adjust its
+   * contents.
+   * Defaults to `false`.
+   *
+   * <Dialog modal /> -> A modal dialog with automatic sizing.
+   * <Dialog modal="sm" /> -> A small-size modal dialog.
+   * <Dialog modal="md" /> -> A medium-size modal dialog.
+   * <Dialog modal="lg" /> -> A large-size modal dialog.
+   */
+  modal?: boolean | ModalSize;
 };
 
 export type DialogNextProps = PresentationalProps &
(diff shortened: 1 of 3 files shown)


acelaya:
I have been trying to adopt the `TransitionComponent` API as we did with previous dialog components, and there are two considerations:

1. `TransitionComponents` were build in a generic way, which is very flexible, but adopting them dynamically adds a lot of complexity.
2. The differences between how this new dialog works internally and the previous ones mean we have to spend time solving the same problem again. It was not easy then, and I haven't nailed it yet. In fact, this one presents new problems that might be trickier to address.
3. It still has the same problems we faced with transition components in previous dialogs.

Because of this, I'm thinking on a less flexible but simpler solution, which implies adding CSS transitions directly to the dialog, as explained in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

This would add less flexibility, as every type of transition would need to be implemented via tailwind classes and config.

However, since `TransitionComponents` were created we have used only one implementation, the `Slider`, so we could start with that.

PR #1471, opened 2024-02-21
Update codecov action to fix node.js version warning

acelaya:


robertknight (approved):

acelaya:
> We need to set a token when uploading reports now. See [hypothesis/client#6225 (review)](https://github.com/hypothesis/client/pull/6225#pullrequestreview-1895443139).

I have just set the secret in the repository, and pushed a new change to make use of it.

robertknight (approved):
LGTM assuming the upload succeeds.

PR #1482, opened 2024-03-04
Allow ordering DataTables by clicking a header

acelaya:
Add new `order` and `onOrderChange` optional props to `DataTable`, that can be used to handle which column is being used to order the rows and display a visual hint.

The type of `order` is `null | { field: Field; direction: 'ascending' | 'descending' }`, where `Field` represents any key of the generic `Row` type from the table.

The value `null` is also the default, and represents that no ordering is being applied.

Every time a column is clicked, if `onOrderChange` was provided, it tries to determine what's the new ordering following this logic.

* When the same column is clicked it transitions from `null`->`{ field: 'the_column', direction: 'ascending' }`->`{ field: 'the_column', direction: 'descending' }`->`null`->...
* When a different column is clicked, it resets whatever ordering is active into `{ field: 'the_new_column', direction: 'ascending' }`

Once determined, `onOrderChange` is invoked with the new value.

### Examples

This PR includes examples in the docs where ordering the rows happens client-side and server-side. This is possible thanks to the fact that both props are controlled by the consumer component.

https://github.com/hypothesis/frontend-shared/assets/2719332/58e955aa-4461-4f8b-bfb4-cea1aaf7b46e

https://github.com/hypothesis/frontend-shared/assets/2719332/2eddace6-517f-4447-b236-192390e58b9a

### A11y

Accessibility is implemented as described in https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/

* The active sorting column includes `aria-sort="ascending|descending"`.
* Ordering icons are hidden from screen readers via `aria-hidden`.
* Columns that can be used to order the table, render a header button with special focus style.
* Orderable columns include an `aria-label` and `title` which explains that clicking on them will order the table.

### TODO

 - [x]  Make the solution accessible
 - [x] Add an example where data is loaded asynchronously
 - [x] Consider allowing to define the columns that can be used to order
 - [x] Add tests

### DIFF
--- a/src/components/data/DataTable.tsx
+++ b/src/components/data/DataTable.tsx
@@ -6,7 +6,8 @@ import { useStableCallback } from '../../hooks/use-stable-callback';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
 import type { CompositeProps } from '../../types';
 import { downcastRef } from '../../util/typing';
-import { SpinnerSpokesIcon } from '../icons';
+import { ArrowDownIcon, ArrowUpIcon, SpinnerSpokesIcon } from '../icons';
+import { Button } from '../input';
 import ScrollContext from './ScrollContext';
 import Table from './Table';
 import TableBody from './TableBody';
@@ -15,15 +16,20 @@ import TableFoot from './TableFoot';
 import TableHead from './TableHead';
 import TableRow from './TableRow';
 
-export type TableColumn = {
-  field: string;
+export type TableColumn<Field> = {
+  field: Field;
   label: string;
   classes?: string;
 };
 
+export type Order<Field> = {
+  field: Field;
+  direction: 'ascending' | 'descending';
+} | null;
+
 type ComponentProps<Row> = {
   rows: Row[];
-  columns: TableColumn[];
+  columns: TableColumn<keyof Row>[];
 
   /** Content to render if rows is empty (and not in a loading state) */
   emptyMessage?: ComponentChildren;
@@ -59,6 +65,25 @@ type ComponentProps<Row> = {
    */
   onConfirmRow?: (r: Row) => void;
 
+  /** Current sort order */
+  order?: Order<keyof Row>;
+
+  /**
+   * Callback invoked when user clicks a column header to change the sort order.
+   * When a header is clicked, if that's not the active order, it is set with
+   * order='ascending'.
+   * If the active header is clicked consecutively, direction rotates from
+   * 'ascending' to 'descending', and from 'descending' to no-active-order.
+   */
+  onOrderChange?: (order: Order<keyof Row>) => void;
+
+  /**
+   * Columns that can be used to order the table. Ignored if `onOrderChange` is
+   * not provided.
+   * No columns will be orderable if this is not provided.
+   */
+  orderableColumns?: Array<keyof Row>;
+
   /** Callback to render an individual table cell */
   renderItem?: (r: Row, field: keyof Row) => ComponentChildren;
   title: string;
@@ -73,6 +98,38 @@ export type DataTableProps<Row> = CompositeProps &
 
 function defaultRenderItem<Row>(r: Row, field: keyof Row): ComponentChildren {
   return r[field] as ComponentChildren;
+}
+
+function calculateNewOrder<T>(newField: T, prevOrder: Order<T>): Order<T> {
+  if (newField !== prevOrder?.field) {
+    return { field: newField, direction: 'ascending' };
+  }
+
+  return prevOrder.direction === 'ascending'
+    ? { field: newField, direction: 'descending' }
+    : null;
+}
+
+type HeaderComponentProps = {
+  onClick?: () => void;
+  children: ComponentChildren;
+  field: string;
+};
+
+function HeaderComponent({ children, onClick, field }: HeaderComponentProps) {
+  const commonClasses = 'flex justify-between items-center';
+  return onClick ? (
+    <Button
+      classes={`${commonClasses} w-full !p-3`}
+      variant="custom"
+      onClick={onClick}
+      data-testid={`${field}-order-button`}
+    >
+      {children}
+    </Button>
+  ) : (
+    <div className={commonClasses}>{children}</div>
+  );
 }
 
 /**
@@ -94,6 +151,10 @@ export default function DataTable<Row>({
   onConfirmRow,
   emptyMessage,
 
+  order = null,
+  onOrderChange,
+  orderableColumns = [],
+
   // Forwarded to Table
   borderless,
 
@@ -101,6 +162,13 @@ export default function DataTable<Row>({
 }: DataTableProps<Row>) {
   const tableRef = useSyncedRef(elementRef);
   const scrollContext = useContext(ScrollContext);
+  const updateOrder = useCallback(
+    (newField: keyof Row) => {
+      const newOrder = calculateNewOrder(newField, order);
+      onOrderChange?.(newOrder);
+    },
+    [onOrderChange, order],
+  );
 
   const noContent = loading || (!rows.length && emptyMessage);
   const fields = useMemo(() => columns.map(column => column.field), [columns]);
@@ -251,11 +319,38 @@ export default function DataTable<Row>({
     >
       <TableHead>
         <TableRow>
-          {columns.map(column => (
-            <TableCell key={column.field} classes={column.classes}>
-              {column.label}
-            </TableCell>
-          ))}
+          {columns.map(column => {
+            const isOrderable =
+              !!onOrderChange && orderableColumns.includes(column.field);
+            const isActiveOrder = order?.field === column.field;
+
+            return (
+              <TableCell
+                key={column.field}
+                classes={column.classes}
+                unpadded={isOrderable}
+                aria-sort={isActiveOrder ? order.direction : undefined}
+              >
+                <HeaderComponent
+                  field={column.field.toString()}
+                  onClick={
+                    isOrderable ? () => updateOrder(column.field) : undefined
+                  }
+                >
+                  <div>{column.label}</div>
+                  {isActiveOrder && (
+                    <div className="bg-white rounded p-1" aria-hidden>
+                      {order.direction === 'ascending' ? (
+                        <ArrowUpIcon />
+                      ) : (
+                        <ArrowDownIcon />
+                      )}
+                    </div>
+                  )}
+                </HeaderComponent>
+              </TableCell>
+            );
+          })}
         </TableRow>
       </TableHead>
       <TableBody>
--- a/src/components/data/TableCell.tsx
+++ b/src/components/data/TableCell.tsx
@@ -8,7 +8,10 @@ import TableContext from './TableContext';
 import TableSectionContext from './TableSectionContext';
 
 export type TableCellProps = PresentationalProps &
-  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'>;
+  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'> & {
+    /** Remove default padding, allowing consuming code to control it */
+    unpadded?: boolean;
+  };
 
 /**
  * Render a single table cell
@@ -17,6 +20,7 @@ export default function TableCell({
   children,
   classes,
   elementRef,
+  unpadded = false,
 
   ...htmlAttributes
 }: TableCellProps) {
@@ -31,8 +35,8 @@ export default function TableCell({
       {...htmlAttributes}
       ref={downcastRef(elementRef)}
       className={classnames(
-        'p-3',
         {
+          'p-3': !unpadded,
           // Set horizontal borders here for table headers. This needs to be
           // done here (versus on the row or table) to prevent a 1-pixel wiggle
           // on scroll with sticky headers.
--- a/src/components/data/test/DataTable-test.js
+++ b/src/components/data/test/DataTable-test.js
@@ -406,5 +406,83 @@ describe('DataTable', () => {
     });
   });
 
+  context('when table can be ordered', () => {
+    [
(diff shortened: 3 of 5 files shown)


src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -1,17 +1,87 @@
...
+  rows,
+  ...rest
+}: Omit<DataTableProps<StrictNabokovNovel>, 'order' | 'onOrderChange'>) {
+  const [loading, setLoading] = useState(false);
+  const [order, setOrder] = useState<Order<keyof StrictNabokovNovel>>(null);
+  const changeOrder = useCallback(
+    (newOrder: Order<keyof StrictNabokovNovel>) => {
+      setOrder(newOrder);
+      setLoading(true);
+      setTimeout(() => setLoading(false), 600);
```
acelaya, line 68: The async loading of ordered rows is faked here.

In a real scenario, we would be using a fetcher, where the order would be part of the key somehow, and would also receive the order object. The we would pass the list of rows returned by the fetcher asynchronously, and use its `loading` to provide it to the DataTable.

src/components/data/DataTable.tsx
```diff
@@ -75,6 +97,38 @@ function defaultRenderItem<Row>(r: Row, field: keyof Row): ComponentChildren {
...
+  onClick?: () => void;
+  children: ComponentChildren;
+  label: string;
+};
+
+function HeaderComponent({ children, onClick, label }: HeaderComponentProps) {
+  const commonClasses = 'flex justify-between items-center';
+  return onClick ? (
+    <Button
+      classes={`${commonClasses} w-full !p-3`}
```
acelaya, line 120: Setting `!p-3`is a bit ugly. I considered adding `unstyled`, so that then we can simply add the regular `p-3` class, but it has the side effect that the focus ring styles are not added.

src/components/data/DataTable.tsx
```diff
@@ -15,15 +16,20 @@ import TableFoot from './TableFoot';
...
 
+export type Order<Field> = {
+  field: Field;
+  direction: 'ascending' | 'descending';
+} | null;
+
 type ComponentProps<Row> = {
   rows: Row[];
-  columns: TableColumn[];
+  columns: TableColumn<keyof Row>[];
```
acelaya, line 32: Making this generic requires being more strict in downstream projects, but it's not critical. I tested this in LMS, and the only requirement was typing the array of columns as `DataTableProps<foo>['columns']`.

In the docs I used a different approach, which is setting the values `as const` on the fly. Both options should work and we might choose one over the other depending on the case.

Other than that, it is very useful to avoid typos, and plays well with the new `orderableColumns` prop.

src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -1,17 +1,87 @@
...
+      }
+
+      if (order.direction === 'ascending') {
+        return a[order.field] > b[order.field] ? 1 : -1;
+      }
+
+      return a[order.field] > b[order.field] ? -1 : 1;
+    });
+  }, [order, rows]);
+}
```
acelaya, line 39: I reckon we might want to eventually extract this hook and expose it as a utility.

src/components/data/DataTable.tsx
```diff
@@ -59,6 +65,25 @@ type ComponentProps<Row> = {
...
+   * 'ascending' to 'descending', and from 'descending' to no-active-order.
+   */
+  onOrderChange?: (order: Order<keyof Row>) => void;
+
+  /**
+   * Columns that can be used to order the table. Ignored if `onOrderChange` is
+   * not provided.
+   * Defaults to all columns.
+   */
+  orderableColumns?: Array<keyof Row>;
```
acelaya, line 85: We don't usually use the `Array<>` notation to type arrays, but I think `Array<keyof Row>` is easier to read than `(keyof Row)[]`

src/components/data/DataTable.tsx
robertknight: This assumes that the `Row` objects are always going to be objects with properties that match the columns. I expect that is usually true. It might turn out to be too strict a requirement in some cases. An alternative would be to define a second generic for `ComponentProps` which is the type of columns, defaulting to `keyof Row`.

src/components/data/DataTable.tsx
```diff
@@ -59,6 +65,25 @@ type ComponentProps<Row> = {
    */
   onConfirmRow?: (r: Row) => void;
 
+  /** Currently active order */
```
robertknight, line 68:
```suggestion
  /** Current sort order */
```

src/components/data/DataTable.tsx
```diff
@@ -59,6 +65,25 @@ type ComponentProps<Row> = {
...
+  order?: Order<keyof Row>;
+
+  /**
+   * Callback when active order changes, by clicking a column header.
```
robertknight, line 72:
```suggestion
   * Callback invoked when user clicks a column header to change the sort order.
```

There might be other triggers that change the sort order in future, but I think it makes sense to describe this in terms of the main trigger.

src/components/data/DataTable.tsx
robertknight: I agree. I usually use `Array` too in such cases.

src/components/data/DataTable.tsx
```diff
@@ -75,6 +100,38 @@ function defaultRenderItem<Row>(r: Row, field: keyof Row): ComponentChildren {
...
+};
+
+function HeaderComponent({ children, onClick, label }: HeaderComponentProps) {
+  const commonClasses = 'flex justify-between items-center';
+  return onClick ? (
+    <Button
+      classes={`${commonClasses} w-full !p-3`}
+      variant="custom"
+      onClick={onClick}
+      title={`Click to order table by ${label}`}
```
robertknight, line 126: I think we can omit this title here, as it is a strong convention that tables with button headers will sort their columns. Including the "Click to order table by" text makes navigating the table a bit annoying because it adds uninformative text in front of the column name. The example you mentioned at https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/ for example just uses regular buttons. There is also a single table-wide caption to provide a hint, but I don't think that's essential.

src/components/data/DataTable.tsx
robertknight, line 85: The caller will need to provide code for every column they want to support sorting by. For that reason it might be better to require the caller to list all of the sortable columns instead of defaulting to assuming they are all sortable.

src/components/data/TableCell.tsx
```diff
@@ -8,7 +8,10 @@ import TableContext from './TableContext';
 import TableSectionContext from './TableSectionContext';
 
 export type TableCellProps = PresentationalProps &
-  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'>;
+  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'> & {
+    /** Do not apply a default padding, allowing consuming code to control it */
+    unPadded?: boolean;
```
robertknight, line 13:
```suggestion
    unpadded?: boolean;
```

src/components/data/test/DataTable-test.js
```diff
@@ -406,5 +406,81 @@ describe('DataTable', () => {
...
+      // Clicking the same column when initially ascending, transitions to descending
+      {
+        initialOrder: { field: 'name', direction: 'ascending' },
+        clickedColumn: 'Name',
+        expectedNewOrder: {
+          field: 'name',
+          direction: 'descending',
+        },
+      },
+      // Clicking the same column when initially descending, transitions to no order
```
robertknight, line 435: Was the choice to cycle between `ascending -> descending -> no order` rather than `ascending -> descending -> ascending ...` based on an existing UI? I wonder if it might be better to simplify the behavior so that there is no option to remove the sorting once applied, ie. it would just cycle through ascending/descending.

src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -1,17 +1,87 @@
...
 const nabokovColumns = [
-  { field: 'title', label: 'Title' },
-  { field: 'year', label: 'Year' },
-  { field: 'language', label: 'Language' },
+  { field: 'title' as const, label: 'Title' },
+  { field: 'year' as const, label: 'Year' },
+  { field: 'language' as const, label: 'Language' },
 ];
 
+type StrictNabokovNovel = Omit<NabokovNovel, 'translatedTitle'>;
```
robertknight, line 16: Why `Strict`?

src/pattern-library/components/patterns/data/DataTablePage.tsx
robertknight: If you used the same approach as the fake implementation here but used varying timeouts then you might get the wrong result if clicking to change the order while the table is loading. If the order becomes part of the key in a proper fetch mechanism then that problem should be avoided.

robertknight (commented):
Overall this looks great. I had some small comments on the API and header labeling, but I think this is pretty close.

src/components/data/DataTable.tsx
acelaya: > An alternative would be to define a second generic for ComponentProps which is the type of columns, defaulting to keyof Row.

I like this idea.

acelaya: I tried to apply a less strict logic, but TypeScript ends up complaining one way or another because the logic assumes a correlation between columns and the fields on every row for different tasks. In fact, I realized we are already using `keyof Row` at least in two places. See [line 63](https://github.com/hypothesis/frontend-shared/blob/8c9a30c6dbf03e883a33b94212a199d65744d22f/src/components/data/DataTable.tsx#L63) and [line 74](https://github.com/hypothesis/frontend-shared/blob/8c9a30c6dbf03e883a33b94212a199d65744d22f/src/components/data/DataTable.tsx#L74).

On top of that, when it comes to logic related specifically with ordering, a new generic type that defaults to `keyof Row` is still not strict enough. Typescript starts spotting errors unless this new generic type is also defined as `extends keyof Row`, which means you can still define only keys of Row, defeating the purpose.

So if you are ok with it, I think I'm going to leave it as is for now. If we find a case in which columns do not strictly match Row fields, there are pieces of logic that would't work anyway, so we can come back to it when we have a concrete case.

robertknight: OK, that makes sense. Let's continue with this as-is.

src/components/data/DataTable.tsx
acelaya: I had mixed feelings about this title, but I also didn't love the caption approach suggested in https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/, so if you think we can just skip this, I'm happy with it 🙂

src/components/data/DataTable.tsx
acelaya: Yep, I think you are right.

src/components/data/DataTable.tsx
robertknight: I think this is a common-enough pattern that we can skip the caption.

src/pattern-library/components/patterns/data/DataTablePage.tsx
acelaya: Nah, because I couldn't come up with a better wording 😅. I just wanted to exclude `translatedTitle` because it's optional, hence the word "Strict".

I guess we could name it `BasicNabokovNovel`, `BaseNabokovNovel`, `SimpleNabokovNovel`, or something in those lines.

src/pattern-library/components/patterns/data/DataTablePage.tsx
acelaya: Good catch. I'll try to handle that here.

I guess in the case of using `useFetch`, the in-progress request would be aborted and a new one would be triggered, so we should be covered as you said.

acelaya:
@robertknight you can find the reasoning behind the order direction loop (ascending -> descending -> no order) in this comment https://github.com/hypothesis/frontend-shared/pull/1482/files#r1519783744

One example of what's explained there can actually be seen in [the pattern library](http://localhost:4001/data-datatable). The list of novels has an initial arbitrary order, which does not match ordering by any of the columns, ascending nor descending.

src/components/data/TableCell.tsx
```diff
@@ -8,7 +8,10 @@ import TableContext from './TableContext';
 import TableSectionContext from './TableSectionContext';
 
 export type TableCellProps = PresentationalProps &
-  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'>;
+  Omit<JSX.HTMLAttributes<HTMLElement>, 'size'> & {
+    /** Do not apply a default padding, allowing consuming code to control it */
```
robertknight, line 12:
```suggestion
    /** Remove default padding, allowing consuming code to control it */
```

I think this makes it a bit easier to reason about what `true` and `false` values mean ("Do/don't remove default padding" vs "Do/don't not apply a default padding").

src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -1,17 +1,104 @@
...
 const nabokovColumns = [
-  { field: 'title', label: 'Title' },
-  { field: 'year', label: 'Year' },
-  { field: 'language', label: 'Language' },
+  { field: 'title' as const, label: 'Title' },
+  { field: 'year' as const, label: 'Year' },
+  { field: 'language' as const, label: 'Language' },
 ];
 
+type SimpleNabokovNovel = Omit<NabokovNovel, 'translatedTitle'>;
```
robertknight, line 16: I think this name is better. "Strict" sounded like it was something to do with TypeScript strict-ness.

robertknight:
Don't forget to update the PR title, as it is no longer a "POC"!

PR #1494, opened 2024-03-21
Add download icon

acelaya:
Part of https://github.com/hypothesis/client/issues/6255

Add new `DownloadIcon` matching the one used in the new pending updates notification designs.

image

The new icon component was added by running `node ./scripts/generate-icons.js`, which as a side effect updated/optimized some of the existing icons, probably due to changes on the `svgo` dependency.


robertknight (approved):
LGTM. I believe you're right about the changes to other files.

PR #1510, opened 2024-04-02
Adjusting overall size of padding within `Panel`, `Card` and `Dialog` instances

jaredpdesigns:
To provide more overall padding within the `Dialog` component, I adjusted various sub-components used and tied into the `paddingSize` prop used within various `Panel` and `Card` components. I specifically mapped the sizing to `lg` and changed the default `paddingSize` prop on `Dialog` to account for this.

### DIFF
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -100,7 +100,7 @@ export default function Dialog({
   const {
     buttons,
     icon,
-    paddingSize = 'md',
+    paddingSize = 'lg',
     title = '',
     scrollable = true,
     ...htmlAttributes
--- a/src/components/feedback/Dialog.tsx
+++ b/src/components/feedback/Dialog.tsx
@@ -15,12 +15,17 @@ import { useClickAway } from '../../hooks/use-click-away';
 import { useFocusAway } from '../../hooks/use-focus-away';
 import { useKeyPress } from '../../hooks/use-key-press';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
-import type { PresentationalProps, TransitionComponent } from '../../types';
+import type {
+  PresentationalProps,
+  Size,
+  TransitionComponent,
+} from '../../types';
 import { downcastRef } from '../../util/typing';
 import CloseableContext from '../CloseableContext';
 import type { CloseableInfo } from '../CloseableContext';
 import Panel from '../layout/Panel';
 import type { PanelProps } from '../layout/Panel';
+import type { ModalSize } from './ModalDialog';
 
 type ComponentProps = {
   closeOnClickAway?: boolean;
@@ -100,7 +105,7 @@ export default function Dialog({
   const {
     buttons,
     icon,
-    paddingSize = 'lg',
+    paddingSize = 'md',
     title = '',
     scrollable = true,
     ...htmlAttributes
@@ -234,6 +239,14 @@ export default function Dialog({
     title: closeTitle,
   };
 
+  const modalSize: ModalSize = (htmlAttributes as any)[
+    'data-modal-size'
+  ] as ModalSize;
+  const modalSizeScoped: undefined | Size =
+    modalSize === 'sm' || modalSize === 'md' || modalSize === 'lg'
+      ? modalSize
+      : undefined;
+
   return (
     <CloseableContext.Provider value={closeableContext}>
       <Wrapper
@@ -258,7 +271,7 @@ export default function Dialog({
               buttons={buttons}
               fullWidthHeader={true}
               icon={icon}
-              paddingSize={paddingSize}
+              paddingSize={modalSizeScoped || paddingSize}
               title={title}
               scrollable={scrollable}
             >
--- a/src/components/layout/CardHeader.tsx
+++ b/src/components/layout/CardHeader.tsx
@@ -10,7 +10,7 @@ import CardTitle from './CardTitle';
 
 type ComponentProps = {
   title?: string;
-  size?: 'sm' | 'md' | 'lg';
+
   /**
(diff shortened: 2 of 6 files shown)


src/components/layout/CardHeader.tsx
```diff
@@ -10,7 +10,7 @@ import CardTitle from './CardTitle';
 
 type ComponentProps = {
   title?: string;
-
+  size?: 'sm' | 'md' | 'lg';
```
acelaya, line 13: Not sure if I'm missing something, but I don't see the `sm` and `md` values being used differently?

src/components/layout/Panel.tsx
```diff
@@ -93,7 +93,11 @@ export default function Panel({
       classes={heightConstraintClasses}
       elementRef={downcastRef(elementRef)}
     >
-      <CardHeader onClose={onClose} fullWidth={scrollable || fullWidthHeader}>
+      <CardHeader
+        onClose={onClose}
+        fullWidth={scrollable || fullWidthHeader}
+        size={paddingSize !== 'none' ? paddingSize : undefined}
```
acelaya, line 99: Inferring inner component props based on the main component props, which have an initial different purpose, could become confusing and produce unexpected results, as one could wonder why changing `paddingSize` affects other UI aspects.

If we want to be able to change the `Card`/`CardHeader` size when creating a `Panel`, perhaps it makes more sense to explicitly expose a `cardSize` prop.

src/components/layout/Panel.tsx
```diff
@@ -111,7 +115,11 @@ export default function Panel({
         <>{panelContent}</>
       )}
       {buttons && (
-        <CardContent data-testid="panel-buttons">
+        <CardContent
+          data-testid="panel-buttons"
+          size={paddingSize !== 'none' ? paddingSize : undefined}
+          style={{ paddingBlockStart: 'unset' }}
```
acelaya, line 121: Why is this needed?

src/components/layout/Panel.tsx
acelaya: On top of that, if both components will need a matching size, perhaps it makes more sense to set it in the `Card` and make it expose it to children components via context.

src/components/layout/CardHeader.tsx
acelaya: By looking at how this is used, perhaps it makes more sense to pass `classes="px-6"` explicitly from some parent component when some other prop is provided there, instead of defining a new prop here.

src/components/feedback/Dialog.tsx
```diff
@@ -100,7 +100,7 @@ export default function Dialog({
   const {
     buttons,
     icon,
-    paddingSize = 'md',
+    paddingSize = 'lg',
```
acelaya, line 103: Let's make sure this does not become a breaking change. For that we need to check every place where `Dialog` or `ModalDialog` is used in client, lms and via, and make sure the new style is what we want.

Otherwise, let's keep the existing default value, and change it in those places where we actually want to use a large padding.

acelaya (changes requested):
I had some questions about some of the changes introduced in this PR.

Also, it would be great if you could add some screenshots of before/after.

src/components/layout/Panel.tsx
jaredpdesigns: @acelaya I like your solution here of passing in a class when some property from the parent is `true`, implemented that here instead.

jaredpdesigns:
@acelaya thanks for your thorough review here. I moved things around and tried inferring type as often as possible. The biggest issue with this scope of work is that there are multiple references to sizes which don't really have a relationship to one another. In one context, we are specifying sizes in relationship to a modal which includes `custom` and `none` in addition to the ones we care about `sm, md, lg`, so my solution was to have a global `Size` type and when necessary scope out only the `sm, md, lg` sizes when I need them instead of worrying about the custom values which don't affect children.

This ended up being a large amount of work then I expected! 😅

acelaya (commented):
I think this PR is getting a bit too complex. I'll try to come back to it as soon as I can, but I need to focus on other things at the moment.

acelaya:
> I think this PR is getting a bit too complex

I have been thinking in some ways to achieve this with a bit less complexity. Some context first.

The root problem is the fact that there's already a bit too much component nesting here.

When dealing with `Dialog` or `ModalDialog`, there's in fact a `Panel`, containing a `Card`, containing a `CardContent`, and finding the balance between abstracting that component tree, which is an implementation detail of `Dialog` and `ModalDialog`, while allowing to customize how they end up looking, is tricky.

Also, any change there needs to be done carefully, because it's very easy to introduce a breaking change. We could just decide to change the classes applied for different sizes, but that would require checking every place where they are used to make sure they look good there. Same if we decide to change the default values for props.

Additionally, it's important that we stick with just one way to style components, via Tailwind classes in this case. If we start mixing that with `style={}`, it's going to quickly become hard to maintain, as style specificity will become even less predictable and we'll have to remember to look at two different places for styles.

Because of that, I would suggest not merging this PR, and instead try to style things directly in LMS, with the available tools. Some possible ideas:

* Add extra padding to the lowermost element in the modal, which we control there, making it add-up to the Card padding. We are already adding extra top padding there, so we could also add right/left padding as well.
* Switch to `lg` size. The default is `md`, which has less padding.

In both cases, the main problem is that this would cause a miss-alignment between the `CardContent` and the `CardHeader` paddings. This is probably something that should be fixed here, perhaps by making `paddingSize` affect `CardHeader` as well, or by allowing `title` to be of type `ComponentChildren` instead of `string`, so that nested components with extra added padding can be passed if needed.

Last one is more hacky, but less likely to introduce BC breaks.

PR #1517, opened 2024-04-11
Set z-index > 0 for `DataTable` header

robertknight:
This ensures that sticky headers are drawn above any positioned content in table cells. See https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_without_z-index.

This fixes an issue in https://github.com/hypothesis/lms/pull/6187 where introducing the use of `position: relative` for thumbnail elements caused them to be rendered above the sticky header when scrolled.

### DIFF
--- a/src/components/data/TableHead.tsx
+++ b/src/components/data/TableHead.tsx
@@ -37,6 +37,9 @@ export default function TableHead({
         {...htmlAttributes}
         ref={downcastRef(elementRef)}
         className={classnames(
+          // This ensures the header is drawn on top of positioned content
+          // in table cells.
+          'z-1',
           'bg-grey-2',
           {
             'sticky top-0': tableContext?.stickyHeader,


src/components/data/TableHead.tsx
```diff
@@ -37,6 +37,9 @@ export default function TableHead({
         {...htmlAttributes}
         ref={downcastRef(elementRef)}
         className={classnames(
+          // This ensures the header is drawn on top of positioned content
+          // in table cells.
+          'z-10',
```
acelaya, line 42: Do we need it to be that high? Maybe `z-1` is enough for the purpose of this PR while introducing less potential side effects?

robertknight: `z-10` is the lowest "standard value" according to https://tailwindcss.com/docs/z-index, although `z-[1]` should in principle work.

acelaya: Oh! Yeah, I remember seeing my IDE auto-complete `x-1`, that's why I suggested it. I wonder why is it not documented 😅

acelaya (approved):

PR #1523, opened 2024-04-17
Export useOrderedRows hook

acelaya:
Export the `useOrderedRows` hook that was created for the pattern library, but can now be useful in other places, [like LMS](https://github.com/hypothesis/lms/pull/6174).

### DIFF
--- a/src/hooks/test/use-ordered-rows-test.js
+++ b/src/hooks/test/use-ordered-rows-test.js
@@ -1,0 +1,129 @@
+import { mount } from 'enzyme';
+import { useState } from 'preact/hooks';
+
+import { useOrderedRows } from '../use-ordered-rows';
+
+const starWarsCharacters = [
+  { name: 'Luke Skywalker', age: 20 },
+  { name: 'Princess Leia Organa', age: 20 },
+  { name: 'Han Solo', age: 25 },
+];
+
+describe('useOrderedRows', () => {
+  function FakeComponent() {
+    const [order, setOrder] = useState();
+    const orderedRows = useOrderedRows(starWarsCharacters, order);
+
+    return (
+      <div>
+        {orderedRows.map((character, index) => (
+          <div key={`${character.name}${index}`}>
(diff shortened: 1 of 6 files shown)


src/hooks/test/use-ordered-rows-test.js
```diff
@@ -0,0 +1,129 @@
...
+  function assertDefaultOrder(wrapper) {
+    assertOrder(wrapper, starWarsCharacters);
+  }
+
+  function assertOrder(wrapper, rows) {
+    rows.forEach((character, index) => {
+      assert.equal(
+        character.name,
+        wrapper.find(`[data-testid="name-${index}"]`).text(),
+      );
```
robertknight, line 72: Assert syntax should be `assert.equal(actual, expected)`. See [the docs](https://www.chaijs.com/api/assert/#method_equal). Therefore the parameters need to be swapped. Naming `rows` to `expectedRows` would make it more immediately obvious that `rows` refers to the expected values.

src/hooks/use-ordered-rows.ts
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'preact/hooks';
+
+import type { Order } from '../types';
+
+export function useOrderedRows<Row>(rows: Row[], order?: Order<keyof Row>) {
```
robertknight, line 5: A return type annotation here would be helpful for the reader. Can you also add a short JSDoc comment explaining usage and behavior.

src/types.ts
```diff
@@ -34,3 +34,8 @@ export type TransitionComponent = FunctionComponent<{
   direction?: 'in' | 'out';
   onTransitionEnd?: (direction: 'in' | 'out') => void;
 }>;
+
+export type Order<Field> = {
+  field: Field;
```
robertknight, line 39: From the implementation I see that `Field` is expected to be a string which can index a row object. In that case `Order<FieldName extends string>` would make that clearer.

robertknight (approved):
LGTM with a few comments on naming and docs.

PR #1524, opened 2024-04-19
Add variant prop to Table

acelaya:
Add a new `variant` prop to `Table` component, that accepts the values `'striped'` and `'grid'`.

* `'striped'`: Current behavior. Renders borderless cells, and uses a slightly darker background color for odd rows. It's the default value.
* `'grid'`: Renders borders for every cell, and does not set a different background color for rows.

Comparison:

image
image

### TODO

- [ ] Add tests
- [ ] Test in combination with interactive tables
- [ ] Test in combination with `borderless`
- [ ] Test in combination with `Scroll`
- [ ] Test in `DataTable`

### To be decided

- Should the header not use grey background when using the grid variant?
- Should the header not use grid separators?

### DIFF
--- a/src/components/data/Table.tsx
+++ b/src/components/data/Table.tsx
@@ -6,17 +6,14 @@ import { useSyncedRef } from '../../hooks/use-synced-ref';
 import type { PresentationalProps } from '../../types';
 import { downcastRef } from '../../util/typing';
 import TableContext from './TableContext';
+import type { CommonTableInfo } from './TableContext';
 import type { TableInfo } from './TableContext';
 
-export type TableProps = PresentationalProps & {
-  stickyHeader?: boolean;
-  /** Sets accessible aria-label */
-  title: string;
-  /** This table has rows that can be selected */
-  interactive?: boolean;
-  /** Turn off outer table borders */
-  borderless?: boolean;
-} & Omit<JSX.HTMLAttributes<HTMLElement>, 'rows'>;
+export type TableProps = PresentationalProps &
+  CommonTableInfo & {
+    /** Sets accessible aria-label */
+    title: string;
+  } & Omit<JSX.HTMLAttributes<HTMLElement>, 'rows'>;
 
 /**
  * Render table content
@@ -30,6 +27,7 @@ export default function Table({
   interactive = false,
   stickyHeader = false,
   borderless = false,
+  variant = 'striped',
 
   ...htmlAttributes
 }: TableProps) {
@@ -40,9 +38,10 @@ export default function Table({
       interactive,
       stickyHeader,
       borderless,
+      variant,
       tableRef: ref,
     }),
-    [borderless, interactive, stickyHeader, ref],
+    [borderless, interactive, stickyHeader, variant, ref],
   );
 
   return (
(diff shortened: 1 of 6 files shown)


robertknight:
Can you add a screenshot of what this looks like with the assignment dashboard?

src/components/data/Table.tsx
```diff
@@ -30,6 +27,7 @@ export default function Table({
   interactive = false,
   stickyHeader = false,
   borderless = false,
+  variation = 'striped',
```
acelaya, line 30: The `striped` term is after how Bootstrap names it https://getbootstrap.com/docs/5.3/content/tables/#striped-rows, but I'm open to other suggestions

acelaya:
> Can you add a screenshot of what this looks like with the assignment dashboard?

Here

image

robertknight:
Thanks. This now also makes it clearer why the columns seemed positioned incorrectly on the right. Outside the scope of this PR, but I think it would look better if the column heading alignment matched the contents. See some of the [React Spectrum TableView](https://react-spectrum.adobe.com/react-spectrum/TableView.html#column-alignment) examples.

Regarding grid lines and striping, I wonder if we should be using two separate boolean properties here to determine whether striping is used and whether visible grid lines are displayed. Our current use cases call for choosing one of two combinations of these, so we could implement that and revise later.

src/pattern-library/components/patterns/data/TablePage.tsx
```diff
@@ -334,6 +334,70 @@ export default function TablePage() {
...
             </Library.Demo>
           </Library.Example>
+          <Library.Example id="table-props-borderless" title="variation">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Overall aspect of the table. When set to{' '}
+                <code>
+                  {"'"}striped{"'"}
+                </code>
+                , the table has not cell borders, and every odd row has a
```
robertknight, line 344:
```suggestion
                , the table has no cell borders, and every odd row has a
```

src/pattern-library/components/patterns/data/TablePage.tsx
```diff
@@ -334,6 +334,70 @@ export default function TablePage() {
               </Table>
             </Library.Demo>
           </Library.Example>
+          <Library.Example id="table-props-borderless" title="variation">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Overall aspect of the table. When set to{' '}
```
robertknight, line 340:
```suggestion
                Visual style of the table. When set to{' '}
```

src/pattern-library/components/patterns/data/TablePage.tsx
```diff
@@ -334,6 +334,70 @@ export default function TablePage() {
...
+                <code>
+                  {"'"}striped{"'"}
+                </code>
+                , the table has not cell borders, and every odd row has a
+                slightly darker background color. When set to{' '}
+                <code>
+                  {"'"}grid{"'"}
+                </code>
+                , cell borders are rendered and all rows have no background.
```
robertknight, line 349:
```suggestion
                , cell borders are rendered and rows have no background.
```

robertknight (commented):
Looks good. A single `variant` option covers our current needs. I'm not sure if we might find we need separate options to control grid lines and row backgrounds in future.

acelaya:
> My feeling is that, if we now introduce variants, and boolean props are added later, it will be more intuitive that the flags overwrite variant styles than the oposite, so perhaps we should start with that?

I have implemented this feature using boolean props just to be sure https://github.com/hypothesis/frontend-shared/pull/1525, and now I'm not so sure we should start with `variant`, as the code for boolean props is actually simpler.

acelaya:
Closing in favor of https://github.com/hypothesis/frontend-shared/pull/1525

PR #1525, opened 2024-04-22
Add grid and unstriped props to table

acelaya:
This is an alternative to https://github.com/hypothesis/frontend-shared/pull/1524, where we manage the UI variants via `grid` and `striped` boolean props that can be combined if needed, instead of a single `variant` prop.

image
image
image

### TODO

- [x] Add tests
- [x] Test in combination with interactive tables
- [x] Test in combination with `borderless`
- [x] Test in combination with `Scroll`
- [x] Test in `DataTable`

### DIFF
--- a/src/components/data/Table.tsx
+++ b/src/components/data/Table.tsx
@@ -9,13 +9,18 @@ import TableContext from './TableContext';
 import type { TableInfo } from './TableContext';
 
 export type TableProps = PresentationalProps & {
+  /** This table has a sticky header */
   stickyHeader?: boolean;
   /** Sets accessible aria-label */
   title: string;
-  /** This table has rows that can be selected */
+  /** This table has click-able, focus-able rows */
   interactive?: boolean;
   /** Turn off outer table borders */
   borderless?: boolean;
+  /** Show a different background every other row */
+  striped?: boolean;
+  /** Show grid lines around table cells */
+  grid?: boolean;
 } & Omit<JSX.HTMLAttributes<HTMLElement>, 'rows'>;
 
 /**
@@ -30,6 +35,8 @@ export default function Table({
   interactive = false,
   stickyHeader = false,
   borderless = false,
+  striped = true,
+  grid = false,
 
   ...htmlAttributes
 }: TableProps) {
@@ -40,9 +47,11 @@ export default function Table({
       interactive,
       stickyHeader,
       borderless,
+      striped,
+      grid,
       tableRef: ref,
     }),
-    [borderless, interactive, stickyHeader, ref],
+    [interactive, stickyHeader, borderless, striped, grid, ref],
   );
 
   return (
--- a/src/components/data/TableCell.tsx
+++ b/src/components/data/TableCell.tsx
@@ -25,7 +25,7 @@ export default function TableCell({
   ...htmlAttributes
 }: TableCellProps) {
   const sectionContext = useContext(TableSectionContext);
-  const { borderless } = useContext(TableContext);
+  const { borderless, grid } = useContext(TableContext);
   const isHeadCell = sectionContext && sectionContext.section === 'head';
   const Cell = isHeadCell ? 'th' : 'td';
 
@@ -42,18 +42,25 @@ export default function TableCell({
           // on scroll with sticky headers.
           'text-left border-b border-b-grey-5': isHeadCell,
           'border-t': isHeadCell && !borderless,
-          'border-none': !isHeadCell,
+          'border-none': !grid && !isHeadCell,
           // Apply a very subtle bottom border to the last row in the table (not
           // in the head). This can help delineate the end of data in tables
           // with sparse row data. Only apply border if row is not selected.
           // This uses Tailwind's nested-group syntax. See
           // https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups
           'group-last/unselected:border-b group-last/unselected:border-grey-2 group-last/unselected:border-dotted':
-            !isHeadCell,
+            !grid && !isHeadCell,
+          // We would typically set `border` in cells together with
+          // `border-collapse` in the parent table. However, that doesn't work
+          // for sticky headers, so we mimic the behavior by conditionally
+          // setting individual cell borders.
+          'group-[:not(:last-child)]:border-b': grid,
         },
         classes,
       )}
       scope={isHeadCell ? 'col' : undefined}
+      data-grid={grid}
+      data-borderless={borderless}
     >
       {children}
     </Cell>
--- a/src/components/data/TableContext.ts
+++ b/src/components/data/TableContext.ts
@@ -8,6 +8,10 @@ export type TableInfo = {
   stickyHeader: boolean;
   /** Turn off outer table borders */
   borderless: boolean;
+  /** Show a different background every other row */
+  striped: boolean;
+  /** Show grid lines around table cells */
+  grid: boolean;
   tableRef: RefObject<HTMLElement | undefined>;
 };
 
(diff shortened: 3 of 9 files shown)


src/components/data/TableContext.ts
```diff
@@ -1,13 +1,20 @@
...
-  interactive: boolean;
+  interactive?: boolean;
   /** This table has a sticky header */
-  stickyHeader: boolean;
+  stickyHeader?: boolean;
   /** Turn off outer table borders */
-  borderless: boolean;
+  borderless?: boolean;
+  /** Do not show a different background every other row */
+  unStriped?: boolean;
```
acelaya, line 12: Probably `striped` with a default value of `true` would be easier to reason about, but this follows our rule that boolean props should have a `false` default value, so that we don't have to overwrite them as `striped={false}`.

That said, we could define a `striped` prop with value false, and then add it everywhere we are using tables, even if it's a breaking change.

robertknight: > Probably striped with a default value of true would be easier to reason about, but this follows our rule that boolean props should have a false default value

The original rationale for it is that you can write something like `<button disabled>`, as in HTML. In that case it is quite easy to think about "the disabled state". This only works if the property name has a good one-word opposite though. Once the opposite needs a prefix like "no" or "un" this gets into the realm of double-negatives, which I find confusing to think about. I would find it more readable to just have `stripes={false}` in this case.

acelaya: Fully agree. I have just pushed the changes to use `striped` with default value `true` instead of `unstriped` with default value `false`.

robertknight (commented):
The `DataTable` docs currently mention the `borderless` property but not the new properties here. Rather than trying to duplicate all of the `Table` props in the `DataTable` docs, it would be worth adding a note that points the user to the `Table` docs for controlling presentational aspects. This could be added in the "Component API" section of that page.

src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -211,6 +211,20 @@ export default function DataTablePage() {
...
+                  <DataTable
+                    title="A subset of Nabokov's novels with publish date and original language"
+                    rows={nabokovRows}
+                    columns={nabokovColumns}
+                    grid
+                  />
+                </Scroll>
+              </div>
+            </Library.Demo>
+          </Library.Example>
```
acelaya, line 227: This example was used to test the usage of a grid table inside a scrollable container.

It highlighted an issue with the use of `border-collapse` with sticky headers (which was in fact documented already inside `Table`), so I had to set borders individually/conditionally in cells to mimic border-collapse, without breaking sticky headers when using the grid.

src/pattern-library/components/patterns/data/TablePage.tsx
```diff
@@ -334,6 +334,145 @@ export default function TablePage() {
...
+                    </TableCell>
+                  </TableRow>
+                </TableBody>
+              </Table>
+            </Library.Demo>
+          </Library.Example>
+          <Library.Example title="grid">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Makes cell borders to be rendered.
```
robertknight, line 391:
```suggestion
                Renders borders for table cells.
```

src/components/data/Table.tsx
```diff
@@ -6,17 +6,14 @@ import { useSyncedRef } from '../../hooks/use-synced-ref';
...
-  stickyHeader?: boolean;
-  /** Sets accessible aria-label */
-  title: string;
-  /** This table has rows that can be selected */
-  interactive?: boolean;
-  /** Turn off outer table borders */
-  borderless?: boolean;
-} & Omit<JSX.HTMLAttributes<HTMLElement>, 'rows'>;
+export type TableProps = PresentationalProps &
+  CommonTableInfo & {
```
robertknight, line 13: `CommonTableInfo` is an internal type that wasn't intended to appear in the public API and isn't exported. I actually think the previous approach of duplicating the props was OK. If you did need to share the props, it would be better to do in the other way around with the internal type borrowing from the public API.

src/components/data/TableCell.tsx
```diff
@@ -42,18 +42,25 @@ export default function TableCell({
...
           // This uses Tailwind's nested-group syntax. See
           // https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups
           'group-last/unselected:border-b group-last/unselected:border-grey-2 group-last/unselected:border-dotted':
-            !isHeadCell,
+            !grid && !isHeadCell,
+          // We would typically set `border` in cells together with
+          // `border-collapse` in the parent table. However, that doesn't work
+          // for sticky headers, so we mimic the behavior by conditionally
+          // setting individual cell borders.
+          'group-[:not(:last-child)]:border-b': grid,
```
robertknight, line 57: Can you clarify why this doesn't work with sticky headers?

src/components/data/TableContext.ts
```diff
@@ -1,13 +1,20 @@
...
-  stickyHeader: boolean;
+  stickyHeader?: boolean;
   /** Turn off outer table borders */
-  borderless: boolean;
+  borderless?: boolean;
+  /** Show a different background every other row */
+  striped?: boolean;
+  /** Show grid lines around table cells */
+  grid?: boolean;
+};
```
robertknight, line 15: As noted elsewhere, this is an internal type that isn't exported, so I wouldn't use it to define `TableProps`. Also it isn't immediately obvious what belongs in `CommonTableInfo` vs `TableInfo`.

src/components/data/TableCell.tsx
```diff
@@ -42,18 +42,25 @@ export default function TableCell({
           // on scroll with sticky headers.
           'text-left border-b border-b-grey-5': isHeadCell,
           'border-t': isHeadCell && !borderless,
-          'border-none': !isHeadCell,
+          'border-none': !grid && !isHeadCell,
```
robertknight, line 45: Hmm... by passing down styling props in JS and consuming them in the child we are emulating something that could be done in CSS directly. That would be more of a problem if the properties changed frequently and we were concerned about performance of updating them, which isn't an issue here.

robertknight (commented):
I had some comments on the `CommonTableInfo` prop, a query about the sticky headers issue, and a small correct to the `grid` docs. Otherwise this looks good.

src/components/data/TableCell.tsx
acelaya: When the `thead` element is set with `position: sticky` and `border-collapse` is not set as `separate` in the `table`, this happens:

https://github.com/hypothesis/frontend-shared/assets/2719332/b8af99f7-26cd-4c96-a9a4-9419d1c93942

As you can see, the header itself is sticky, but the borders of its cells are not, and they "move up" when scrolling.

I'm not sure if this is a bug or there's a reason for browsers to do this (I'll try to investigate a bit), but it's something Lyza already faced and documented here https://github.com/hypothesis/frontend-shared/blob/main/src/components/data/Table.tsx#L60-L63

In my initial iteration of this PR I decided to ignore that because I didn't fully understand the problem, but once I tested the grid table with a scroll container, I saw it.

The solution applied here fulfills with the requirements documented by Lyza, "A side effect is that borders need to be set primarily on table cells, not rows", by setting a bottom border to all cells except those from the last row (via `group-[:not(:last-child)]:border-b`) and right border to all cells except the last one on every row (via `divide-x` in every row).

src/components/data/TableCell.tsx
acelaya: > by passing down styling props in JS and consuming them in the child we are emulating something that could be done in CSS directly

Can you provide an example? I'm not sure if I follow what you mean.

acelaya: Oh, I think I get what you mean. We could be conditionally setting group classes to the row element, and use `group/whatever:border-none` here unconditionally. That way the styles in this element can change based on logic happening in the parent, without re-rendering it.

Is that what you meant?

acelaya: I'll create an issue to look into that separately.

robertknight (approved):

PR #1530, opened 2024-04-24
Allow case-insensitive comparisons in useOrderedRows hook

acelaya:
Add an `options` argument to `useOrderedRows`, which can be used to configure how comparisons are done.

For now this introduces the `caseSensitive` option, which allows consumers to decide if strings should be compared in a case-insensitive or a case-sensitive way.

This option defaults to `false`, changing current behavior, but I consider it was a bug to not do case-insensitive comparisons by default.

Case sensitive comparison:

https://github.com/hypothesis/frontend-shared/assets/2719332/ca872ba9-e095-40ca-854a-504de7512295

Case insensitive comparison (default):

https://github.com/hypothesis/frontend-shared/assets/2719332/ab679132-4b10-484d-8d78-42374b4a54be

### DIFF
--- a/src/hooks/use-ordered-rows.ts
+++ b/src/hooks/use-ordered-rows.ts
@@ -2,29 +2,44 @@ import { useMemo } from 'preact/hooks';
 
 import type { Order } from '../types';
 
+export type UseOrderedRowsOptions = {
+  /**
+   * Make case-sensitive comparisons for string values.
+   * Defaults to false.
+   */
+  caseSensitive?: boolean;
+};
+
 /**
  * Orders a list of rows based on provided order options.
  * Provided rows are not mutated, but a copy is returned instead.
+ * Values are compared preserving their type.
  */
 export function useOrderedRows<Row>(
   rows: Row[],
   order?: Order<keyof Row>,
+  { caseSensitive = false }: UseOrderedRowsOptions = {},
 ): Row[] {
   return useMemo(() => {
     if (!order) {
       return rows;
     }
 
-    return [...rows].sort((a, b) => {
-      if (a[order.field] === b[order.field]) {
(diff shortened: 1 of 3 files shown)


src/hooks/use-ordered-rows.ts
```diff
@@ -2,29 +2,44 @@ import { useMemo } from 'preact/hooks';
...
       return rows;
     }
 
-    return [...rows].sort((a, b) => {
-      if (a[order.field] === b[order.field]) {
+    return [...rows].sort(({ [order.field]: a }, { [order.field]: b }) => {
+      const aField =
+        !caseSensitive && typeof a === 'string' ? a.toLowerCase() : a;
+      const bField =
+        !caseSensitive && typeof b === 'string' ? b.toLowerCase() : b;
```
robertknight, line 32: This handles ASCII but doesn't deal with things like accented characters. With this implementation, "José" would sort after "Joshua" for example. One way to resolve this would be to build on `Intl.Collator`. That has a `sensitivity` option amongst others.

robertknight (commented):
The change here makes sense. While we're resolving this it would be a good time to consider non-English names as well. I gave an example where a name with accents would sort unexpectedly in the current implementation. The [`Intl.Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) API can be used to do more internationally-aware sorting, so we could build on that - allow a subset of `Collator` options to be passed as an argument to the hook, then create a collator internally and use it's `compare` method.

src/hooks/use-ordered-rows.ts
acelaya: Hah, good point. Let me take a look at `Intl.Collator`.

PR #1531, opened 2024-04-24
Fix case-insensitive comparisons in useOrderedRows hook

acelaya:
Refactor `useOrderedRows` to internally order items via `Intl.Collator` when their value is a string, with two main purposes:

1. Fix incorrect order caused by case-sensitive comparisons, causing all uppercase letters to be placed before all lowercase letters, resulting in `ABCabc` instead of `AaBbCc`.
2. Fix ASCI-only comparisons, where `a`, `ä` and `á` were all considered different characters.

### DIFF
--- a/src/hooks/use-ordered-rows.ts
+++ b/src/hooks/use-ordered-rows.ts
@@ -2,9 +2,13 @@ import { useMemo } from 'preact/hooks';
 
 import type { Order } from '../types';
 
+const collator = new Intl.Collator(undefined, { sensitivity: 'case' });
+
 /**
  * Orders a list of rows based on provided order options.
  * Provided rows are not mutated, but a copy is returned instead.
+ * Strings are compared using `Intl.Collator`, other types are compared using
+ * standard JavaScript comparisons.
  */
 export function useOrderedRows<Row>(
   rows: Row[],
@@ -15,16 +19,18 @@ export function useOrderedRows<Row>(
       return rows;
     }
 
-    return [...rows].sort((a, b) => {
-      if (a[order.field] === b[order.field]) {
+    return [...rows].sort(({ [order.field]: a }, { [order.field]: b }) => {
+      const [x, y] = order.direction === 'ascending' ? [a, b] : [b, a];
+
+      if (typeof x === 'string' && typeof y === 'string') {
+        return collator.compare(x, y);
+      }
+
+      if (x === y) {
         return 0;
       }
 
-      if (order.direction === 'ascending') {
-        return a[order.field] > b[order.field] ? 1 : -1;
-      }
-
-      return a[order.field] > b[order.field] ? -1 : 1;
+      return x > y ? 1 : -1;
     });
   }, [order, rows]);
 }
--- a/src/hooks/test/use-ordered-rows-test.js
+++ b/src/hooks/test/use-ordered-rows-test.js
@@ -3,10 +3,14 @@ import { useState } from 'preact/hooks';
 
 import { useOrderedRows } from '../use-ordered-rows';
 
+// Some start with lowercase to test case-insensitive ordering
 const starWarsCharacters = [
   { name: 'Luke Skywalker', age: 20 },
-  { name: 'Princess Leia Organa', age: 20 },
+  { name: 'leia Organa', age: 20 },
   { name: 'Han Solo', age: 25 },
+  { name: 'Baby Yoda', age: 2 },
+  { name: 'baby yöda The Second', age: 2 },
+  { name: 'young Anakin Skywalker', age: 10 },
 ];
 
 describe('useOrderedRows', () => {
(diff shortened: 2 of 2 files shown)


src/hooks/use-ordered-rows.ts
```diff
@@ -15,16 +20,12 @@ export function useOrderedRows<Row>(
...
-    return [...rows].sort((a, b) => {
-      if (a[order.field] === b[order.field]) {
-        return 0;
-      }
-
-      if (order.direction === 'ascending') {
-        return a[order.field] > b[order.field] ? 1 : -1;
-      }
+    return [...rows].sort(({ [order.field]: a }, { [order.field]: b }) => {
+      const collator = typeof a === 'string' ? stringCollator : numberCollator;
```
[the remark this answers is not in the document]
acelaya: I think that's actually not correct 🤔 

image

image

This branch includes test cases where an incorrect ordering of numbers makes tests fail.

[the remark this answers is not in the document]
acelaya: Actually, I have realized tests pass by always using the numeric one 🤔 

I didn't add any test with mixed values ase you said. Let me test that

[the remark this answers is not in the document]
robertknight: See https://tc39.es/ecma402/#sec-collator-compare-functions steps 5 and 6.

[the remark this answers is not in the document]
acelaya: > compare(10, 2) isn't comparing numbers, it coerces the arguments to strings and compares those.

I added that just to showcase it was not a casting thing. In fact it's typed as string in TypeScript, so we have to cast it to string.

Just check this screenshot and ignore the other one:

image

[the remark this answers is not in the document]
acelaya: > My point is that if the data we want to compare are both numbers, we shouldn't use `Intl.Collator` at all

ooooh, gotcha 😅. Yeah, that fully makes sense.

src/hooks/use-ordered-rows.ts
```diff
@@ -2,9 +2,12 @@ import { useMemo } from 'preact/hooks';
 
 import type { Order } from '../types';
 
+const collator = new Intl.Collator(undefined, { sensitivity: 'case' });
```
acelaya, line 5: Due to how we export modules in this library, this object will be created as soon as anything is imported from `@hypothesis/frontend-shared`.

We could implement some lazy-loading mechanism that ensures there's only one collator instance that's not created until the first time it's going to be used, but I didn't do it for simplicity.

```ts
let collator: Intl.Collator;

function getCollator(): Intl.Collator {
  if (!collator) {
    collator = new Intl.Collator(undefined, { sensitivity: 'case' });
  }

return collator;
}

// [...]

return getCollator().compare(x, y);
```

robertknight: I think this should be OK as this instantiation doesn't have global side effects. From my past experience collators are expensive enough that you don't want to create them eg. a sorting comparison function or on each render of a frequently used component (like an icon or something), but creating one or two globally should be fine.

src/hooks/use-ordered-rows.ts
```diff
@@ -2,9 +2,12 @@ import { useMemo } from 'preact/hooks';
...
+
 /**
  * Orders a list of rows based on provided order options.
  * Provided rows are not mutated, but a copy is returned instead.
+ * Values are compared preserving their type.
```
robertknight, line 10: I would say something like, "strings are compared using `Intl.Collator`, other types are compared using standard JavaScript comparisons". The reader can refer to the `Intl.Collator` docs to understand the fine details.

src/hooks/use-ordered-rows.ts
```diff
@@ -15,16 +18,19 @@ export function useOrderedRows<Row>(
       return rows;
     }
 
-    return [...rows].sort((a, b) => {
-      if (a[order.field] === b[order.field]) {
-        return 0;
+    return [...rows].sort(({ [order.field]: a }, { [order.field]: b }) => {
+      const x = order.direction === 'ascending' ? a : b;
+      const y = order.direction === 'ascending' ? b : a;
```
robertknight, line 23: Could also do `const [x, y] = order.direction === 'ascending' ? [a, b] : [b, a]`.

src/hooks/use-ordered-rows.ts
```diff
@@ -15,16 +18,19 @@ export function useOrderedRows<Row>(
...
       }
 
-      if (order.direction === 'ascending') {
-        return a[order.field] > b[order.field] ? 1 : -1;
+      if (x === y) {
+        return 0;
       }
 
-      return a[order.field] > b[order.field] ? -1 : 1;
+      return x > y ? 1 : -1;
```
robertknight, line 33: As someone who has been working a lot with floats recently, I feel tempted to point out `NaN` issues (seriously though, don't worry about that!)

robertknight (approved):

src/hooks/use-ordered-rows.ts
acelaya: 😅

PR #1540, opened 2024-04-30
Add popover capabilities to SelectNext listbox

acelaya:
Depends on https://github.com/hypothesis/frontend-shared/pull/1548

Add `popover` attribute to the `SelectNext` listbox, and open it via `element.togglePopover()` when the browser supports it.

For browsers not supporting the native popover, we continue to handle it as before.

### TODO

- [x] Test with a screen reader, making sure the use of `[popover]` does not have any side effects on how components are announced.
- [x] Add/fix tests
- [x] Update documentation mentioning the popover behavior of the listbox in browsers that support it.

### Considerations

- Adding the `popover` attribute to the listbox makes it render in the top layer, and therefore, it is no longer positioned relative to the toggle button.
    Because of that, we have to manually calculate where should it be based on some heuristics, and absolutely position it there.
- ~Related with the point above, and because of how `useArrowKeyNavigation` internally works, I had to disable the automatic option focusing that `useArrowKeyNavigation` provides, and add an extra `toggle` event listener that mimics that logic.~
   **EDIT:** This has been solved. See https://github.com/hypothesis/frontend-shared/pull/1540#issuecomment-2100546401
- ~Another side effect of rendering the listbox in the top layer, is that it won't be affected by scrolling on inner elements.~
    ~**EDIT:** This is now mitigated via `listboxAsPopover` prop. With that we can keep the listbox as is, and enable the "popover mode" only when really needed.~
    **EDIT 2:** Finally we are capturing the scroll event and re-adjust the listbox positioning dynamically.
- At some point, I considered creating a new `Popover` component, that can be used in other places, but this is complex enough, without having to think in more abstractions. I will delay that to after this PR has been merged, and we have at least other use case for a popover.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -5,6 +5,7 @@ import {
   useContext,
   useId,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
 } from 'preact/hooks';
@@ -15,6 +16,7 @@ import { useFocusAway } from '../../hooks/use-focus-away';
 import { useKeyPress } from '../../hooks/use-key-press';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
 import type { CompositeProps } from '../../types';
+import { ListenerCollection } from '../../util/listener-collection';
 import { downcastRef } from '../../util/typing';
 import { MenuCollapseIcon, MenuExpandIcon } from '../icons';
 import { inputGroupStyles } from './InputGroup';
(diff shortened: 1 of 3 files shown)


src/components/input/SelectNext.tsx
```diff
@@ -213,7 +267,7 @@ function SelectMain<T>({
   useArrowKeyNavigation(listboxRef, {
     horizontal: false,
     loop: false,
-    autofocus: true,
+    autofocus: !nativePopoverSupported,
```
acelaya, line 270: Automatically focusing the corresponding option results in the item being focused before we had the chance to reposition the popover, causing the page to scroll to the very top.

Instead, when the browser supports popovers natively, we do not focus the item as part of `useArrowKeyNavigation`, but via `toggle` event on the listbox. See https://github.com/hypothesis/frontend-shared/pull/1540/files#diff-92242547c6b4471729edbe9319b52f8bf068f28ed9a0d6ed681a4cdeba48e777R153

src/components/input/SelectNext.tsx
```diff
@@ -273,21 +336,22 @@ function SelectMain<T>({
       </button>
       <SelectContext.Provider value={{ selectValue, value }}>
         <ul
+          // @ts-expect-error Popover attribute not yet part of type
+          popover
```
acelaya, line 340: The `popover attribute is missing in preact's JSX types. We discussed defining our custom type, but I ended up just using `@ts-expect-error`, as it's more straightforward and easier to remove once this has been fixed upstream.

We have also used a similar approach for some other attributes, like in [`Slider`](https://github.com/hypothesis/frontend-shared/blob/3d2b282325399cd65276ceb1a4107070c7b078a6/src/components/transition/Slider.tsx#L84)

robertknight:
> Adding the popover attribute to the listbox makes it render in the top layer, and therefore, it is no longer positioned relative to the toggle button.
Because of that, we have to manually calculate where should it be based on some heuristics, and absolutely position it there.

I think this is fine. In future you could use CSS anchor positioning for this. That is an extremely recent addition to the web platform, so not viable yet.

> Without this, the option gets focused first, when the listbox is still positioned at the very top of the page, making the page scroll up. Then the listbox gets repositioned.

My understanding is that after the listbox is opened, if `autofocus` is enabled in the `useArrowKeyNavigation` hook, the sequence of events is:

1. The `SelectNext` component's layout effect runs and schedules a state update for the position values
2. The `useArrowKeyNavigation` hook's effect runs and autofocuses its preferred element
3. The `SelectNext` component is re-rendered. This applies the position value update.

So the problem here is the ordering of steps (2) and (3).

If step (1) updated the DOM style synchronously, then the listbox would have its correct position by the time `useArrowKeyNavigation` runs.

> At some point, I considered creating a new Popover component, that can be used in other places, but this is complex enough, without having to think in more abstractions. I will delay that to after this PR has been merged, and we have at least other use case for a popover.

Yes, that makes sense.

robertknight:
One other thing that occurred to me is that you could try to make the popover and non-popover use cases more similar by rendering the listbox children into their own element which is positioned at the top-level of the DOM, like how React portals work. The downside is that event bubbling won't work, so you'd either have to emulate it, or warn consumers about this hazard.

acelaya:
> My understanding is that after the listbox is opened, if `autofocus` is enabled in the `useArrowKeyNavigation` hook, the sequence of events is:
> 
> 1. The `SelectNext` component's layout effect runs and schedules a state update for the position values
> 2. The `useArrowKeyNavigation` hook's effect runs and autofocuses its preferred element
> 3. The `SelectNext` component is re-rendered. This applies the position value update.
> 
> So the problem here is the ordering of steps (2) and (3).
> 
> If step (1) updated the DOM style synchronously, then the listbox would have its correct position by the time `useArrowKeyNavigation` runs.

You are right! If the styles are directly set via `listboxElement.style.{prop} = ...` instead of setting via the `style={positioningStyle}` prop with a piece of state, then there's no need to re-render and things work in the right order.

acelaya:
The only uncovered line that codecov is complaining about, is marked as `istanbul ignore next`.

src/components/input/SelectNext.tsx
```diff
@@ -103,38 +105,108 @@ function SelectOption<T>({
...
+      return setListboxCSSProps({
+        minWidth: `${buttonWidth}px`,
+        top: shouldListboxDropUp
+          ? `calc(${absBodyTop + buttonDistanceToTop - listboxHeight}px - ${LISTBOX_TOGGLE_GAP})`
+          : `calc(${absBodyTop + buttonDistanceToTop + buttonHeight}px + ${LISTBOX_TOGGLE_GAP})`,
+        left:
+          right && listboxWidth > buttonWidth
+            ? `${buttonLeft - (listboxWidth - buttonWidth)}px`
+            : `${buttonLeft}px`,
+      });
```
robertknight, line 177: If the cleanup was skipped, this could be simplified to an `Object.assign` on the listbox element's `styles` property.

src/components/input/SelectNext.tsx
```diff
@@ -166,6 +238,13 @@ export type SelectProps<T> = CompositeProps & {
 
   'aria-label'?: string;
   'aria-labelledby'?: string;
+
+  /**
+   * Test seam.
+   * Used to determine if current browser supports native popovers.
+   * Defaults to `'popover' in document.body`
```
robertknight, line 245: How the component determines whether native popovers are supported can be an implementation detail, so this last line isn't needed.

Also, perhaps `usePopover` or `useNativePopover` would be a more accurate name, since this prop determines whether the Popover API is used, not whether the browser actually supports it.

robertknight (commented):
The implementation generally looks good and worked as expected in all browsers that I tested. The main feedback I have is to add some comments for some of the more subtle issues in the code: why the listbox is being positioned manually, why the styles are applied synchronously, why we're using a popover in the first place. You don't need to repeat everything in the PR description. Just brief notes to raise awareness will do.

src/components/input/SelectNext.tsx
```diff
@@ -103,38 +105,108 @@ function SelectOption<T>({
...
+     * Sets some CSS props in the listbox, and returns a cleanup function that
+     * resets them to their default values.
+     */
+    const setListboxCSSProps = (
+      props: Partial<Record<ListboxCSSProps, string>>,
+    ) => {
+      const entries = Object.entries(props) as Array<[ListboxCSSProps, string]>;
+      entries.forEach(([prop, value]) => (listboxEl.style[prop] = value));
+
+      return () => entries.map(([prop]) => (listboxEl.style[prop] = ''));
```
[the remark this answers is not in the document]
acelaya: We do, because the effect depends on `isListboxOpen`, so it runs every time the listbox is shown/hidden.

Without the cleanup, if the listbox was first slidden-up, then you scroll further down and click the toggle again, it keeps the previous styles which conflict with the new ones.

Basically, we reset the styles every time the listbox is closed, and set new ones when opened.

src/components/input/SelectNext.tsx
acelaya: We can't skip the cleanup, but the suggestion on using `Object.assign` might still be interesting. Let me try that.

src/components/input/SelectNext.tsx
acelaya: Good point.

acelaya: I'm going with `listboxAsPopover`, because `use...` sounds like a hook (for a moment I thought your suggestion was going in that line)

src/components/input/test/SelectNext-test.js
```diff
@@ -21,13 +21,26 @@ describe('SelectNext', () => {
    *        Whether to renders SelectNext.Option children with callback notation.
    *        Used primarily to test and cover both branches.
    *        Defaults to true.
+   * @param {boolean} [options.defaultListboxAsPopover] -
+   *        Whether we should let the `listboxAsPopover` prop use default value
+   *        if not explicitly provided, or we should initialize it instead.
```
robertknight, line 26: Assuming most tests don't set this option, it means we'll be explicitly disabling the popover usage, right? We probably want tests to use the default behavior of the component except for tests that are specifically opting out.

robertknight (approved):
The added comments LGTM. I did have a minor query about the default value of `listboxAsPopover` in tests. It looks like the default is to turn this behavior off except for tests that opt in. Is that intended?

src/components/input/test/SelectNext-test.js
acelaya: Yep, I agree. However, there are some tests failing when using the popover mode, mostly around testing that focusing works as expected in certain scenarios (first option gets focused when opened, the toggle recovers focus when closed, etc.). I assume it's related with the use of the top layer, but I couldn't confirm it.

I tried a couple of times to make them pass, but finally decided to disable the popover mode by default for the sake of getting existing tests to pass, and focus on popover mode only for new tests.

I will address this in a follow-up PR, as I think you are definitely right, but I wanted to offload some indirect complexity from this one.

Issue #1532, opened 2024-04-24
Generate and publish API documentation


robertknight:
https://patterns.hypothes.is currently includes manually created pages describing individual components. However this is not a comprehensive API reference. Publishing API docs would make it possible to link to them in discussions and from the guide pages. I've used [TypeDoc](https://typedoc.org) for this before, but there might be other tools worth looking at.

Slack thread: https://hypothes-is.slack.com/archives/C1M8NH76X/p1713959531923039

PR #1541, opened 2024-05-03
Add API reference documentation

robertknight:
This PR adds API reference documentation to the pattern library website, built using TypeDoc.

The docs are built by running `yarn apidoc`, which will generate them in `build/api/`. The contents of this directory are exposed under the `/api` path on the website. There is a link to the main page of the API reference in the left navigation bar, plus links to corresponding pages at the top of each component page.

The generated documentation has some issues:

- The index page is very noisy because of all the near-identical `{IconName}Props` types for icons, and the `{ComponentName}Props` type for each component. I think ideally we just want the actual functions on the entry page, and these will link to relevant types.
- The types for components are comprised of a union of several subtypes. One of the most important of these is `ComponentProps`, which contains the props unique to a particular component, but this is not exported. This means the component-specific props are missing. One way to fix this would be to revise the pattern so that the `ComponentProps` type is inlined into the `{ComponentName}Props` type.
- References to external types from Preact are not links
- Each component's `props` argument is named `__namedParameters` in the docs, because no formal name is specified. This can be fixed by adding a `@param` tag, or by giving the parameter a proper name and destructuring into props inside the function.

Part of https://github.com/hypothesis/frontend-shared/issues/1532.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM node:19.8.1-alpine as builder
 COPY . /frontend-shared
 RUN cd /frontend-shared && \
     yarn install --frozen-lockfile && \
-    yarn build-pattern-lib
+    yarn build-pattern-lib && \
+    yarn apidoc
 
 
 FROM nginx:1.24.0-alpine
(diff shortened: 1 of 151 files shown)


robertknight:
I will come back to this another time when I have more bandwidth.

PR #1548, opened 2024-05-09
Use utilities from @hypothesis/frontend-testing over local ones

acelaya:
This PR removes all local testing utilities and replaces them with the ones from @hypothesis/frontend-testing


robertknight (approved):
👍

PR #1552, opened 2024-05-10
Run SelectNext tests in popover mode by default

acelaya:
Remove the logic making `SelectNext` test disable the popover mode by default, and only enable it when explicitly requested.

This was initially done because it caused some problems while working on https://github.com/hypothesis/frontend-shared/pull/1540, but I found all tests keep passing after doing these changes now. I suppose the first issues I faced might have been caused by some preliminary logic that was finally refactored.

I took the opportunity to add a few more combinations of props.


robertknight (approved):
LGTM

PR #1558, opened 2024-05-15
Allow dynamically loaded example files

acelaya:
This PR adds support to  individual examples files that can be used in the pattern library docs both to generate actual working examples that are part of the bundle, while at the same time we load them as un-processed plain text that can be inlined as code snippets.

Those files are type-checked, linted and formatted, ensuring they are always up to date with the actual source code and they follow our standard rules.

In order to achieve this, this PR adds a new `examples` folder inside `src/pattern-library`. Files in this folder are statically exposed via express/nginx, so that we can do a `fetch('/examples/filename.tsx')` and load them as plain text.

At the same time, we can dynamically import those files in order to use the actual source code, with `import('/examples/filename.tsx')`.

This PR extends the `Library.Demo` component to accept an `exampleFile` prop, `<Library.Demo exampleFile="some-file" />`, which transparently does both things internally by loading the contents of the referenced example file.

Additionally, the existing `<Library.Code />` component has been enhanced to also allow an `exampleFile` prop instead of `content`, in order to load the source code from an external file.

This PR also adds a couple examples as a proof of concept of this new functionality.

### TODO

- [x] ~Consider caching/memoizing the content of examples that have already been `fetch`ed.~ I finally refactored a bit the logic, so that the file is loaded when `Demo` renders, not every time the tab changes. That way all files are only fetched once when the page loads.
- [x] Document limitations of the approach (extension must be `.tsx`, no nested dirs supported, recommendation to prefix file names, requirement of a default export, etc.)

### Considerations

Some alternative approach to get the static example file content could go into using some babel or rollup plugin that captures the use of `fs.readFileSync` or similar. I found a couple but they were mostly unmaintained and outdated, and I saw some reported issues regarding incompatibilities with modern practices, so I decided to go the `fetch` way instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ RUN rm -r /usr/share/nginx/html && rm /etc/nginx/conf.d/default.conf
 COPY --from=builder /frontend-shared/build /usr/share/nginx/html
 COPY ./templates/index.html /usr/share/nginx/html/index.html
 COPY ./images /usr/share/nginx/html/images
+COPY ./src/pattern-library/examples /usr/share/nginx/html/examples
 COPY conf/nginx.conf /etc/nginx/conf.d/default.conf
 
 EXPOSE 5001
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "@hypothesis/frontend-testing": "^1.2.2",
     "@rollup/plugin-babel": "^6.0.0",
     "@rollup/plugin-commonjs": "^25.0.0",
+    "@rollup/plugin-dynamic-import-vars": "^2.1.2",
     "@rollup/plugin-node-resolve": "^15.0.0",
     "@rollup/plugin-virtual": "^3.0.0",
     "@trivago/prettier-plugin-sort-imports": "^4.1.1",
--- a/scripts/serve-pattern-library.js
+++ b/scripts/serve-pattern-library.js
@@ -11,6 +11,10 @@ export function servePatternLibrary(port = 4001) {
   app.use('/scripts', express.static(path.join(dirname, '../build/scripts')));
   app.use('/styles', express.static(path.join(dirname, '../build/styles')));
   app.use('/images', express.static(path.join(dirname, '../images')));
+  app.use(
+    '/examples',
+    express.static(path.join(dirname, '../src/pattern-library/examples')),
+  );
 
   // For any other path, serve the index.html file to allow client-side routing
   app.get('/:path?', (req, res) => {
--- a/src/pattern-library/components/Library.tsx
+++ b/src/pattern-library/components/Library.tsx
@@ -1,7 +1,7 @@
 import classnames from 'classnames';
 import { toChildArray, createContext } from 'preact';
 import type { ComponentChildren, JSX } from 'preact';
-import { useMemo, useState, useContext } from 'preact/hooks';
+import { useState, useContext, useEffect } from 'preact/hooks';
 import { Link as RouteLink } from 'wouter-preact';
 
 import {
@@ -11,7 +11,7 @@ import {
   Scroll,
   ScrollContainer,
 } from '../../';
-import { jsxToHTML } from '../util/jsx-to-string';
+import { highlightCode, jsxToHTML } from '../util/jsx-to-string';
 
 /**
  * Components for rendering component documentation, examples and demos in the
@@ -165,8 +165,50 @@ function Example({ children, title, ...htmlAttributes }: LibraryExampleProps) {
   );
 }
 
+function SimpleError({ message }: { message: string }) {
+  return (
+    <div className="w-full text-red-600 p-2 border rounded border-red-600">
+      {message}
+    </div>
+  );
+}
+
+/**
+ * Fetches provided example file and returns its contents as text, excluding
+ * the import statements.
+ * An error is thrown if the file cannot be fetched for any reason.
+ */
+async function fetchCodeExample(
+  exampleFile: string,
+  signal: AbortSignal,
+): Promise<string> {
+  const res = await fetch(`/examples/${exampleFile}.tsx`, { signal });
+  if (res.status >= 400) {
+    throw new Error(`Failed loading ${exampleFile} example file`);
+  }
+
+  const text = await res.text();
+
+  // Remove import statements and trim trailing empty lines
+  return text.replace(/^import .*;\n/gm, '').replace(/^\s*\n*/, '');
+}
+
 export type LibraryDemoProps = {
-  children: ComponentChildren;
+  children?: ComponentChildren;
+
+  /**
+   * Example file to read and use as content (to be rendered with syntax
+   * highlighting).
+   * It should be relative to the `pattern-library/examples` dir, and include no
+   * file extension: `exampleFile="some-example"`.
+   *
+   * The file needs to have a default export, which will be used to render the
+   * interactive example.
+   *
+   * If provided together with `children`, `children` will take precedence.
+   */
+  exampleFile?: string;
+
   /** Extra CSS classes for the demo content's immediate parent container */
   classes?: string | string[];
   /** Inline styles to apply to the demo container */
@@ -179,6 +221,62 @@ export type LibraryDemoProps = {
   withSource?: boolean;
 };
 
+type DemoContentsResult =
+  | { children: ComponentChildren }
+  | {
+      code?: string;
+      example?: ComponentChildren;
+      codeError?: string;
+      exampleError?: string;
+    };
+
+function isStaticDemoContent(
+  contentResult: DemoContentsResult,
+): contentResult is { children: ComponentChildren } {
+  return 'children' in contentResult;
+}
+
+/**
+ * Determines what are the contents to be used for a Demo, which can be either
+ * an explicitly provided set of children, or the contents of an example file
+ * which is dynamically imported.
+ */
+function useDemoContents(
+  props: Pick<LibraryDemoProps, 'children' | 'exampleFile'>,
+): DemoContentsResult {
+  const [code, setCode] = useState<string>();
+  const [example, setExample] = useState<ComponentChildren>();
+  const [codeError, setCodeError] = useState<string>();
+  const [exampleError, setExampleError] = useState<string>();
+
+  useEffect(() => {
+    if (!props.exampleFile) {
+      return () => {};
+    }
+
+    import(`../examples/${props.exampleFile}.tsx`)
+      .then(({ default: Example }) => setExample(<Example />))
+      .catch(() =>
+        setExampleError(
+          `Failed loading ../examples/${props.exampleFile}.tsx module`,
+        ),
+      );
+
+    const controller = new AbortController();
+    fetchCodeExample(props.exampleFile, controller.signal)
+      .then(setCode)
+      .catch(e => setCodeError(e.message));
+
+    return () => controller.abort();
+  }, [props.exampleFile, props.children]);
+
+  if (props.children) {
+    return { children: props.children };
+  }
+
+  return { code, example, codeError, exampleError };
+}
+
 /**
  * Render a "Demo", with optional source. This will render the children as
  * provided in a tabbed container. If `withSource` is `true`, the JSX source
@@ -186,25 +284,16 @@ export type LibraryDemoProps = {
  * rendered Demo content.
  */
 function Demo({
-  children,
   classes,
   withSource = false,
   style = {},
   title,
+  ...rest
 }: LibraryDemoProps) {
   const [visibleTab, setVisibleTab] = useState('demo');
-  const source = toChildArray(children).map((child, idx) => {
-    return (
-      <li key={idx}>
-        <code>
-          <pre
-            className="font-pre whitespace-pre-wrap break-words text-sm"
-            dangerouslySetInnerHTML={{ __html: jsxToHTML(child) }}
-          />
-        </code>
-      </li>
-    );
-  });
+  const demoContents = useDemoContents(rest);
+  const isStaticContent = isStaticDemoContent(demoContents);
+
   return (
     <div className="my-8 p-2 space-y-1">
       <div className="flex items-center px-2">
@@ -251,14 +340,41 @@ function Demo({
                 classes,
               )}
             >
-              {children}
+              {!isStaticContent ? demoContents.example : demoContents.children}
+              {!isStaticContent && demoContents.exampleError && (
+                <SimpleError message={demoContents.exampleError} />
+              )}
             </div>
           </div>
         )}
         {visibleTab === 'source' && (
-          <div className="border w-full rounded-md bg-slate-7 text-color-text-inverted p-4">
-            <ul>{source}</ul>
-          </div>
+          <>
+            {!isStaticContent ? (
+              demoContents.code && <Code content={demoContents.code} />
+            ) : (
+              <div className="border w-full rounded-md bg-slate-7 text-color-text-inverted p-4">
+                <ul>
+                  {toChildArray(demoContents.children).map((child, idx) => {
+                    return (
+                      <li key={idx}>
+                        <code>
+                          <pre
+                            className="font-pre whitespace-pre-wrap break-words text-sm"
+                            dangerouslySetInnerHTML={{
+                              __html: jsxToHTML(child),
+                            }}
+                          />
+                        </code>
+                      </li>
+                    );
+                  })}
+                </ul>
+              </div>
+            )}
+            {!isStaticContent && demoContents.codeError && (
+              <SimpleError message={demoContents.codeError} />
+            )}
+          </>
         )}
       </div>
     </div>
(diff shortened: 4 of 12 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "@hypothesis/frontend-testing": "^1.2.2",
     "@rollup/plugin-babel": "^6.0.0",
     "@rollup/plugin-commonjs": "^25.0.0",
+    "@rollup/plugin-dynamic-import-vars": "^2.1.2",
```
acelaya, line 17: This plugin allows dynamic imports where the path is also dynamic, like `import(`../examples/${somefile}.tsx`).

It has the limitation that it scans for potential files when building the first time, but if new files are added to the examples directory, you need to restart the dev server.

scripts/serve-pattern-library.js
```diff
@@ -11,6 +11,10 @@ export function servePatternLibrary(port = 4001) {
   app.use('/scripts', express.static(path.join(dirname, '../build/scripts')));
   app.use('/styles', express.static(path.join(dirname, '../build/styles')));
   app.use('/images', express.static(path.join(dirname, '../images')));
+  app.use(
+    '/examples',
+    express.static(path.join(dirname, '../src/pattern-library/examples')),
+  );
```
acelaya, line 17: This allows for the example files to be loaded as plain text static files via `fetch('/examples/foo.tsx')`.

Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN rm -r /usr/share/nginx/html && rm /etc/nginx/conf.d/default.conf
 COPY --from=builder /frontend-shared/build /usr/share/nginx/html
 COPY ./templates/index.html /usr/share/nginx/html/index.html
 COPY ./images /usr/share/nginx/html/images
+COPY ./src/pattern-library/examples /usr/share/nginx/html/examples
```
acelaya, line 16: This allows for the example files to be loaded as plain text static files via `fetch('/examples/foo.tsx')`. I haven't fully tested it though.

src/pattern-library/components/Library.tsx
```diff
@@ -265,6 +265,86 @@ function Demo({
...
 
+export type LibraryExampleDemoProps = Omit<LibraryDemoProps, 'children'> & {
+  /**
+   * Example file to read and use as content (to be rendered with syntax
+   * highlighting)
+   */
+  exampleFile: string;
+};
+
+function ExampleDemo({
```
acelaya, line 276: This component duplicates a lot of code from the existing `Demo` one. The plan is to consolidate that as much as possible.

src/pattern-library/examples/non-popover-listbox.tsx
acelaya, line 1: One limitation of these files is that they need to work by themeselves, with as little external resources as possible. That can lead to duplicated code between examples.

However, since they are regular modules, we could extract parts that are duplicated and import them, as long as we consider those are not critical for the code example when looking at the source code.

For example, here `items.map` is mostly self-explanatory, and we could extract the list of items to another bundle.

src/pattern-library/components/Library.tsx
```diff
@@ -265,6 +265,86 @@ function Demo({
...
+  withSource,
+  style,
+}: LibraryExampleDemoProps) {
+  const [visibleTab, setVisibleTab] = useState('demo');
+  const [children, setChildren] = useState<ComponentChildren>();
+
+  useEffect(() => {
+    import(`../examples/${exampleFile}.tsx`).then(({ App }) => {
+      setChildren(<App />);
+    });
```
acelaya, line 289: When importing an example file as a module, this assumes it exports an `App` symbol. We could think on better approaches, perhaps by relaying on the default export instead, as it is less restrictive.

src/pattern-library/examples/non-popover-listbox.tsx
acelaya, line 1: One limitation of these files is that they need to work by themeselves, with as little external resources as possible. That can lead to duplicated code between examples.

However, since they are regular modules, we could extract parts that are duplicated and import them, as long as we consider those are not critical for the code example when looking at the source code.

For example, here items.map is mostly self-explanatory, and we could extract the list of items to another bundle.

src/pattern-library/examples/non-popover-listbox.tsx
```diff
@@ -0,0 +1,70 @@
+import { useId, useState } from 'preact/hooks';
+
+import SelectNext from '../../components/input/SelectNext';
```
robertknight, line 3: The import here uses an internal path, but it would be nicer if written using the same syntax as a consumer of the library would write.

Since the import path is already documented on the page, we could also just save some space when the source is rendered by stripping out imports.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -578,43 +572,7 @@ export default function SelectNextPage() {
...
-              </div>
-            </>
-          )}
-        </SelectNext.Option>
-      ))}
-    </SelectNext>
-  );
-}`}
-          />
+          <Library.Code exampleFile="select-next-main" />
```
robertknight, line 575: The basic "how do I use this" example needs to be at the top of the page, with a demo/source toggle.

package.json
robertknight: > It has the limitation that it scans for potential files when building the first time, but if new files are added to the examples directory, you need to restart the dev server.

From reading the [source](https://github.com/rollup/plugins/blob/master/packages/dynamic-import-vars/src/index.js) for the plugin, it looks to me like modifying the source code of the file which contains the import statement should also work, as it executes the glob as part of processing such an import.

src/pattern-library/components/Library.tsx
robertknight, line 276: This would benefit from a comment to state what the component is for. "ExampleDemo" is not super descriptive. Did you plan to consolidate this with `Library.Demo` so there is just one component, or just share parts of the implementation?

robertknight (commented):
The basic idea of having the example in an external `.tsx` file which is then imported as script to render as a live demo, and also imported as text to render as syntax highlighted code, makes sense. It will probably be useful to keep the ability to have "inline" demos where the code is generated from the component, for simple use cases.

package.json
acelaya: Oh, good point.

src/pattern-library/components/Library.tsx
acelaya: Yep, it's a common practice in the framework world (Remix routes and Next.js pages, for example)

src/pattern-library/components/Library.tsx
acelaya: > Did you plan to consolidate this with `Library.Demo` so there is just one component, or just share parts of the implementation?

I haven't decided yet, but having a single component would be good if it doesn't make things too messy.

In an ideal world, we should aim to allow these three scenarios:

1. We pass explicit children to <Library.Demo />` and the source example is generated from it (the already supported use case):
    ```tsx
    <Library.Demo title="Basic Thumbnail" withSource>
      <div className="w-[250px]">
        <Thumbnail>
          <img src="https://placekitten.com/400/400" alt="kitty" />
        </Thumbnail>
      </div>
    </Library.Demo>
    ```
2. We pass an `exampleFile` prop and let both children and source be generated from it (what this PR is introducing)
    ```tsx
    <Library.Demo title="Basic Thumbnail" withSource exampleFile="basic-thumbnail" />
    ```
3. Any of the options above, but the source code is also explicitly provided the same way we do in `<Library.Code />`, just in case explicit children or the external example file do not fully cover the usage example we want to showcase. 
    ```tsx
    <Library.Demo 
      title="Basic Thumbnail" 
      withSource 
      exampleFile="basic-thumbnail"
      code={`
        import Foo from '...';

        function Foo() {
          // [...]
        }
      `}
    />
    ```
    ```tsx
    <Library.Demo 
      title="Basic Thumbnail" 
      withSource 
      code={`
        import Foo from '...';

        function Foo() {
          // [...]
        }
      `}
    >
      <div className="w-[250px]">
        <Thumbnail>
          <img src="https://placekitten.com/400/400" alt="kitty" />
        </Thumbnail>
      </div>
    </Library.Demo>
    ```

We can leave case 3 out of this PR for simplicity, but have it in mind as part of the design.

All those cases will require making some decisions in the component's props API, like not allowing to pass both children and `exampleFile`, or assuming `withSource` is true if `code` is provided.

src/pattern-library/examples/non-popover-listbox.tsx
acelaya: Right! I thought about that, as it would be nicer to have `import { SelectNext } from '@hypothesis/frontend-shared';`.

I also thought about stripping imports as a simple workaround. I think I'll try that for now.

src/pattern-library/components/Library.tsx
acelaya: I have finally included the new behavior in the existing `Demo` component via a new `useDemoContentAndSource` hook which has the conditional logic.

With that, cases 1 and 2 from my comment above should be possible.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
acelaya: I have converted the first example (the basic one) into one using this new example-file approach, and removed this example at the bottom entirely.

It was only a workaround due to the lack of code snippets for the other existing examples, so it's no longer necessary anyway.

src/pattern-library/components/Library.tsx
```diff
@@ -332,48 +396,91 @@ function ChangelogItem({ status, children }: LibraryChangelogItemProps) {
...
+  return 'content' in props;
+}
+
+function useCodeContent(
+  props: CodeContentProps,
+): [string | undefined, Error | undefined] {
+  const [codeMarkup, setCodeMarkup] = useState<string>();
+  const [error, setError] = useState<Error>();
+  const setJSXContent = useCallback(
+    (content: ComponentChildren) => setCodeMarkup(jsxToHTML(content)),
```
robertknight, line 431: `jsxToHTML` expects a JSX expression (`ComponentChildren`), but we are passing a string here. That happens to work because `string` is a valid `ComponentChildren` and `jsxToHTML` handles strings by not touching them ... however it is rather accidental. What would make more sense is to extract out the part of `jsxToHTML` which takes code as input and highlights it, and then call that here.

While you're at it, a very easy improvement in `jsxToHTML` will be to change the language from JavaScript to TypeScript. That will fix the lack of highlighting for the `type` keyword.

robertknight (commented):
I took another pass over this. One thing I notice is that the code passes source code to `jsxToHTML` and that happens to work, but the function is clearly expecting JSX expressions. It would make sense to add a separate function which takes in TypeScript source and returns the highlighted code. `jsxToHTML` can then call that internally.

src/pattern-library/components/Library.tsx
```diff
@@ -179,32 +199,94 @@ export type LibraryDemoProps = {
...
+
+  // Remove import statements and trim trailing empty lines
+  return text.replace(/^import .*;\n/gm, '').replace(/^\s*\n*/, '');
+}
+
+function useDemoChildrenAndSource(
+  props: Pick<LibraryDemoProps, 'children' | 'exampleFile'>,
+) {
+  const [source, setSource] = useState<ComponentChildren>();
+  const [children, setChildren] = useState<ComponentChildren>();
```
robertknight, line 226: In the case where the source and children are directly available, the values could be computed and used to initialize the returned state synchronously.

src/pattern-library/components/Library.tsx
```diff
@@ -332,48 +410,95 @@ function ChangelogItem({ status, children }: LibraryChangelogItemProps) {
...
+  return 'content' in props;
+}
+
+/**
+ * Dynamically resolves the content based on provided props.
+ * An error is optionally returned in case loading the content failed.
+ */
+function useCodeContent(
+  props: CodeContentProps,
+): [string | undefined, Error | undefined] {
```
robertknight, line 445: This function could return `undefined | string | Error` instead, to reflect the three possible states (loading, successful, failed).

src/pattern-library/components/Library.tsx
```diff
@@ -332,48 +410,95 @@ function ChangelogItem({ status, children }: LibraryChangelogItemProps) {
...
+  const [codeMarkup, setCodeMarkup] = useState<string>();
+  const [error, setError] = useState<Error>();
+
+  useEffect(() => {
+    if (isCodeWithContent(props)) {
+      setCodeMarkup(jsxToHTML(props.content));
```
robertknight, line 451: If the content is synchronously available, it could be returned on the first render rather than waiting for an effect.

src/pattern-library/examples/select-next-non-popover-listbox.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export default function App() {
+  const [value, setValue] = useState<ItemType>();
+  const buttonId = useId();
+
+  return (
+    <div className="w-full">
+      <p>
+        When not using the <code>popover</code> API, the listbox will be
+        constrained by its container dimensions.
```
robertknight, line 27: I think "container dimensions" is ambiguous here. Perhaps "may be clipped by parent elements that are styled to hide overflow".

src/pattern-library/util/jsx-to-string.tsx
```diff
@@ -156,20 +156,31 @@ export function jsxToString(vnode: ComponentChildren): string {
...
-  const code = jsxToString(vnode);
+
   return hljs.highlightAuto(code).value;
 }
+
+/**
+ * Render a JSX expression as syntax-highlighted HTML markup.
+ *
+ * The content returned by this function is sanitized and safe to use as
+ * `dangerouslySetInnerHTML` prop.
```
robertknight, line 181: Can you add this note to `highlightCode` as well.

src/pattern-library/components/Library.tsx
```diff
@@ -179,32 +199,94 @@ export type LibraryDemoProps = {
...
+  return text.replace(/^import .*;\n/gm, '').replace(/^\s*\n*/, '');
+}
+
+function useDemoChildrenAndSource(
+  props: Pick<LibraryDemoProps, 'children' | 'exampleFile'>,
+) {
+  const [source, setSource] = useState<ComponentChildren>();
+  const [children, setChildren] = useState<ComponentChildren>();
+
+  useEffect(() => {
```
robertknight, line 228: There was something about this effect that bothered me. I've realized now the issue is that it fetches data _and_ renders output. The more idiomatic approach would be to just fetch data and set state in the effect, then render the output based on the current state during the render.

src/pattern-library/components/Library.tsx
```diff
@@ -255,11 +337,7 @@ function Demo({
             </div>
           </div>
         )}
-        {visibleTab === 'source' && (
-          <div className="border w-full rounded-md bg-slate-7 text-color-text-inverted p-4">
-            <ul>{source}</ul>
-          </div>
-        )}
+        {visibleTab === 'source' && source}
```
robertknight, line 340: I noticed there is a slight visual flicker when switching between demo and code, both in Chrome and Safari. Locally I found that I could eliminate this by rendering both the demo and the source tabs but hiding the non-visible tab using `display: 'none'`.

robertknight (approved):
I think this can be merged, but I noted a few improvements for consideration:

- I noticed some visual flicker when switching between the demo and code tabs in Safari and Chrome. I noted a way to fix this using `display: 'none'`, but it would mean creating DOM nodes for the code even if the user hasn't looked at them, unless you add more code to lazily render the source.
- I suggested in the `useEffect` that it should only fetch the data and set state, but do the actual rendering synchronously as part of the render.
- `highlightCode` should have a note about its suitability for use with `dangerouslySetInnerHTML`

src/pattern-library/components/Library.tsx
acelaya: Hmm, that's a good idea.

src/pattern-library/components/Library.tsx
acelaya: It started like that, but I ended up moving to this approach for some reason I don't remember now 🤔 😅 

Let me check again.

acelaya: Oh, yeah. I moved to this approach because otherwise you end up with a `codeMarkupOrError` var, which needs to be type-checked in order to decide what to render.

Right now it looks like this:

```ts
const [codeMarkup, error] = useCodeContent(rest);

return (
  {codeMarkup && (...)}
  {error && <SimpleError message={error.message} />}
);
```

Using just one var, it would look like this:

```ts
const codeMarkupOrError = useCodeContent(rest);

return (
  {typeof codeMarkupOrError === 'string' && (...)}
  {typeof codeMarkupOrError === 'object' && <SimpleError message={codeMarkupOrError.message} />}
);
```

Which is a bit harder to read. I'll leave it as is, but we can change it afterwards if needed.

PR #1563, opened 2024-05-21
Move more examples from SelectNextPage to example files

acelaya:
This PR migrates a few more examples from `Selectnext` page to the `exampleFile` capability introduced in https://github.com/hypothesis/frontend-shared/pull/1558

While working on this, I have found a few nice-to-have things that we could eventually implement:

* Allow passing props to example files, so that the same example can be reused for very similar cases.
* Allow certain parts of the code snippet to be hidden, so that we can use relatively complex examples, but avoid distractions due to the boilerplate code.
  This could be done by introducing an opening/closing comment pattern that can be used to identify blocks and replace them with, let's say, a comment like `// [...]`.

### DIFF
--- a/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
@@ -1,11 +1,11 @@
 import classnames from 'classnames';
-import { useCallback, useId, useMemo, useState } from 'preact/hooks';
+import { useId, useState } from 'preact/hooks';
 
 import { Link } from '../../../..';
-import { ArrowLeftIcon, ArrowRightIcon } from '../../../../components/icons';
 import type { SelectNextProps } from '../../../../components/input';
-import { IconButton, InputGroup } from '../../../../components/input';
 import SelectNext from '../../../../components/input/SelectNext';
+import SelectNextInInputGroup from '../../../examples/select-next-in-input-group';
+import SelectNextWithManyOptions from '../../../examples/select-next-with-custom-options';
 import Library from '../../Library';
 
 type ItemType = {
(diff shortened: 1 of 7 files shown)


src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -1,11 +1,11 @@
...
-import { useCallback, useId, useMemo, useState } from 'preact/hooks';
+import { useId, useState } from 'preact/hooks';
 
 import { Link } from '../../../..';
-import { ArrowLeftIcon, ArrowRightIcon } from '../../../../components/icons';
 import type { SelectNextProps } from '../../../../components/input';
-import { IconButton, InputGroup } from '../../../../components/input';
 import SelectNext from '../../../../components/input/SelectNext';
+import SelectNextInInputGroup from '../../../examples/select-next-in-input-group';
+import SelectNextWithManyOptions from '../../../examples/select-next-with-custom-options';
```
acelaya, line 8: Explicitly importing some examples to reuse them in some cases where the code snippet is not that important, and it's more important to see how it looks when rendered. This way we don't duplicate that much code.

This could be addressed differently if we had a mechanism to pass props to example files loaded via `exampleFile`, but that is not the case at the moment.

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -23,20 +23,12 @@ const defaultItems: ItemType[] = [
...
 }: Pick<
   SelectNextProps<ItemType>,
-  | 'aria-label'
-  | 'aria-labelledby'
-  | 'buttonClasses'
-  | 'containerClasses'
-  | 'listboxClasses'
-  | 'disabled'
-  | 'right'
-  | 'listboxAsPopover'
```
acelaya, line 39: This local `SelectExample` component is still used in some cases, but I have removed all props which are no longer needed because they have been moved to example files.

robertknight (approved):
Looks good. Duplicating the example source to show variations in one prop is unfortunate, but the alternative is parametrizing the example which would make the individual examples more complex, so the duplication is probably the nicer thing for consumers of the library.

PR #1564, opened 2024-05-27
Allow to provide initial order per DataTable column

acelaya:
Add a new `initialColumnsOrderDir` prop to `DataTable` which is a partial record mapping column names with an order direction.

When a column becomes ordered, instead of always defaulting to `ascending` order, we now first check if that column is defined there. If not defined, we fall back to initially order in `ascending` direction. 

https://github.com/hypothesis/frontend-shared/assets/2719332/907b0a0d-3dbe-42fd-b46e-fe1457836d49

### DIFF
--- a/src/components/data/DataTable.tsx
+++ b/src/components/data/DataTable.tsx
@@ -4,7 +4,7 @@ import { useCallback, useContext, useEffect, useMemo } from 'preact/hooks';
 import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
 import { useStableCallback } from '../../hooks/use-stable-callback';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
-import type { CompositeProps, Order } from '../../types';
+import type { CompositeProps, Order, OrderDirection } from '../../types';
 import { downcastRef } from '../../util/typing';
 import { ArrowDownIcon, ArrowUpIcon, SpinnerSpokesIcon } from '../icons';
 import { Button } from '../input';
@@ -80,8 +80,13 @@ type ComponentProps<Row> = Pick<
    * Columns that can be used to order the table. Ignored if `onOrderChange` is
    * not provided.
    * No columns will be orderable if this is not provided.
+   *
+   * This can be a map of columns and order directions, to indicate the initial
+   * direction to use when a column becomes the ordered one
    */
-  orderableColumns?: Array<keyof Row>;
+  orderableColumns?:
+    | Array<keyof Row>
+    | Partial<Record<keyof Row, OrderDirection>>;
 
   /** Callback to render an individual table cell */
   renderItem?: (r: Row, field: keyof Row) => ComponentChildren;
@@ -98,9 +103,13 @@ function defaultRenderItem<Row>(r: Row, field: keyof Row): ComponentChildren {
 function calculateNewOrder<T extends string | number | symbol>(
   newField: T,
   prevOrder?: Order<T>,
+  initialOrderForColumn?: Partial<Record<T, OrderDirection>>,
 ): Order<T> {
   if (newField !== prevOrder?.field) {
-    return { field: newField, direction: 'ascending' };
+    return {
+      field: newField,
+      direction: initialOrderForColumn?.[newField] ?? 'ascending',
+    };
   }
 
   const newDirection =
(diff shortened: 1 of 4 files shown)


src/components/data/test/DataTable-test.js
```diff
@@ -74,7 +74,7 @@ describe('DataTable', () => {
     );
   }
 
-  const createComponent = (Component, props = {}) => {
+  const createComponent = ({ Component = DataTable, ...props } = {}) => {
```
acelaya, line 77: Not directly related with the main changes in this PR, but to avoid duplicating `createComponent(DataTable)` over and over, I added a new `Component` pseudo-prop/option here, which defaults to `DataTable`.

With this, we can just call `createComponent()` or `createComponent(props)` in most of the cases, and when a different component is needed, we can additionally pass `createComponent({ Component: Whatever, ...otherProps })`.

src/components/data/test/DataTable-test.js
```diff
@@ -425,7 +427,7 @@ describe('DataTable', () => {
     [
       // Clicking the same column when initially ascending, transitions to descending
       {
-        initialOrder: { field: 'name', direction: 'ascending' },
+        startingOrder: { field: 'name', direction: 'ascending' },
```
acelaya, line 430: I renamed `initialOrder` to `startingOrder` to avoid confusions with the new `initialColumnsOrderDir` prop.

src/components/data/DataTable.tsx
```diff
@@ -83,6 +83,11 @@ type ComponentProps<Row> = Pick<
    */
   orderableColumns?: Array<keyof Row>;
 
+  /**
+   * The initial order direction to set when a column becomes the ordered one.
+   */
+  initialColumnsOrderDir?: Partial<Record<keyof Row, OrderDirection>>;
```
robertknight, line 89: The name is a little awkward. Maybe something like `initialOrderForColumn`. Another possible API direction would be to combine this with `orderableColumns` so you'd specify which columns can be used for sorting as well as the associated direction in one prop.

src/components/data/test/DataTable-test.js
```diff
@@ -434,26 +436,54 @@ describe('DataTable', () => {
...
+      // Clicking another column sets direction default order when defined
+      {
+        startingOrder: { field: 'name', direction: 'ascending' },
+        clickedColumn: 'color',
+        expectedNewOrder: {
+          field: 'color',
+          direction: 'descending',
+        },
+      },
+      // No order sets direction as ascending
```
robertknight, line 461:
```suggestion
      // Change sort column to a column with no initial order specified
```

Same comment as above. The difference here is that there is no initial order, but I think stating that in a comment would just repeat what's obvious from the example data.

src/components/data/test/DataTable-test.js
```diff
@@ -434,26 +436,54 @@ describe('DataTable', () => {
...
+      {
+        startingOrder: undefined,
+        clickedColumn: 'consistency',
+        expectedNewOrder: {
+          field: 'consistency',
+          direction: 'ascending',
+        },
+      },
+      // No order sets direction as default order when defined
```
robertknight, line 470:
```suggestion
      // Change sort column to a column with an initial order specified
```

src/components/data/test/DataTable-test.js
```diff
@@ -434,26 +436,54 @@ describe('DataTable', () => {
...
-        initialOrder: { field: 'name', direction: 'ascending' },
+        startingOrder: { field: 'name', direction: 'ascending' },
         clickedColumn: 'consistency',
         expectedNewOrder: {
           field: 'consistency',
           direction: 'ascending',
         },
       },
-    ].forEach(({ initialOrder, clickedColumn, expectedNewOrder }) => {
+      // Clicking another column sets direction default order when defined
```
robertknight, line 452:
```suggestion
      // Change sort column to a column with no initial order specified
```

src/pattern-library/components/patterns/data/DataTablePage.tsx
```diff
@@ -725,6 +725,37 @@ export default function DataTablePage() {
             </Library.Demo>
           </Library.Example>
 
+          <Library.Example title="initialColumnsOrderDir">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                The initial order direction (<code>ascending</code> or{' '}
+                <code>descending</code>) to set when a column becomes the
```
robertknight, line 732:
```suggestion
                <code>descending</code>) to use when a column becomes the
```

robertknight (approved):
I had some feedback on the name, but looks good otherwise.

src/components/data/DataTable.tsx
acelaya: > Another possible API direction would be to combine this with `orderableColumns` so you'd specify which columns can be used for sorting as well as the associated direction in one prop.

Aha! I actually really like this approach. I even tried to restrict keys in `initialOrderForColumn` to the values in `orderableColumns`, but failed. This would solve that problem.

I'll try to allow both an array or a record to be passed to `orderableColumns`, for backwards compatibility and because it's more convenient if you don't want change the default direction, but let's see if that doesn't cause any problems.

PR #1574, opened 2024-06-04
Do not call onConfirmRow when Enter is pressed in DataTable row children

acelaya:
This PR fixes a bug where elements that can be interacted with <kbd>Enter</kbd> key would stop the event propagation when rendered inside a `DataTable` row, via `renderItem`.

### DIFF
--- a/src/components/data/DataTable.tsx
+++ b/src/components/data/DataTable.tsx
@@ -247,7 +247,9 @@ export default function DataTable<Row>({
 
   const handleKeyDown = useCallback(
     (event: KeyboardEvent, row: Row) => {
-      if (event.key === 'Enter') {
+      // Avoid preventing Enter key interactions in children elements by
+      // ignoring events not triggered by the row element itself
+      if (event.key === 'Enter' && event.target === event.currentTarget) {
         confirmRow(row);
         event.preventDefault();
         event.stopPropagation();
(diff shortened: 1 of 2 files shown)


src/components/data/test/DataTable-test.js
```diff
@@ -273,6 +273,18 @@ describe('DataTable', () => {
...
+      const onConfirmRow = sinon.stub();
+      const wrapper = createComponent({
+        onConfirmRow,
+        renderItem: (row, field) => <a href="/">{field}</a>,
+      });
+
+      wrapper.find('tbody tr a').first().simulate('keydown', { key: 'Enter' });
+
+      assert.notCalled(onConfirmRow);
+    });
```
acelaya, line 287: This test fails without the changes introduced in `DataTable`, while the test above keeps passing.

src/components/data/DataTable.tsx
```diff
@@ -247,7 +247,8 @@ export default function DataTable<Row>({
 
   const handleKeyDown = useCallback(
     (event: KeyboardEvent, row: Row) => {
-      if (event.key === 'Enter') {
+      // Ignore events triggered in children elements
```
robertknight, line 250: This comment says what the code does, but not why. The goal here is to avoid confirming the selection if the user presses "Enter" to activate an interactive child control.

robertknight (commented):
The change looks sensible, but I think the comment could do a better job of explaining the _why_ rather than _what_ of the logic.

robertknight (approved):
The change looks sensible, but I think the comment could do a better job of explaining the why rather than what of the logic.

PR #1586, opened 2024-06-19
Add fill attribute to pointer-up and pointer-down icons

acelaya:
Add a missing `fill="currentColor"` attribute to `PointerUp` and `PointerDown` colors. I added it to the svg files, and then regenerated icons.

Notice the color difference of the two bottom-left icons compored to the rest.

Before:

Captura desde 2024-06-19 09-04-09

After:

Captura desde 2024-06-19 09-04-15


robertknight (approved):
LGTM

PR #1587, opened 2024-06-19
Allow useOrderedRows to put all null elements at the bottom

acelaya:
This PR changes `useOrderedRows` so that it can make `null`/`undefined` values be put always at the bottom of the list, and makes this the default behavior.

This is done by adding a new `nullsLast` option to the `order` object, which is `true` by default.

Callers will need to explicitly switch from `nullsLast: true` to `nullsLast: false` if they want this to change as the order direction changes:

```tsx
const direction = 'descending';
const order = {
  field: 'display_name',
  direction,
  nullFirst: direction === 'descending',
}
```

Previously we were comparing nulls normally with other non-null values, which produced unintuitive results, because in JS both `'some string' > null` and `null > 'some string'` are `false`.

### DIFF
--- a/src/hooks/use-ordered-rows.ts
+++ b/src/hooks/use-ordered-rows.ts
@@ -19,6 +19,9 @@ export function useOrderedRows<Row>(
       return rows;
     }
 
+    // Order nulls last by default
+    const { nullsLast = true } = order;
+
     return [...rows].sort(({ [order.field]: a }, { [order.field]: b }) => {
       const [x, y] = order.direction === 'ascending' ? [a, b] : [b, a];
 
@@ -28,6 +31,15 @@ export function useOrderedRows<Row>(
 
       if (x === y) {
         return 0;
+      }
+
+      // We check a/b instead of x/y because nulls should not be affected by the
+      // regular order direction.
+      if (a === null || a === undefined) {
+        return nullsLast ? 1 : -1;
+      }
+      if (b === null || b === undefined) {
+        return nullsLast ? -1 : 1;
       }
 
       return x > y ? 1 : -1;
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,4 +64,11 @@ export type OrderDirection = 'ascending' | 'descending';
 export type Order<Field extends string | number | symbol> = {
   field: Field;
   direction: OrderDirection;
+
+  /**
+   * Indicates whether entries where the value for `field` is null/undefined
+   * should go last. Otherwise, they will go first.
+   * Defaults to true.
+   */
+  nullsLast?: boolean;
 };
--- a/src/hooks/test/use-ordered-rows-test.js
+++ b/src/hooks/test/use-ordered-rows-test.js
@@ -14,9 +14,9 @@ const starWarsCharacters = [
 ];
 
 describe('useOrderedRows', () => {
-  function FakeComponent() {
-    const [order, setOrder] = useState();
-    const orderedRows = useOrderedRows(starWarsCharacters, order);
+  function FakeComponent({ rows, initialOrder }) {
+    const [order, setOrder] = useState(initialOrder);
+    const orderedRows = useOrderedRows(rows, order);
 
     return (
       <div>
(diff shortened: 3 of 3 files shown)


src/hooks/use-ordered-rows.ts
```diff
@@ -30,6 +32,17 @@ export function useOrderedRows<Row>(
...
+      // We want null/undefined elements to be initially placed at the bottom of
+      // the list, but still move to the top if order direction changes.
+      // We check a/b instead of x/y because nulls should not be affected by the
+      // regular order direction.
+      if (a === null || a === undefined) {
+        return initialOrder.current?.direction === order.direction ? 1 : -1;
+      }
+      if (b === null || b === undefined) {
+        return initialOrder.current?.direction === order.direction ? -1 : 1;
+      }
```
acelaya, line 44: I think these two conditions should also evaluate that `initialOrder.current?.field === order.field`. I'm trying to verify it.

```suggestion
      if ((a === null || a === undefined) && initialOrder.current?.field === order.field) {
        return initialOrder.current?.direction === order.direction ? 1 : -1;
      }
      if ((b === null || b === undefined) && initialOrder.current?.field === order.field) {
        return initialOrder.current?.direction === order.direction ? -1 : 1;
      }
```

src/hooks/use-ordered-rows.ts
```diff
@@ -14,6 +14,8 @@ export function useOrderedRows<Row>(
   rows: Row[],
   order?: Order<keyof Row>,
 ): Row[] {
+  const initialOrder = useRef(order);
```
robertknight, line 17: Hmm... this makes the hook implicitly stateful, but that isn't mentioned in the documentation. Perhaps we should keep this hook stateless and just add a `nullsLast` option to `Order` to control whether `null` values sort before or after non-null values in a field. It will be up to the caller to then set this as appropriate for each field.

src/hooks/use-ordered-rows.ts
acelaya: Actually, I think the right thing to do is not to track `initialOrder` as a ref to `order`, but instead, keep a ref to a map of the initial order per field. That way each "column" will always start with its null values at the bottom.

robertknight (commented):
This implements what we discussed, but on reflection I wonder if perhaps it would be less surprising if we kept the hook stateless and instead gave the caller an option to determine whether `null` values sort before or after non-null values.

src/hooks/use-ordered-rows.ts
acelaya: Yep, this was the right approach ☝🏼

src/hooks/use-ordered-rows.ts
acelaya: Just FYI, I changed a bit what that ref tracks after you commented here, but it still makes it stateful, so your comment is still valid.

src/types.ts
```diff
@@ -64,4 +64,11 @@ export type OrderDirection = 'ascending' | 'descending';
 export type Order<Field extends string | number | symbol> = {
   field: Field;
   direction: OrderDirection;
+
+  /**
+   * Indicates entries where the value for `field` is null/undefined should go
+   * last. Otherwise, they will get compared with non-null values normally.
```
robertknight, line 70: Ah. My intention was that nulls would always sort _either_ before or after non-null values. This would be `true` by default, or `false` to make them sort before other values.

acelaya: Done

src/types.ts
```diff
@@ -64,4 +64,11 @@ export type OrderDirection = 'ascending' | 'descending';
 export type Order<Field extends string | number | symbol> = {
   field: Field;
   direction: OrderDirection;
+
+  /**
+   * Indicates entries where the value for `field` is null/undefined should go
```
robertknight, line 69:
```suggestion
   * Indicates whether entries where the value for `field` is null/undefined should go
```

robertknight (approved):

PR #1601, opened 2024-07-02
Allow multiple selection in SelectNext

acelaya:
This PR extends `SelectNext` so that it supports multiple items to be selected at once.

It does it by allowing `value` to be an array of items instead of a single item. When that happens, items in the list do not set themselves as selected, but they toggle themselves from that list instead.

Additionally, it adds a new `multiple` prop to be passed, which causes the listbox to be kept open when selecting options from the list.

State management is delegated to consumers, simplifying the implementation.

The implementation is inspired in how [HeadlessUI](https://headlessui.com/react/listbox#selecting-multiple-values) does it.

https://github.com/hypothesis/frontend-shared/assets/2719332/6c4a859b-5330-4652-80f0-ab6f9911a379

### TODO

- [x] Add tests
- [x] Check if there's any a11y implication to take into consideration.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -27,7 +27,12 @@ export type SelectOptionStatus = {
 };
 
 export type SelectOptionProps<T> = {
-  value: T;
+  /**
+   * An undefined value in a multiple select will cause the selection to reset
+   * to an empty array.
+   */
+  value: T | undefined;
+
   disabled?: boolean;
   children:
     | ComponentChildren
@@ -59,8 +64,34 @@ function SelectOption<T>({
     throw new Error('Select.Option can only be used as Select child');
   }
 
-  const { selectValue, value: currentValue } = selectContext;
-  const selected = !disabled && currentValue === value;
+  const { selectValue, value: currentValue, multiple } = selectContext;
+  const selected =
+    !disabled &&
+    ((multiple && currentValue.includes(value)) || currentValue === value);
+
+  const selectOrToggle = useCallback(() => {
+    // In single-select, just set current value
+    if (!multiple) {
+      selectValue(value);
+      return;
+    }
+
+    // In multi-select, clear selection for nullish values
+    if (!value) {
+      selectValue([]);
+      return;
+    }
+
+    // In multi-select, toggle clicked items
+    const index = currentValue.indexOf(value);
+    if (index === -1) {
+      selectValue([...currentValue, value]);
+    } else {
+      const copy = [...currentValue];
+      copy.splice(index, 1);
+      selectValue(copy);
+    }
+  }, [currentValue, multiple, selectValue, value]);
 
   return (
     <li
@@ -75,13 +106,13 @@ function SelectOption<T>({
       )}
       onClick={() => {
         if (!disabled) {
-          selectValue(value);
+          selectOrToggle();
         }
       }}
       onKeyPress={e => {
         if (!disabled && ['Enter', 'Space'].includes(e.code)) {
           e.preventDefault();
-          selectValue(value);
+          selectOrToggle();
         }
       }}
       role="option"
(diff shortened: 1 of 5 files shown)


src/components/input/SelectNext.tsx
```diff
@@ -60,7 +60,35 @@ function SelectOption<T>({
...
+    if (!valueIsArray) {
+      selectValue(value);
+      return;
+    }
+
+    // In multi-select, clear selection for nullish values
+    if (!value) {
+      selectValue([]);
+      return;
+    }
```
acelaya, line 80: This is a bit opinionated and not entirely obvious behavior.

Perhaps a better option would be to allow `<SelectOption />` to provide its own `onChange`, in which case it gets invoked instead of the default `selectValue(...)` logic.

src/components/input/SelectNext.tsx
```diff
@@ -387,6 +435,7 @@ function SelectMain<T>({
           role="listbox"
           ref={listboxRef}
           id={listboxId}
+          aria-multiselectable={multiple}
```
acelaya, line 438: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -425,6 +425,39 @@ export default function SelectNextPage() {
               withSource
             />
           </Library.Example>
+          <Library.Example title="multiple">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Determines if more than one item can be selected at once,
+                causing the listbox to stay open when an option is selected on
+                it.
+                <p>Considerations:</p>
```
robertknight, line 434: I think this paragraph and list can be simplified to just a description of how other aspects of the API change: "When multi-selection is enabled, `value` must be an array and `onChange` will receive an array as an argument". I think it is easier to follow if you only mention what changes and not aspects (eg. the fact that the caller maintains the state) which remain unchanged.

src/components/input/SelectNext.tsx
```diff
@@ -60,7 +60,35 @@ function SelectOption<T>({
   }
 
   const { selectValue, value: currentValue } = selectContext;
-  const selected = !disabled && currentValue === value;
+  const valueIsArray = Array.isArray(currentValue);
```
robertknight, line 63: An alternative to this would be testing the `multiple` prop and throwing an error if `currentValue` is not an array in that case, as such a situation is incorrect use of the API.

src/components/input/SelectNext.tsx
robertknight: I suggested above that it might help to avoid errors if the component just threw an error if in multi-select mode and a non-array value is passed.

robertknight:
I experimented with making `multiple` part of the `SingleValueProps` and `MultiValueProps` types so that TS can infer the types of `value`/`onChange` as appropriate. It works, except that it turned out to be necessary to explicitly specify the generic type for `SelectNext` rather than have TS infer it. That's quite an unusual thing to do, so it is probably better if that is not required.

```diff
diff --git a/src/components/input/SelectNext.tsx b/src/components/input/SelectNext.tsx
index 89762a4..c5b500c 100644
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -246,11 +246,13 @@ function useListboxPositioning(
 type SingleValueProps<T> = {
   value: T;
   onChange: (newValue: T) => void;
+  multiple: false | undefined;
 };
 
 type MultiValueProps<T> = {
   value: T[];
   onChange: (newValue: T[]) => void;
+  multiple: true;
 };
 
 export type SelectProps<T> = CompositeProps &
@@ -258,13 +260,6 @@ export type SelectProps<T> = CompositeProps &
     buttonContent?: ComponentChildren;
     disabled?: boolean;
 
-    /**
-     * Whether this select should allow multi-selection or not.
-     * When this is true, the listbox is kept open when an option is selected.
-     * Defaults to false.
-     */
-    multiple?: boolean;
-
     /**
      * `id` attribute for the toggle button. This is useful to associate a label
      * with the control.
diff --git a/src/pattern-library/examples/select-next-multiple.tsx b/src/pattern-library/examples/select-next-multiple.tsx
index 9034b4b..90b2778 100644
--- a/src/pattern-library/examples/select-next-multiple.tsx
+++ b/src/pattern-library/examples/select-next-multiple.tsx
@@ -22,7 +22,7 @@ export default function App() {
   return (
     <div className="w-96 mx-auto">
       <label htmlFor={selectId}>Select students</label>
-      <SelectNext
+      <SelectNext<ItemType>
         multiple
         value={values}
         onChange={setSelected}
```

robertknight (commented):
This looks good and accessibility works well with VoiceOver. I suggested that rather than trying to handle `values` being a non-array when `multiple` is true in some undocumented way, the component could simply throw an error.

src/components/input/SelectNext.tsx
robertknight, line 63: I think the multi vs single selection behavior should be provided explicitly by the context rather than trying to infer it from the type of `value`. Otherwise the component could behave in an unexpected way if you tried to use single-selection with items that were tuples.

src/components/input/SelectNext.tsx
```diff
@@ -60,7 +60,35 @@ function SelectOption<T>({
...
+
+  const selectOrToggle = useCallback(() => {
+    // In single-select, just set current value
+    if (!valueIsArray) {
+      selectValue(value);
+      return;
+    }
+
+    // In multi-select, clear selection for nullish values
+    if (!value) {
```
robertknight, line 77: As the component now enforces that `value` is an array for a multi-select, this shouldn't be possible any more.

robertknight (commented):
I noticed a possible hazard if single-selection is used but a user tried to use a tuple type as the item.

src/components/input/SelectNext.tsx
acelaya: True. We need to expose `multiple` via context as well.

I suspect TypeScript is going to force us to still verify `currentValue` is an array so that we can use it as such. Maybe I can try what you attempted [here](https://github.com/hypothesis/frontend-shared/pull/1601#issuecomment-2205588786), but only in the context type.

src/components/input/SelectNext.tsx
acelaya: `value` is the value passed for the particular option, not the `selectedValue` of the select.

This is used so that we can have something like `<SelectNext.Option value={undefined}>Reset selection</SelectNext.Option>`.

acelaya: [I mentioned](https://github.com/hypothesis/frontend-shared/pull/1601#discussion_r1662624153) this decision is probably too opinionated and confusing, and I think this is proving it.

I'll try to think on a different approach to achieve this, or perhaps delay this to a follow-up PR.

In my first comment I was proposing an alternative approach in the lines of `<SelectNext.Option onSelect={() => setValue([])}>Reset selection</SelectNext.Option>`, where `onSelect` is invoked only if provided, in addition, or perhaps instead, of the regular internal logic implemented in `SelectNext.Option`.

robertknight: > value is the value passed for the particular option, not the selectedValue of the select.

🤦‍♂️ . I follow. OK, the current API design actually seems reasonable to me in that case. Just make sure it is documented.

robertknight (approved):
LGTM. I do have a UX question over whether clicking a reset option should close the listbox automatically, rather than keep it open as clicking other options does. I suggest we ship this and then discuss that.

acelaya:
> I do have a UX question over whether clicking a reset option should close the listbox automatically, rather than keep it open as clicking other options does

Yep, I thought the same.

PR #1603, opened 2024-07-04
Allow to pass onListboxScroll callback to SelectNext

acelaya:
This PR adds a new `onListboxScroll` prop to `SelectNext`, which is basically a callback that gets propagated to the listbox `onScroll`.

We'll use this to dynamically update the options in the list when loading long lists in chunks.

I considered making this smarter, but then I thought there's too many possible requirements that could come from this, and finally decided to make it a dumb callback and let consumers implement the whole logic for now.

### DIFF
--- a/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
@@ -450,6 +450,21 @@ export default function SelectNextPage() {
               withSource
             />
           </Library.Example>
+          <Library.Example title="onListboxScroll">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                A callback passed to the listbox <code>onScroll</code>.
+              </Library.InfoItem>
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -450,6 +450,21 @@ export default function SelectNextPage() {
...
             />
           </Library.Example>
+          <Library.Example title="onListboxScroll">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                A callback passed to the listbox <code>onScroll</code>.
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
+                <code>
+                  () {'=>'} void {'|'} undefined
```
robertknight, line 460: This callback actually receives an `Event`, although I'm not sure whether you want to promise that in the API.

robertknight (approved):

PR #1619, opened 2024-07-16
Expose Select and MultiSelect aliases for SelectNext

acelaya:
This PR exposes two new aliases for `SelectNext`, which are `Select` and `MultiSelect`. These two implicitly have `multiple={false}` and `multiple={true}` respectively, and do not allow the `multiple` prop to be provided.

This change serves two purposes:

* The main one is to solve an issue introduced in https://github.com/hypothesis/frontend-shared/pull/1601, which causes `onChange`'s argument type to no longer be inferred based on the type of `value`.
`Select` and `MultiSelect` are now enforcing one of the two type paths when `mutiple` is true or false, fixing the type inference.
For more context on the issue, check [this comment](https://github.com/hypothesis/lms/pull/6449#issuecomment-2230591951) and the ones below.
* Additionally, it creates new simpler component names, where `Select` can transparently replace existing `SelectNext` usages, allowing us to eventually drop the `SelectNext` symbol in the next major release.

I tried to find a solution to the type inference at pure types definition level, but didn't quite get it, even after making types more complex, so I think this option is better, specially taking into consideration to additional benefit explained above.

> [!NOTE]  
> This PR is easier to review [ignoring whitespaces](https://github.com/hypothesis/frontend-shared/pull/1619/files?w=1).

### TODO

- [x] Document new component aliases and use them in examples to prove they work.
- [x] Test `Select` and `MultiSelect` in client and LMS projects, where `<SelectNext />` and `<SelectNext multiple />` is used, to confirm they work.
  - EDIT: Exposed these changes via yalc, and imported them in client and LMS, then replaced all `SelectNext` with `Select`. Everything works after a manual test, tests pass with a few minor changes, and typechecks pass.

### Out of scope

We can probably replace all usages in this library of `SelectNext` with either `Select` or `MultiSelect`, as they will provide the same coverage, and promote their usage over `SelectNext`.

Additionally, we could even mark `SelectNext` as deprecated as a standalone component, and recommend using one of the other two.

All that will come in follow-up PRs, to avoid distractions from the new API being introduced here.

### DIFF
--- a/src/components/input/SelectNext.tsx
+++ b/src/components/input/SelectNext.tsx
@@ -257,52 +257,57 @@ type MultiValueProps<T> = {
   onChange: (newValue: T[]) => void;
 };
 
-export type SelectProps<T> = CompositeProps &
-  (SingleValueProps<T> | MultiValueProps<T>) & {
-    buttonContent?: ComponentChildren;
-    disabled?: boolean;
+type BaseSelectProps = CompositeProps & {
+  buttonContent?: ComponentChildren;
+  disabled?: boolean;
 
-    /**
-     * Whether this select should allow multi-selection or not.
-     * When this is true, the listbox is kept open when an option is selected
-     * and the value must be an array.
-     * Defaults to false.
-     */
-    multiple?: boolean;
+  /**
+   * `id` attribute for the toggle button. This is useful to associate a label
+   * with the control.
+   */
+  buttonId?: string;
 
-    /**
-     * `id` attribute for the toggle button. This is useful to associate a label
-     * with the control.
-     */
-    buttonId?: string;
+  /** Additional classes to pass to container */
+  containerClasses?: string | string[];
+  /** Additional classes to pass to toggle button */
+  buttonClasses?: string | string[];
+  /** Additional classes to pass to listbox */
+  listboxClasses?: string | string[];
 
-    /** Additional classes to pass to container */
-    containerClasses?: string | string[];
-    /** Additional classes to pass to toggle button */
-    buttonClasses?: string | string[];
-    /** Additional classes to pass to listbox */
-    listboxClasses?: string | string[];
+  /**
+   * Align the listbox to the right.
+   * Useful when the listbox is bigger than the toggle button and this component
+   * is rendered next to the right side of the page/container.
+   * Defaults to false.
+   */
+  right?: boolean;
 
-    /**
-     * Align the listbox to the right.
-     * Useful when the listbox is bigger than the toggle button and this component
-     * is rendered next to the right side of the page/container.
-     * Defaults to false.
-     */
-    right?: boolean;
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
 
-    'aria-label'?: string;
-    'aria-labelledby'?: string;
+  /**
+   * Used to determine if the listbox should use the popover API.
+   * Defaults to true, as long as the browser supports it.
+   */
+  listboxAsPopover?: boolean;
 
-    /**
-     * Used to determine if the listbox should use the popover API.
-     * Defaults to true, as long as the browser supports it.
-     */
-    listboxAsPopover?: boolean;
+  /** A callback passed to the listbox onScroll */
+  onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
+};
 
-    /** A callback passed to the listbox onScroll */
-    onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
-  };
+export type SelectProps<T> = BaseSelectProps & SingleValueProps<T>;
+
+export type MultiSelectProps<T> = BaseSelectProps & MultiValueProps<T>;
+
+export type SelectNextProps<T> = (SelectProps<T> | MultiSelectProps<T>) & {
+  /**
+   * Whether this select should allow multi-selection or not.
+   * When this is true, the listbox is kept open when an option is selected
+   * and the value must be an array.
+   * Defaults to false.
+   */
+  multiple?: boolean;
+};
 
 function SelectMain<T>({
   buttonContent,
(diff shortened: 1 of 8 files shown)


src/components/input/SelectNext.tsx
```diff
@@ -257,51 +257,53 @@ type MultiValueProps<T> = {
...
+     * Whether this select should allow multi-selection or not.
+     * When this is true, the listbox is kept open when an option is selected
+     * and the value must be an array.
+     * Defaults to false.
      */
-    listboxAsPopover?: boolean;
-
-    /** A callback passed to the listbox onScroll */
-    onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
+    multiple?: boolean;
```
acelaya, line 306: considered moving this to `SingleValueProps` and `MultiValueProps`, but that breaks other things, so I decided to leave it as is.

With `Select` and `MultiSelect`, this prop's value will be implicit and consumers will not have to provide it, so it's not too bad.

src/components/input/SelectNext.tsx
```diff
@@ -474,8 +476,25 @@ function SelectMain<T>({
...
+    return <SelectNext<T> {...props} />;
+  },
+  { Option: SelectOption, displayName: 'Select' },
+);
 
-const SelectNext = Object.assign(SelectMain, { Option: SelectOption });
+export type MultiSelectProps<T> = BaseSelectProps & MultiValueProps<T>;
 
-export default SelectNext;
+export const MultiSelect = Object.assign(
```
acelaya, line 495: I choose `MultiSelect` over `SelectMulti` for consistency with `SingleValueProps` and `MultiValueProps`, but the latter allows for a nicer grouping of symbols when sorted alphabetically.

src/components/input/SelectNext.tsx
```diff
@@ -257,52 +257,57 @@ type MultiValueProps<T> = {
...
+export type MultiSelectProps<T> = BaseSelectProps & MultiValueProps<T>;
+
+export type SelectNextProps<T> = (SelectProps<T> | MultiSelectProps<T>) & {
+  /**
+   * Whether this select should allow multi-selection or not.
+   * When this is true, the listbox is kept open when an option is selected
+   * and the value must be an array.
+   * Defaults to false.
+   */
+  multiple?: boolean;
```
acelaya, line 309: I considered moving this to `SingleValueProps` and `MultiValueProps`, but that breaks other things, so I decided to leave it as is.

With `Select` and `MultiSelect`, this prop's value will be implicit and consumers will not have to provide it, so it's not too bad.

src/components/input/test/SelectNext-test.js
```diff
@@ -21,20 +21,26 @@ describe('SelectNext', () => {
    *        Whether to renders SelectNext.Option children with callback notation.
    *        Used primarily to test and cover both branches.
    *        Defaults to true.
+   * @param {MultiSelect | Select | SelectNext} [options.Component] -
```
acelaya, line 24: I'm not completely sure this JSDoc is correct

src/pattern-library/components/patterns/prototype/SelectNextPage.tsx
```diff
@@ -99,14 +99,15 @@ export default function SelectNextPage() {
...
+          <code>SelectNext</code> (and its aliases <code>Select</code> and{' '}
+          <code>MultiSelect</code>) are composite components which behave like
           the native <code>{'<select>'}</code> element.
         </p>
       }
     >
       <Library.Section>
         <Library.Pattern>
-          <Library.Usage componentName="SelectNext" />
+          <Library.Usage componentName="MultiSelect, Select, SelectNext" />
```
acelaya, line 110: image

src/pattern-library/examples/select-next-basic.tsx
```diff
@@ -17,18 +17,18 @@ export default function App() {
   return (
     <div className="w-96 mx-auto">
       <label htmlFor={selectId}>Select a person</label>
-      <SelectNext
+      <Select
         value={value}
-        onChange={setSelected}
+        onChange={newValue => setSelected(newValue)}
```
acelaya, line 22: I used an inline callback here just to demonstrate type is properly inferred with `Select`, as this file is type-checked.

src/components/input/SelectNext.tsx
```diff
@@ -257,52 +257,57 @@ type MultiValueProps<T> = {
...
+};
+
+export type SelectProps<T> = BaseSelectProps & SingleValueProps<T>;
 
-    /** A callback passed to the listbox onScroll */
-    onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
-  };
+export type MultiSelectProps<T> = BaseSelectProps & MultiValueProps<T>;
+
+export type SelectNextProps<T> = (SelectProps<T> | MultiSelectProps<T>) & {
```
robertknight, line 302: Other than being a breaking change, are there any reasons to keep `SelectNext` around? Can we replace all uses with either `Select` or `MultiSelect`?

acelaya: Nop, just the breaking change. I already have PRs in both client and LMS replacing all `SelectNext` instances and they all transparently work.

src/components/input/test/SelectNext-test.js
robertknight: The JSDoc here is only for information and not typed checked, so this is fine as long as the meaning is clear.

src/components/input/SelectNext.tsx
```diff
@@ -474,8 +479,21 @@ function SelectMain<T>({
...
-
-export default SelectNext;
+export const SelectNext = Object.assign(SelectMain, {
+  Option: SelectOption,
+  displayName: 'SelectNext',
+});
+
+export const Select = Object.assign(
+  function <T>(props: SelectProps<T>) {
+    return <SelectNext<T> {...props} />;
```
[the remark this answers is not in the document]
acelaya: Hmm, let me try that.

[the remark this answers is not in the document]
acelaya: Done

PR #1620, opened 2024-07-17
Refactor Select documentation to no longer mention SelectNext

acelaya:
> This PR follows-up on https://github.com/hypothesis/frontend-shared/pull/1619

This PR replaces the `/input-select-next` documentation page with `/input-select`, which has very similar content and examples, but always using `Select` and `MultiSelect` and no longer mentioning `SelectNext`.

Additionally, it deprecates `SelectNext`, and recommends using one of the other two components as drop-in replacements.

Technically speaking, two breaking changes are introduced here:

* The `/input-select-next` URL will no longer work. I guess we could potentially make it redirect to `/input-select` if we think this is a big problem.
* The `Option` component's `displayName` has changed from `'SelectNext.Option'` to `'Select.Option'`, which could make some test break, but I already have branches almost ready to transition away from SelectNext anyway, so it shouldn't be a big problem.
  * EDIT: I can confirm no tests break here, because we were finding options via `Select.Option` function reference, rather than via its display name.


acelaya:
Codecov is reporting a drop in code coverage, but I think it's a false positive. Locally, it reports a coverage of 100%, and if you go to codecov's report, there's no line market as uncovered.

src/pattern-library/components/patterns/prototype/SelectPage.tsx
```diff
@@ -527,6 +492,35 @@ export default function SelectNextPage() {
...
         </Library.Pattern>
       </Library.Section>
+
+      <Library.Section
+        title="MultiSelect"
+        intro={
+          <>
+            <p>
+              <code>MultiSelect</code> has the same public API as{' '}
+              <code>Select</code>, but allows multiple options to be selected at
```
robertknight, line 502: The public API is not _exactly_ the same, since `value` and `onChange` have different types. I would say something like, "MultiSelect is a variant of Select that allows multiple options to be selected at once. It has the same API except that `value` accepts, and `onChange` receives, arrays rather than a single item."

src/pattern-library/routes.ts
```diff
@@ -196,10 +196,10 @@ const routes: PlaygroundRoute[] = [
     route: '/input-option-button',
   },
   {
-    title: 'SelectNext',
+    title: 'Selects',
```
robertknight, line 199:
```suggestion
    title: 'Select',
```

All the other menu items are named after the main component on that page, so I think we should do the same here.

robertknight (approved):

src/pattern-library/routes.ts
acelaya: I used the plural form because that's what we do for `Dialog` and `ModalDialog` elements. The page's title is `Dialogs` and then there are two sub-sections for them.

PR #1629, opened 2024-07-22
Add new usePopoverShouldClose hook

acelaya:
This PR adds a convenient `usePopoverShouldClose` hook similarly to the deprecated `useElementShouldClose`, but internally combining `useClickAway`, `useFocusAway` and `useKeyPress` to make sure there's no implementation differences.

It also provides a more contextual name that makes it clear in which cases is supposed to be used, as opposed to the cases in which some of the other three individual hooks should be used instead.


src/hooks/test/use-popover-should-close-test.js
```diff
@@ -0,0 +1,52 @@
...
+  function createElement() {
+    return mount(<FakeComponent />);
+  }
+
+  it('combines useClickAway, useFocusAway and useKeyPress', () => {
+    createElement();
+
+    assert.called(fakeUseClickAway);
+    assert.called(fakeUseFocusAway);
+    assert.called(fakeUseKeyPress);
```
acelaya, line 50: Just mocking the other hooks and verifying they are called, as those hooks have their own tests.

src/hooks/test/use-popover-should-close-test.js
```diff
@@ -0,0 +1,69 @@
...
+        closeHandler,
+        options,
+      );
+      assert.calledWith(
+        fakeUseFocusAway,
+        sinon.match.any,
+        closeHandler,
+        options,
+      );
+      assert.calledWith(fakeUseKeyPress, ['Escape'], closeHandler, options);
```
acelaya, line 66: Just mocking the other hooks and verifying they are called, as those hooks have their own tests.

src/hooks/test/use-popover-should-close-test.js
```diff
@@ -0,0 +1,73 @@
...
+        sinon.match.any,
+        closeHandler,
+        options ?? {},
+      );
+      assert.calledWith(
+        fakeUseKeyPress,
+        ['Escape'],
+        closeHandler,
+        options ?? {},
+      );
```
acelaya, line 70: Just mocking the other hooks and verifying they are called, as those hooks have their own tests.

robertknight:
The missing line of coverage is being reported against `SelectMain`, unrelated to this PR. Looks like more fun with default arguments:

Missing coverage

acelaya:
> The missing line of coverage is being reported against `SelectMain`, unrelated to this PR. Looks like more fun with default arguments:
> 
> Missing coverage

Yeah, that is now always hardcoded as true or false when called through `Select` or `MultiSelect`. Once we remove `SelectNext` that can become a non-optional prop which will solve this.

robertknight:
> Once we remove SelectNext that can become a non-optional prop which will solve this.

Sounds good. Let's move forward and do that soon.

acelaya:
I thought I was merging something else 🤦🏼 

Now I need to create another PR to export the symbol 😅

PR #1630, opened 2024-07-23
Export usePopoverShouldClose from entry point

acelaya:
After pointing out in https://github.com/hypothesis/frontend-shared/pull/1629#pullrequestreview-2191827532 the fact that I forgot to expose `usePopoverShouldClose` from the library's entry point, I ended up forgetting to update that PR before merging it.

This PR adds what was missing there.

### DIFF
--- a/src/hooks/use-popover-should-close.ts
+++ b/src/hooks/use-popover-should-close.ts
@@ -20,7 +20,7 @@ export type UsePopoverShouldCloseOptions = {
  * @param handleClose - Callback invoked to close the popover
  * @param options
  */
-export default function usePopoverShouldClose(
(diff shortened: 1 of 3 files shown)


src/hooks/use-popover-should-close.ts
```diff
@@ -20,7 +20,7 @@ export type UsePopoverShouldCloseOptions = {
  * @param handleClose - Callback invoked to close the popover
  * @param options
  */
-export default function usePopoverShouldClose(
+export function usePopoverShouldClose(
```
acelaya, line 23: I switched to a non-default export for consistency with other hooks.

robertknight (approved):

PR #1640, opened 2024-08-02
Allow Select and MultiSelect listbox auto-closing to be configured

acelaya:
This PR adds a new optional `autoCloseListbox` prop to `Select` and `MultiSelect`, that can be used to decide if the listbox should automatically close after an option is selected.

Its default value is `true` for `Select` and `false` for `MultiSelect`, but in some cases, we may want the `MultiSelect` to auto-close the listbox.

It's probably less likely that we want the `Select`'s listbox to stay open, but providing the capability is cheap.

https://github.com/user-attachments/assets/4289140e-2901-4878-b631-21323c338d17

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -293,6 +293,12 @@ type BaseSelectProps = CompositeProps & {
    */
   listboxAsPopover?: boolean;
 
+  /**
+   * Whether the listbox should automatically close after an option is selected.
+   * Defaults to true for Select and false for MultiSelect.
+   */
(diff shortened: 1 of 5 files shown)


acelaya:
Closing for now, as we may be changing a bit how Select/MultiSelect work, or even provide a "combined" behavior somehow.

The reason is that some users got confused by how multiu-selects currently work, so we have some new UI and UX proposals. See https://hypothes-is.slack.com/archives/C07NXBDNW/p1723063781084279 for context.

PR #1647, opened 2024-08-08
Sortable indicator

acelaya:
Improve a bit discoverability of orderable `DataTable` columns by adding an icon indicating it.

Up until now, only the actively ordered column had an icon, making it impossible to tell which other columns could be used for ordering.

This is a need that was initially requested in https://hypothes-is.slack.com/archives/C07NXBDNW/p1722017120383689?thread_ts=1722016950.558659&cid=C07NXBDNW, and the design was provided here https://www.figma.com/design/rugQ2tONStRF3RL9UXZaGW/Search-Explorations?node-id=1291-1432.

Before:

After:

[Grabación de pantalla desde 2024-08-08 09-29-16.webm](https://github.com/user-attachments/assets/e44b9b32-12d0-409c-9ff7-50b23fae7570)


images/icons/orderable.svg
acelaya, line 1: I took this icon from Figma

robertknight (commented):
Conceptually the change makes sense. The icon shifts slightly when it transitions from being the orderable to ordered icons. Putting the `OrderableIcon` inside the same div might help, but with the background removed when the sorting is not active.

acelaya:
> Conceptually the change makes sense. The icon shifts slightly when it transitions from being the orderable to ordered icons. Putting the `OrderableIcon` inside the same div might help, but with the background removed when the sorting is not active.

Good catch

robertknight (approved):
LGTM

PR #1656, opened 2024-08-12
Use placehold.co to generate placeholder images

acelaya:
We were using `placekitten.com` to generate placeholder images, but the service is no longer available.

This PR replaces that service with https://placehold.co, which generates a bit more boring images, but should work as a valid replacement.

### DIFF
--- a/src/components/data/test/AspectRatio-test.js
+++ b/src/components/data/test/AspectRatio-test.js
@@ -12,11 +12,11 @@ describe('AspectRatio', () => {
 
   it('applies bottom-padding to container based on `ratio` value', () => {
     const wrapper16_9 = createComponent(AspectRatio, {
-      children: <img src="https://placekitten.com/400/400" alt="kitty" />,
+      children: <img src="https://placehold.co/400x400" alt="placeholder" />,
     });
 
     const wrapper4_3 = createComponent(AspectRatio, {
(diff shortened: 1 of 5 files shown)


robertknight (approved):
This is better than dead URLs, although it would be nice to have placeholders that allow you to see if the border gets cropped easily.

acelaya:
> This is better than dead URLs, although it would be nice to have placeholders that allow you to see if the border gets cropped easily.

Yep, that was my thinking, but I couldn't find a service returning "actual" images. Most follow this pattern of returning an image with the size in the middle.

I found one returning random images from flickr. I thought that was a no go 😅

robertknight:
> Yep, that was my thinking, but I couldn't find a service returning "actual" images. Most follow this pattern of returning an image with the size in the middle.

I discovered https://picsum.photos in my travels.

acelaya:
There's one thing we could do though. We always use the 400x400 size, except in one test where we use the 300x300 one.

I have the suspicion the test one is not important, and for the others, we could just have a locally-stored image. There's no need for an external service to always use the same image.

acelaya:
> > Yep, that was my thinking, but I couldn't find a service returning "actual" images. Most follow this pattern of returning an image with the size in the middle.
> 
> I discovered https://picsum.photos in my travels.

Hah! This one is nice

Issue #1658, opened 2024-08-19
New Select design that works better when MultiSelect is used


acelaya:
When we introduced support for multiple selection in selects via `MultiSelect`, we did not make any visual change to the component itself.

That means the listbox allows multiple options in the "selected" state when using that component, but we got some user feedback that it is not immediately obvious this is possible.

A new design was proposed, where checkboxes are displayed on every option when multi-selection is allowed, helping users understand when a select allows multiple options and when it doesn't: https://www.figma.com/design/rugQ2tONStRF3RL9UXZaGW/Search-Explorations?node-id=1862-693&t=dYvG3ythTTeUNTCy-0

![image](https://github.com/user-attachments/assets/1fa7b299-a265-48a4-b38d-39bb3b5b404d)

> Notice the design above also covers the topic of including an optional search box. That's not meant to be covered by this issue and will be addressed separately.

For consistency, the standard select (AKA single-select or simply `Select`) will suffer some UI changes as well (they can be seen in the design), which we want to propagate everywhere where the component is used.

### Behavior definition

Based on what is [described in slack](https://hypothes-is.slack.com/archives/C07NXBDNW/p1723063781084279), the listbox should have three possible modes, via the two existing components:

* `Select` -> single-selection only. Users won't be able to select more than one option.
* `MultiSelect` -> multi-selection. Options will always display a checkbox next to them to indicate more than one can be selected.
* `MultiSelect` with <kbd>Shift</kbd> key -> it will look like single-selection mode and behave like it (clicking an option will deselect previous one), but while <kbd>Shift</kbd> key is pressed, checkboxes will appear and multi-selection will be allowed.

I would suggest deferring the third option for now, and focusing on having a single select and a multi-select mode, each with the needed UI changes to make it more user friendly, and later enhance those modes if needed.

acelaya:
Only two pieces of work are left, but we have decided to defer them and implement them afterwards if needed.

I have tested the component in downstream projects and it works well:

https://github.com/user-attachments/assets/9570215f-d32d-431d-9610-fb9a25b4bfa6

https://github.com/user-attachments/assets/57f5b88f-9de4-4744-9181-d49275060fda

https://github.com/user-attachments/assets/74c8efc5-3d21-476f-ac2e-c716318442eb

https://github.com/user-attachments/assets/dadfcb91-774e-49da-a3ef-3b2d0005dab7

I'll close this as completed and release a new version of the library.

PR #1659, opened 2024-08-19
Adjust Select selected options UI

acelaya:
This PR is the first one for https://github.com/hypothesis/frontend-shared/issues/1658

This PR only changes how selected options look like in both `Select` and `MultiSelect`, to look like in the [new UI designs](https://www.figma.com/design/rugQ2tONStRF3RL9UXZaGW/Search-Explorations?node-id=1862-693&t=dYvG3ythTTeUNTCy-0):

image

Everything related with `MultiSelect` checkboxes will come after.

### Changes

Before:

[Grabación de pantalla desde 2024-08-19 11-56-42.webm](https://github.com/user-attachments/assets/7dde6c18-5d8c-442e-b4ed-8b5da9dc9058)

After:

[Grabación de pantalla desde 2024-08-19 11-55-39.webm](https://github.com/user-attachments/assets/626a36ec-c6fb-4cf6-bd62-149fae096279)

### Considerations

With these changes, selected options have a tick on the right side. This looks slightly awkward with some of the examples in the pattern library, specifically the ones with custom options.

[Grabación de pantalla desde 2024-08-19 11-33-52.webm](https://github.com/user-attachments/assets/705f36b9-5d5e-48a0-91a4-0ed32d242bb4)

However, customizing options is just a capability provided by the component, and not its own responsibility to ensure it looks good. Many other things could look awkward even with previous design, so I think this should be considered the consumer's responsibility.

This PR adds a second commit which adjusts those examples to look better.

[Grabación de pantalla desde 2024-08-19 12-35-24.webm](https://github.com/user-attachments/assets/c7340803-cefb-4ba6-aebb-b54e201a05b0)

### Todo

- [x] Adjust keyboard navigation/focus ring UI
- [x] Update existing examples which don't look great with the new UI

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -17,7 +17,7 @@ import { useSyncedRef } from '../../hooks/use-synced-ref';
 import type { CompositeProps } from '../../types';
 import { ListenerCollection } from '../../util/listener-collection';
 import { downcastRef } from '../../util/typing';
-import { MenuCollapseIcon, MenuExpandIcon } from '../icons';
+import { CheckIcon, MenuCollapseIcon, MenuExpandIcon } from '../icons';
 import { inputGroupStyles } from './InputGroup';
 import SelectContext from './SelectContext';
 
@@ -99,11 +99,10 @@ function SelectOption<T>({
     <li
       className={classnames(
         'w-full ring-inset outline-none rounded-none select-none',
-        'border-t first:border-t-0 whitespace-nowrap',
+        'px-1 mb-1 first:mt-1 whitespace-nowrap group',
         {
           'text-grey-4': disabled,
-          'cursor-pointer focus-visible-ring hover:bg-grey-1': !disabled,
-          'bg-grey-1 hover:bg-grey-2': selected,
+          'cursor-pointer': !disabled,
         },
         classes,
       )}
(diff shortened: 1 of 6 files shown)


src/components/input/Select.tsx
```diff
@@ -99,11 +99,10 @@ function SelectOption<T>({
     <li
       className={classnames(
         'w-full ring-inset outline-none rounded-none select-none',
-        'border-t first:border-t-0 whitespace-nowrap',
+        'px-1 mb-1 first:mt-1 whitespace-nowrap group',
```
acelaya, line 102: I have decided to implement spacing using margins and padding on the options, so that it's easier to eventually include the search boxes.

robertknight:
> If anything, we could update those examples, so that the ones in the pattern library look good, at least.

Yes. Some small tweaks to the item layout in that example would solve the problem.

src/components/input/Select.tsx
robertknight: This is fine, but can you clarify how this makes adding the search boxes easier?

robertknight (approved):
LGTM. The check definitely seems more visually obvious than the red bar.

src/components/input/Select.tsx
acelaya: Yes.

I initially considered adding a small padding to the surrounding listbox, but according to the designs, the search box container should span to the very sides of said listbox.

image

Adding the padding to every individual option instead, ensures that will be eventually possible.

PR #1664, opened 2024-08-19
Add checkboxes to multi-select options

acelaya:
Part of https://github.com/hypothesis/frontend-shared/issues/1658
Depends on https://github.com/hypothesis/frontend-shared/pull/1659

This PR changes `MultiSelect` options so that they show a checked/unchecked checkbox on the corresponding selected/non-selected options.

This brings the component closer to the [new UI design](https://www.figma.com/design/rugQ2tONStRF3RL9UXZaGW/Search-Explorations?node-id=1862-693&t=dYvG3ythTTeUNTCy-0).

This PR does not change the behavior in any way. Further changes will come in follow-up pull requests.

Before:

[Grabación de pantalla desde 2024-08-19 13-17-06.webm](https://github.com/user-attachments/assets/f6372122-101e-426c-8efe-5f6fce0fb0df)

After:

[Grabación de pantalla desde 2024-08-19 13-03-59.webm](https://github.com/user-attachments/assets/d86049ad-f8dd-4358-8f02-a5d528f7f332)

PR #1665, opened 2024-08-19
Make MultiSelect keep single selection unless checkbox is clicked

acelaya:
Depends on https://github.com/hypothesis/frontend-shared/pull/1664

Change `MultiSelect` behavior, so that clicking on an option only keeps that option selected, and you have to click on the checkbox specifically in order to select multiple options.

[Grabación de pantalla desde 2024-08-19 14-03-59.webm](https://github.com/user-attachments/assets/a369ccb7-9ecb-4599-aaab-938aac1bfca2)

### Todo

- [ ] Implement proper accessible markup
- [ ] Implement keyboard logic and focusable checkboxes
- [ ] Add tests


src/components/input/Select.tsx
```diff
@@ -112,15 +124,29 @@ function SelectOption<T>({
...
+      onClick={e => {
+        // Do not invoke callback if clicked element is the checkbox, as it has
+        // its own event handler.
+        if (!disabled && e.target !== checkboxRef.current) {
+          selectOneValue();
         }
       }}
-      onKeyPress={e => {
-        if (!disabled && ['Enter', 'Space'].includes(e.code)) {
+      onKeyDown={e => {
```
acelaya, line 134: This should have always been `onKeyDown`, not `onKeyPress`, as the `keypress` event [is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event).

src/components/input/Select.tsx
```diff
@@ -112,22 +126,37 @@ function SelectOption<T>({
...
       )}
-      onClick={() => {
-        if (!disabled) {
-          selectOrToggle();
+      onClick={e => {
+        // Do not invoke callback if clicked element is the checkbox, as it has
+        // its own event handler.
+        if (!disabled && e.target !== checkboxRef.current) {
+          selectOneValue();
         }
```
acelaya, line 134: I first tried to simply do `e.stopPropagation()` in the checkbox event handler, but it is an `onChange` handler that triggers after this `onClick`, so it didn't work.

That's why I did this explicit check here instead.

robertknight:
> I implemented it like in Sentry. When the right arrow is pressed, the checkbox for the currently focused option is focused, and when the left arrow is pressed, the focus returns to the option.

For the benefit of future developers, we are referring to this control in Sentry's issue browser:

Sentry issue list

src/components/input/Select.tsx
```diff
@@ -112,22 +126,37 @@ function SelectOption<T>({
         },
         classes,
       )}
-      onClick={() => {
-        if (!disabled) {
-          selectOrToggle();
+      onClick={e => {
+        // Do not invoke callback if clicked element is the checkbox, as it has
+        // its own event handler.
+        if (!disabled && e.target !== checkboxRef.current) {
```
robertknight, line 132: I wonder if we should use

src/components/input/Select.tsx
```diff
@@ -112,22 +126,37 @@ function SelectOption<T>({
...
       }}
-      onKeyPress={e => {
-        if (!disabled && ['Enter', 'Space'].includes(e.code)) {
+      onKeyDown={e => {
+        if (disabled) {
+          return;
+        }
+
+        if (
+          ['Enter', 'Space'].includes(e.code) &&
```
robertknight, line 142: Should this use `e.key` instead, ie. the logical rather than physical key that was pressed? I don't know of a keyboard layout where this affects these two keys, but I would guess that it is normally the logical keys that are tested.

src/components/input/test/Select-test.js
```diff
@@ -99,6 +99,45 @@ describe('Select', () => {
...
+
+    checkbox.simulate('change');
+  }
+
+  const pressKeyInOption = (wrapper, id, keyCode) =>
+    wrapper
+      .find(`[data-testid="option-${id}"]`)
+      .closest('[role="option"]')
+      .getDOMNode()
+      .dispatchEvent(new KeyboardEvent('keydown', { code: keyCode }));
```
robertknight, line 122: See comments about `key` vs `code` above.

Note that you can use `simulate('keydown', { args...} )` as well here instead of `getDOMNode().dispatchEvent(...)`.

src/components/input/test/Select-test.js
```diff
@@ -438,26 +453,87 @@ describe('Select', () => {
...
+      // Spy on checkbox focus
+      const checkbox = wrapper
+        .find(`[data-testid="option-${optionId}"]`)
+        .closest('[role="option"]')
+        .find('[type="checkbox"]')
+        .getDOMNode();
+      const focusStub = sinon.stub(checkbox, 'focus');
+
+      pressKeyInOption(wrapper, optionId, 'ArrowRight');
+      // assert.called(focusSpy);
```
robertknight, line 518: Commented out line of code left in

robertknight (commented):
This looks great. One minor detail I flagged up is whether we should be using `key` (the logical key) rather than `code` (the physical key) here. I believe it is very rare that these would be different for the Space and Enter keys specifically but we should be in the habit of knowing which we are using.

src/components/input/Select.tsx
acelaya: Right! It should be `key`, not `code`.

src/components/input/Select.tsx
```diff
@@ -112,22 +126,37 @@ function SelectOption<T>({
...
       }}
-      onKeyPress={e => {
-        if (!disabled && ['Enter', 'Space'].includes(e.code)) {
+      onKeyDown={e => {
+        if (disabled) {
+          return;
+        }
+
+        if (
+          ['Enter', 'Space'].includes(e.key) &&
```
robertknight, line 142:
```suggestion
          ['Enter', ' '].includes(e.key) &&
```

For `key` the value is the printable representation, if one exists.

src/components/input/test/Select-test.js
```diff
@@ -123,38 +160,26 @@ describe('Select', () => {
     assert.notCalled(onChange);
   });
 
-  ['Enter', 'Space'].forEach(code => {
-    it(`changes selected value when ${code} is pressed in option`, () => {
+  ['Enter', 'Space'].forEach(key => {
```
robertknight, line 163:
```suggestion
  ['Enter', ' '].forEach(key => {
```

acelaya:
I think the logic to use right/left arrows to move focus between the checkbox and the option, could potentially be implemented using `useArrowKeyNavigation` with horizontal navigation.

However, I did a quick check and there are some limitations:

* The hook does not expect to be defined in a component and its children, making them conflict with each other for some reason.
* One of the focusable options is the container itself, which the hook does not currently allow. Changing the markup would have other implications.

I will consider refactoring it later, and keep it as is for now.

robertknight (commented):
The value of `KeyboardEvent.key` is `' '` for the space key. Otherwise LGTM. I should add that I'm not completely certain whether `key` or `code` is the "correct" choice here, though `key` is definitely correct if you want to match printed keys. From a quick look at native select implementations, it seems they internally use the deprecated `keyCode`.

src/components/input/Select.tsx
acelaya: Oops. I quickly tested only with <kbd>Enter</kbd>, and missed this.

acelaya:
> I should add that I'm not completely certain whether `key` or `code` is the "correct" choice here, though `key` is definitely correct if you want to match printed keys

I quickly asked ChatGPT, and this was the answer:

> When checking if a key has been pressed, it is generally more correct to use `KeyboardEvent.key` for the following reasons:
> 
> 1. **Readability**: `KeyboardEvent.key` returns the actual value of the key pressed, which is more readable and intuitive (e.g., "a", "Enter", "ArrowDown").
> 2. **Localization**: `KeyboardEvent.key` takes into account the user's keyboard layout and locale, providing the correct character or key name.
> 3. **Consistency**: `KeyboardEvent.key` is consistent across different browsers and platforms.
> 
> `KeyboardEvent.code` is useful when you need to detect the physical key on the keyboard, regardless of the keyboard layout (e.g., "KeyA", "Enter", "ArrowDown"). This can be useful for games or applications where the physical key position is important.
> 
> In most cases, `KeyboardEvent.key` is preferred for general key detection.

Additionally, I quickly checked other keyboard event handler and we seem to use `key` always (or almost), so I think it makes sense to use that.

robertknight (approved):
LGTM

Issue #1668, opened 2024-08-20
Pattern library contains nested `<a>` for navigation links


robertknight:
**Steps to reproduce:**

1. Go to https://patterns.hypothes.is
2. Observe errors in the console

```
Improper nesting of interactive content. Your <a> should not have other anchor tags as child-elements.<a onClick="function () {}" href="/lms-content-selection" className="undefined" />

  in a
  in  (at /frontend-shared/src/pattern-library/components/PlaygroundApp.tsx:65)
  in NavLink (at /frontend-shared/src/pattern-library/components/PlaygroundApp.tsx:202)
  in PlaygroundApp (at /frontend-shared/src/pattern-library/index.tsx:44)
```

The `NavLink` component contains a `<Link>` wrapped in a `<RouteLink>` and both render `<a>` tags.

PR #1669, opened 2024-08-20
Avoid nested anchors when using Link inside RouterLink

acelaya:
Fixes https://github.com/hypothesis/frontend-shared/issues/1668

Add missing `asChild` prop to wouter's `Link`s when wrapping a custom `Link` component.

See documentation for context: https://github.com/molefrog/wouter?tab=readme-ov-file#link-hrefpath-

### DIFF
--- a/src/pattern-library/components/Library.tsx
+++ b/src/pattern-library/components/Library.tsx
@@ -573,7 +573,7 @@ export type LinkProps = {
  */
 function Link({ children, href }: LinkProps) {
   return (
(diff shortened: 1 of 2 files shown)


robertknight (approved):
LGTM. For reference, see the `asChild` implementation in https://github.com/molefrog/wouter/blob/3d1f732dff8d91a0fe3ef5cfa5213cecbd9fa781/packages/wouter/src/index.js#L271.

PR #1671, opened 2024-08-21
Add new Select.ClearOption

acelaya:

This PR is an alternative implementation to https://github.com/hypothesis/frontend-shared/pull/1670, where the "special" option to clear a MultiSelect is not identified by a prop, but instead it is exposed as a separate component, which is only exposed via `MultiSelect` and not via `Select` (although we could choose to expose it in both cases).

```tsx
function App() {
  const [students, setStudents] = useState<number[]>([]);

  return (
    <MultiSelect value={students} onChange={setStudents}>
      <MultiSelect.ClearOption>All students</MultiSelect.ClearOption>
      <MultiSelect.Option value={1}>Joh Doe</MultiSelect.Option>
      <MultiSelect.Option value={2}>Jane Doe</MultiSelect.Option>
    </MultiSelect>
  );
}
```

> Context and considerations here are the same described in https://github.com/hypothesis/frontend-shared/pull/1670

### Todo

- [ ] Document new component in pattern library.

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -5,6 +5,7 @@ import {
   useContext,
   useId,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
 } from 'preact/hooks';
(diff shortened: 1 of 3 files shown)


acelaya:
Benefits of this implementation of https://github.com/hypothesis/frontend-shared/pull/1670:

* The external API is simpler. You either use `Option` with a value, or use `ClearOption`.
* We can decide to expose `ClearOption` only through `MultiSelect`.

acelaya:
I'm putting this on hold, as the original issue that motivated this PR can actually be fixed in a much simpler way.

This still has a value, so we may end up merging it eventually.

PR #1672, opened 2024-08-21
Ensure options with value undefined are selected

acelaya:
Fix an existing issue in `MultiSelect`, where options with value `undefined`, which special purpose is to clear the selection entirely, setting an empty array, are never marked as selected.

Additionally, when this type of option is selected, the listbox is automatically closed regardless of how it is selected (by clicking on it or by clicking on its checkbox). This is because this option will never be selectable together with other options, so it doesn't really make sense to keep it open.

Because of the point above, we should probably consider a different look-and-feel for this type of option, but I'm deferring that.

https://github.com/user-attachments/assets/0ffdfd45-2442-4886-89cb-f62a1e26c9fd

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -5,6 +5,7 @@ import {
   useContext,
   useId,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
 } from 'preact/hooks';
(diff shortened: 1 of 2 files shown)


acelaya:
Using `<Option value={undefined} />` is not entirely intuitive, so we may want to move to a more idiomatic approach with a more clear API.

This draft proposes an alternative approach to do that https://github.com/hypothesis/frontend-shared/pull/1671, that we can merge later after the main issue is solved.

src/components/input/Select.tsx
```diff
@@ -96,10 +111,14 @@ function SelectOption<T>({
...
+    const options: SelectValueOptions = {
+      // Close listbox only if selected value is a "clear" option. Clear options
+      // are those with `undefined` value
+      closeListbox: value === undefined,
+    };
 
-    // In multi-select, clear selection for nullish values
-    if (!value) {
+    // In multi-select, clear selection for `undefined` values
+    if (value === undefined) {
```
acelaya, line 121: Being more explicit on what is the "special" value.

robertknight (commented):
Can you add a test case for the new behavior. Otherwise this looks good.

PR #1674, opened 2024-08-21
Adjust select options gap

acelaya:

Increase the gap between Select/MultiSelect options icons and their content, to match the horizontal padding and make it look less heaped when the content grows to fill all available space.

Before:

image

After:

image


acelaya:
I have tested the component in downstream projects and everything works well, so I'm going to go ahead and merge this, as it's a very small visual change.

PR #1681, opened 2024-08-26
Add a new `RadioButton` comppnent

acelaya:
Add a `RadioButton` component that uses part of the functionality from `Checkbox`, by extracting it to a reusable `ToggleInput` component.

https://github.com/user-attachments/assets/5aece1cc-f04a-4d19-8d56-6748f3fbf1ae

As opposed to the `Checkbox` component, which can be uncontrolled by internally handling state if props are not provided, `RadioButton` is expected to be controlled, because it has to be unchecked if another radio button from the same group gets checked.

Since every individual radio button does not know about the others, the state needs to be collectively handled by a parent component, otherwise, the custom icon never changes.

For convenience, in a follow-up PR I will provide a `RadioButtonGroup` component which will style a group of options as radio buttons.

### DIFF
--- a/src/components/input/test/RadioButton-test.js
+++ b/src/components/input/test/RadioButton-test.js
@@ -1,0 +1,28 @@
+import { mount } from 'enzyme';
+
+import { testCompositeComponent } from '../../test/common-tests';
+import RadioButton from '../RadioButton';
+
+// Relatively simple test, as most of the logic is shared with Checkbox, and
+// covered by Checkbox-test
+describe('RadioButton', () => {
+  const createComponent = (props = {}) => {
+    return mount(<RadioButton {...props}>This is child content</RadioButton>);
+  };
+
+  testCompositeComponent(RadioButton, {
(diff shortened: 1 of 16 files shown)


src/components/input/test/RadioButton-test.js
acelaya, line 1: I just provided a very basic test, as mist of the logic is already covered by the Checkbox test

robertknight: Can you add a comment near the top of the code to this effect.

robertknight:
> Since every individual radio button does not know about the others, the state needs to be collectively handled by a parent component, otherwise, the custom icon never changes.

To restate this for my understanding: a checkbox state change is always followed by an `onChange` event. The `Checkbox` component can use this to trigger a re-render and update its custom icons when it is managing its own state. A radio button however can transition from selected to de-selected without its `onChange` handler being called, by selecting a different button in the group. To solve this a parent component needs to own the state and re-render all the radio buttons in the group.

PR #1683, opened 2024-08-27
Create RadioGroup component

acelaya:
> Alternative implementation to https://github.com/hypothesis/frontend-shared/pull/1682, which does not use the `RadioButton` component from https://github.com/hypothesis/frontend-shared/pull/1681

Add a new `RadioGroup` component with `role="radiogroup"` that renders a group of items with `role="radio"` in it.

image

This implementation is inspired by how popular component libraries like [MUI](https://mui.com/material-ui/react-radio-button/) and [Healdess UI](https://headlessui.com/react/radio-group) do it, but is more simplistic for now.

### Todo

- [x] Add tests
- [x] Follow behavior described in https://www.w3.org/WAI/ARIA/apg/patterns/radio/
- [x] Consider removing the `items` prop and providing radios via `<RadioGroup.Radio />` components.

### Implementation decisions

1. ~To simplify the implementation for now, the list of radios is passed via the `items={[...]}` prop as an array of objects.~
    
    We could alternatively take a more declarative approach as we did with `Select`, and instead expose a `<RadioGroup.Radio />` component that gets the selected value and handlers via context, and lets the rest of the props be passed by the consumers:
    
    ```tsx
    <RadioGroup {...}>
      <RadioGroup.Radio value="1" label="Foo" />
      <RadioGroup.Radio value="2" label="Something" subtitle="Something else but longer" />
      <RadioGroup.Radio value="3" label="Disabled" disabled />
    </RadioGroup>
    ```
    
    EDIT: This is what I finally did, as we found two benefits for this approach:
    * More customizable children.
    * More flexibility when arranging radios without the limitations of `direction` prop.
2. I initially had the intention to build this component on top of `RadioButton` (see https://github.com/hypothesis/frontend-shared/pull/1682), but once I got hands-on, I realized it was impractical for a few reasons:
    * Every radio component needs to be more complex than the `RadioButton` allows. This forces the use of role="radio" so that it can be focused and interacted with the keyboard, but then we end up with a radio containing another radio, which I don't think is correct from an accessibility point of view.
    * A workaround for the above would involve allowing a more customizable layout in `RadioButton`, via classes or similar, but it feels like a hack just for the sake of using `RadioButton`.
    * `RadioButton` provides its own disabling capability via `disabled` prop, which among other things adds a 70% opacity to the component. If we do something similar in the wrapper, we would end up adding-up the 70% opacity twice just in part of the component, making it look like part of it is "more disabled".

### DIFF
--- a/src/components/input/RadioGroup.tsx
+++ b/src/components/input/RadioGroup.tsx
@@ -1,0 +1,153 @@
+import classnames from 'classnames';
+import type { ComponentChildren } from 'preact';
+import { useContext, useRef } from 'preact/hooks';
+
+import { useArrowKeyNavigation } from '../../hooks/use-arrow-key-navigation';
+import { RadioCheckedIcon, RadioIcon } from '../icons';
+import RadioGroupContext from './RadioGroupContext';
+
+type RadioValue = string | number;
+
+export type RadioProps<T extends RadioValue> = {
+  value: T;
+  /** Content provided as children is vertically aligned with the radio icon */
+  children: ComponentChildren;
+
+  /**
+   * Allows to provide extra content to be displayed under the children, in a
+   * smaller and more subtle font color.
+   */
+  subtitle?: ComponentChildren;
+
+  disabled?: boolean;
+};
+
+function Radio<T extends RadioValue>({
+  value,
+  children,
+  subtitle,
+  disabled: radioDisabled,
+}: RadioProps<T>) {
+  const radioGroupContext = useContext(RadioGroupContext);
+  if (!radioGroupContext) {
+    throw new Error('RadioGroup.Radio can only be used as RadioGroup child');
+  }
+
+  const { selected, disabled = radioDisabled, onChange } = radioGroupContext;
+  const isSelected = !disabled && selected === value;
+
+  return (
+    <div
+      role="radio"
+      aria-checked={isSelected}
+      aria-disabled={disabled}
+      className={classnames('focus-visible-ring rounded-lg px-3 py-2 grow', {
+        'bg-grey-2': isSelected,
+        'hover:bg-grey-1': !isSelected && !disabled,
+        'opacity-70': disabled,
+        'cursor-pointer': !disabled,
+      })}
+      data-value={value}
+      onClick={!disabled ? () => onChange(value) : undefined}
+      onKeyDown={
+        disabled
+          ? undefined
+          : e => {
+              if (['Enter', ' '].includes(e.key)) {
+                e.preventDefault();
+                onChange(value);
+              }
+            }
+      }
+      tabIndex={-1}
+    >
+      <div className="flex items-center gap-x-1.5">
+        {isSelected ? <RadioCheckedIcon /> : <RadioIcon />}
+        {children}
+      </div>
+      {subtitle && (
+        <div className="pl-4 ml-1.5 mt-1 text-grey-6 text-sm">{subtitle}</div>
+      )}
+    </div>
+  );
+}
+
+Radio.displayName = 'RadioGroup.Radio';
+
+export type RadioGroup<T extends RadioValue> = {
+  children: ComponentChildren;
+  selected?: T;
+  onChange: (newSelected: T) => void;
+
+  /**
+   * Determines the direction in which radios are stacked.
+   * Defaults to 'horizontal'.
+   */
+  direction?: 'vertical' | 'horizontal';
+
+  disabled?: boolean;
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+
+  /**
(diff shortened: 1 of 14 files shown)


src/pattern-library/components/patterns/input/RadioGroupPage.tsx
```diff
@@ -0,0 +1,189 @@
...
+          <Library.Demo
+            title="Via aria-labelledby"
+            withSource
+            exampleFile="radio-group-aria-labelledby"
+          />
+        </Library.Example>
+      </Library.Pattern>
+
+      <Library.Pattern title="Component API">
+        <Library.Example title="inputs">
```
robertknight, line 64: The logic behind the ordering of the props here is not apparent. I think we should keep it simple and list everything in alphabetical order followed by the `htmlAttributes` rest prop.

src/components/input/RadioGroup.tsx
```diff
@@ -0,0 +1,120 @@
...
+  /** Determines the direction in which radios are stacked */
+  direction?: 'vertical' | 'horizontal';
+
+  disabled?: boolean;
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+
+  /**
+   * If provided, it adds an input[type="hidden"] to the DOM with selected value,
+   * allowing to use this in forms as a regular form control
```
robertknight, line 28:
```suggestion
   * If provided, adds a hidden form control with the given name and the value set to
   * to the selected radio's value, for use in form submissions.
```

src/components/input/RadioGroup.tsx
```diff
@@ -0,0 +1,120 @@
...
+  input: { value, label, subtitle, disabled },
+  isSelected,
+  onChange,
+}: {
+  input: RadioInput<T>;
+  isSelected: boolean;
+  onChange: (newSelected: T) => void;
+}) {
+  return (
+    <div
```
robertknight, line 43: It looks to me like all the necessary semantic attributes have been added. You decided not to go with a hidden input?

robertknight:
> To simplify the implementation for now, the list of radios is passed via the items={[...]} prop as an array of objects.

The prop is called `inputs`, which seems obvious enough (`inputs`, `items` and `options` would all work). The main restriction of doing it this way is that you can't add extra content to children. Do we have a use case for that?

robertknight (commented):
Generally this looks very good. Some notes:

- The US spelling of "Labeled" as a single "l". We should use that everywhere. `aria-labelledby` is a mistake (see links in my comments)
- Setting inputs via an array will work for the moment, but it means we can't add custom content to items, so we may have to revise the API if that use case comes up
- The intro of the documentation needs a revision. I suggest we link to the [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) for cases where our components are implementations of those patterns. These guides describe the purpose, expected keyboard behavior and semantic labeling. If we choose to deviate from aspects of a pattern, we should note why.

src/components/input/RadioGroup.tsx
acelaya: I'm not sure I follow the question.

robertknight: Please ignore this. I was asking why you chose not to use a hidden radio input like the `RadioButton` component but I think that's answered in the PR description.

src/pattern-library/components/patterns/input/RadioGroupPage.tsx
acelaya: Right, I followed an almost random order 😅

robertknight (approved):
LGTM 👍

src/components/input/test/RadioGroup-test.js
```diff
@@ -0,0 +1,191 @@
...
+  function clickRadio(wrapper, index) {
+    getRadio(wrapper, index).simulate('click');
+  }
+
+  function pressKeyOnRadio(wrapper, index, key) {
+    wrapper
+      .find('[role="radio"]')
+      .at(index)
+      .getDOMNode()
+      .dispatchEvent(new KeyboardEvent('keydown', { key }));
```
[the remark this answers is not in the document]
acelaya: Yep, that works as well. I'll change it.

src/components/input/test/RadioGroup-test.js
```diff
@@ -0,0 +1,191 @@
...
+import RadioGroup from '../RadioGroup';
+
+describe('RadioGroup', () => {
+  let container;
+  let wrappers;
+
+  beforeEach(() => {
+    wrappers = [];
+    container = document.createElement('div');
+    document.body.appendChild(container);
```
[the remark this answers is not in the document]
acelaya: Oops

Issue #915, opened 2023-03-10
Document hooks in pattern library


lyzadanger:
At some point, it would be handy to add a documentation page for the hooks exposed by this package's API.

PR #1684, opened 2024-08-27
Add documentation for hooks

acelaya:
Part of https://github.com/hypothesis/frontend-shared/issues/915

This PR introduces a new section in the side menu to document hooks, and adds the documentation page for `useClickAway` as a proof of concept.

After this, we can continue documenting the rest of the hooks.

https://github.com/user-attachments/assets/24ca88b5-e1ef-45fd-9090-7696008a692c

### DIFF
--- a/src/pattern-library/components/patterns/hooks/UseClickAwayPage.tsx
+++ b/src/pattern-library/components/patterns/hooks/UseClickAwayPage.tsx
@@ -1,0 +1,88 @@
+import Library from '../../Library';
+
+export default function UseClickAwayPage() {
+  return (
+    <Library.Page
+      title="useClickAway"
+      intro={
+        <p>
+          <code>useClickAway</code> is a hook to listen for click events on{' '}
+          <code>document.body</code>. If an event{"'"}s target is outside the
+          container element, it invokes a callback.
+        </p>
+      }
+    >
+      <Library.Section>
+        <Library.Pattern>
+          <Library.Usage symbolName="useClickAway" />
+          <Library.Example>
+            <Library.Demo
+              withSource
+              title="Basic useClickAway"
+              exampleFile="use-click-away-basic"
+            />
+          </Library.Example>
+        </Library.Pattern>
+
+        <Library.Pattern title="Hook arguments">
+          <Library.Example title="container">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                A Ref to the DOM element used to evaluate if clicking happened{' '}
+                {'"'}
+                away{'"'}.
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
+                <code>{`RefObject<HTMLElement | undefined>`}</code>
+              </Library.InfoItem>
+            </Library.Info>
+          </Library.Example>
+          <Library.Example title="callback">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                The callback to be invoked when clicking away. The click event
+                object is passed to it.
+              </Library.InfoItem>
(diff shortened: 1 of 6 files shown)


src/pattern-library/components/patterns/hooks/UseClickAwayPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export default function UseClickAwayPage() {
+  return (
+    <Library.Page
+      title="useClickAway"
+      intro={
+        <p>
+          <code>useClickAway</code> is a hook that can be used to invoke a
+          callback when clicking outside of a container.
+        </p>
```
robertknight, line 11: I found this description and demo unclear compared to the JSDoc comment on the `useClickAway` hook. From looking at the demo you'd think that an `onClick` handler on the othermost div would be sufficient.

src/pattern-library/examples/use-click-away-basic.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const wrapperRef = useRef<HTMLDivElement | null>(null);
+
+  useClickAway(containerRef, e => {
+    if (
+      e.target === wrapperRef.current ||
+      wrapperRef.current?.contains(e.target as HTMLElement)
+    ) {
+      alert('You have clicked away');
+    }
+  });
```
robertknight, line 16: In this demo I think it is perhaps unclear why you'd use this hook as opposed to just an `onClick` listener on `wrapperRef`. Something like this where you can click anywhere in the page to trigger the state change would be more obvious:

```ts
export default function App() {
  const [clickedAway, setClickedAway] = useState(false);
  const containerRef = useRef<HTMLDivElement | null>(null);
  useClickAway(containerRef, () => setClickedAway(true));

  return (
    <div
      ref={containerRef}
      className={classnames(
        'rounded-lg h-32 w-full p-2 flex items-center justify-center',
        { 'bg-red-100': clickedAway, 'bg-amber-100': !clickedAway },
      )}
    >
      {clickedAway ? "You clicked outside the area" : "Click outside this area"}
      <Button onClick={() => setClickedAway(false)}>Reset</Button>
    </div>
  );
}
```

robertknight (commented):
The part about extending the menu to support documentation for hooks is fine. I think the documentation and example for this specific hook don't make it that clear why you'd use this hook instead of eg. a click handler on a parent element.

src/pattern-library/examples/use-click-away-basic.tsx
acelaya: Thanks! I have used your example.

src/pattern-library/components/patterns/hooks/UseClickAwayPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+          <Library.Example>
+            <Library.Demo
+              withSource
+              title="Basic useClickAway"
+              exampleFile="use-click-away-basic"
+            />
+          </Library.Example>
+        </Library.Pattern>
+
+        <Library.Pattern title="Hook options">
```
robertknight, line 27: This could use documentation for the positional arguments (`containerRef`, `callback`) as well.

robertknight (approved):

src/pattern-library/components/patterns/hooks/UseClickAwayPage.tsx
acelaya: Right!

acelaya: Done

PR #1685, opened 2024-08-28
Allow a ref to be passed to Select.Option components

acelaya:
This will allow options to be tracked from consuming code, for example, to programmatically focus one option.

The first use case we have for this is to be able to focus the last option of the listbox, before unmounting another element which is focused and would cause the listbox to close when focusing out.

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -1,5 +1,5 @@
 import classnames from 'classnames';
-import type { ComponentChildren, JSX, RefObject } from 'preact';
+import type { ComponentChildren, JSX, RefObject, Ref } from 'preact';
 import {
   useCallback,
   useContext,
@@ -46,6 +46,9 @@ export type SelectOptionProps<T> = {
     | ComponentChildren
     | ((status: SelectOptionStatus) => ComponentChildren);
   classes?: string | string[];
+
+  /** Ref associated with the option's container element */
+  elementRef?: Ref<HTMLElement | undefined>;
 };
 
 function optionChildren(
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/patterns/prototype/SelectPage.tsx
```diff
@@ -489,6 +489,33 @@ export default function SelectPage() {
...
+            <Library.Info>
+              <Library.InfoItem label="description">
+                CSS class(es) that will be appended to the CSS classes applied
+                to the options{"'"}s outermost element.
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
+                <code>string</code>
+              </Library.InfoItem>
+            </Library.Info>
+          </Library.Example>
```
acelaya, line 502: We forgot to document the classes prop, so I took the opportunity to add it.

src/components/input/Select.tsx
```diff
@@ -46,6 +46,9 @@ export type SelectOptionProps<T> = {
     | ComponentChildren
     | ((status: SelectOptionStatus) => ComponentChildren);
   classes?: string | string[];
+
+  /** Ref associated with outermost option element */
```
robertknight, line 50: "outermost" to me implies there will be nested `option` elements. I think what you mean to say is that this will be applied to the container element that has the semantic role `option` and is focusable. The "focusable" property is something that is important for our current use case.

robertknight (approved):
LGTM but I had some comments around the description of what the prop does.

src/components/input/Select.tsx
acelaya: Yes, that's what I meant. I'll reword it.

acelaya: Or perhaps it should have been `Ref associated with the option's outermost element`, but I think the "container" keyword is less confusing.

PR #1695, opened 2024-09-03
Fix too low contrast in subtitles for selected RadioGroup.Radios

acelaya:
This PR fixes an accessibility issue in `RadioGroup` component, where the contrast between the text color used in subtitles and the background color used in selected radios, is too low.

In order to fix it, we now set a darker shade of grey in the subtitles when the radio is hovered or selected.

See before and after:

image
image

Additionally, selected and hover radio background color is now set with an opacity lower than 100%. This ensures there's always a different color for those, even if the container's background color is the same as the color used for selected radios.

See how they look when the container has the former background color:

Before:

https://github.com/user-attachments/assets/350f629f-4235-40bb-b0cd-5857af90eef4

After:

https://github.com/user-attachments/assets/820821d2-8c53-4a23-8e65-8585d31ac6a6

### DIFF
--- a/src/components/input/test/RadioGroup-test.js
+++ b/src/components/input/test/RadioGroup-test.js
@@ -183,6 +183,19 @@ describe('RadioGroup', () => {
 
   it(
     'should pass a11y checks',
-    checkAccessibility({ content: createComponent }),
+    checkAccessibility([
+      {
+        name: 'no selected item',
+        content: createComponent,
+      },
+      {
+        name: 'selected item',
+        content: () => createComponent({ selected: 'one' }),
+      },
+      {
+        name: 'selected item with subtitle',
+        content: () => createComponent({ selected: 'four' }),
+      },
+    ]),
   );
 });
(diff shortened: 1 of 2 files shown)


src/components/input/test/RadioGroup-test.js
```diff
@@ -183,6 +183,19 @@ describe('RadioGroup', () => {
...
+      },
+      {
+        name: 'selected item',
+        content: () => createComponent({ selected: 'one' }),
+      },
+      {
+        name: 'selected item with subtitle',
+        content: () => createComponent({ selected: 'four' }),
+      },
+    ]),
```
acelaya, line 199: axe-core catched the accessibility issue in other project, but we were missing a test with selected radios.

These tests fail in `main`.

robertknight:
On macOS the thinner font looks like:

macOS thin font

An alternative would be to make the text one shade darker when the radio is selected or hovered:

macOS darker when selected

On my system I think I find this a little easier to read.

acelaya:
> An alternative would be to make the text one shade darker when the radio is selected or hovered

I actually thought about this, but forgot to test it! Let me try that.

acelaya:
> On macOS the thinner font looks like:
> 
> macOS thin font
> An alternative would be to make the text one shade darker when the radio is selected or hovered:
> 
> macOS darker when selected
> On my system I think I find this a little easier to read.

Yeah, I think this is a better solution. I have changed it.

robertknight (approved):
LGTM

PR #1699, opened 2024-09-06
Add new FilterClearIcon

acelaya:
Add new `filter-clear` icon to be used in the LMS dashboard for the button to clear filters.

image

> I named it `filter-clear` rather than `clear-filter` so that it was next to the `filter` icon in alphabetical order.


images/icons/filter-clear.svg
acelaya, line 1: I took this icon from Figma https://www.figma.com/design/ctCnzwkAyYArSNBYEshiCm/Hypothesis---LMS-Assignment-Grading?node-id=1615-258&node-type=FRAME&t=V7r4Xxu5lMH7b86Y-0

robertknight (approved):

PR #1700, opened 2024-09-06
Prevent accidental miss-clicks outside multi-select checkboxes

acelaya:
We have experienced (and more people has reported) that it is easy to miss-click outside a checkbox on a `MultiSelect` option. This is problematic, because clicking on a checkbox appends/removes the item to the selection, but clicking outside of it is captured as a click on the main option's body, causing the whole selection to be replaced with just that option.

https://github.com/user-attachments/assets/803908d5-56d5-4ec6-9d7b-ac4f24b3b15d

This PR changes that behavior, so that clicking on "the surroundings" of the checkbox behaves as clicking the checkbox itself.

"The surroundings" in this case means anywhere inside the right side of the option, matching the option's padding. That way the checkbox is always clicked if you click anywhere on the "right" of an option.

This shows how it behaves with these changes:

https://github.com/user-attachments/assets/9c085e7c-813c-431d-bd39-a99a1cf653fc

> [!NOTE]  
> This PR is easier to review [ignoring whitespaces](https://github.com/hypothesis/frontend-shared/pull/1700/files?w=1)

### Todo

- [x] Make checkbox slightly bigger. This was previously done via `scale-125`.

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -72,7 +72,11 @@ function SelectOption<T>({
   elementRef,
 }: SelectOptionProps<T>) {
   const checkboxRef = useRef<HTMLElement | null>(null);
+  const checkboxContainerRef = useRef<HTMLLabelElement | null>(null);
   const optionRef = useSyncedRef(elementRef);
+  const eventTriggeredInCheckbox = (e: Event) =>
+    e.target === checkboxRef.current ||
+    e.target === checkboxContainerRef.current;
 
   const selectContext = useContext(SelectContext);
   if (!selectContext) {
@@ -150,9 +154,12 @@ function SelectOption<T>({
         classes,
       )}
       onClick={e => {
-        // Do not invoke callback if clicked element is the checkbox, as it has
-        // its own event handler.
-        if (!disabled && e.target !== checkboxRef.current) {
+        if (
+          !disabled &&
+          // Do not invoke callback if clicked element is the checkbox or its
+          // container, as it has its own event handler.
+          !eventTriggeredInCheckbox(e)
+        ) {
           selectOneValue();
         }
       }}
@@ -163,9 +170,9 @@ function SelectOption<T>({
 
         if (
           ['Enter', ' '].includes(e.key) &&
-          // Do not invoke callback if event triggers in checkbox, as it has its
-          // own event handler.
-          e.target !== checkboxRef.current
+          // Do not invoke callback if event triggered in the checkbox or its
+          // container, as it has its own event handler.
+          !eventTriggeredInCheckbox(e)
         ) {
           e.preventDefault();
           selectOneValue();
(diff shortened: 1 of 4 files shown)


src/components/input/Select.tsx
```diff
@@ -203,25 +209,32 @@ function SelectOption<T>({
...
-          </div>
+          <Checkbox
+            containerClasses={classnames(
+              // Add negative margin, padding and invisible border to match the
+              // container's, preventing accidental miss-clicks outside the
+              // checkbox that mess the whole selection.
+              '-m-2.5 p-2.5 border border-transparent',
+              // The checkbox is sized based on the container's font size. Make
+              // it a bit larger.
+              'text-xl',
```
acelaya, line 220: The checkbox was previously enlarged via `scale-125`, but that conflicts now with the margin and padding added here.

Increasing the font size was always the right approach.

acelaya:
I just realized this solution does not work if the option has multiple lines, so I'm putting it back in draft while I think in something else.

acelaya:
> I just tested this in the dashboard and realized this solution does not work if the option has multiple lines, so I'm putting it back in draft while I think in something else.

I addressed this by using position absolute and zero top and bottom. That way it will always adapt to any height.

src/components/input/Select.tsx
```diff
@@ -183,46 +190,60 @@ function SelectOption<T>({
...
-              checkedIcon={CheckboxCheckedFilledIcon}
-              elementRef={checkboxRef}
-              onChange={toggleValue}
-              onKeyDown={e => {
-                if (e.key === 'ArrowLeft') {
-                  e.preventDefault();
-                  optionRef.current?.focus();
-                }
-              }}
+          <div className="flex items-center">
```
acelaya, line 208: Since the wrapper is no `items-stretch` instead of `items-center`, every element that is now rendered inside of it needs to define its ow `flex` container with `items-center`, so that everything is visually centered, while the checkbox wrapper (the label) is actually 100% tall, without having to use absolute containers, which have other side effects.

acelaya:
> > I just tested this in the dashboard and realized this solution does not work if the option has multiple lines, so I'm putting it back in draft while I think in something else.
> 
> I addressed this by using position absolute and zero top and bottom. That way it will always adapt to any height.
> 
> EDIT: But that has other side effects I didn't consider:
> 
>  Grabacion.de.pantalla.desde.2024-09-06.14-45-06.webm

I moved back to a non-absolute-position approach, which has less side effects.

Now the option container is `flex items-stretch`, ensuring all items are as high as the taller one. Then every element defines a wrapper of its own with `flex items-center` so that visually the contents are still vertically aligned.

acelaya:
I'm putting this back in draft, as in retrospect, the work I did on this and different attempts I mentioned above, made me realize there might be an easier way to achieve the same.

Since now every "group" inside the option has its own flex container, we could remove the padding from the parent entirely, and then set individual paddings to every group where appropriate.

I'll give that a try and then decide if this is still relevant.

PR #1704, opened 2024-09-09
Prevent accidental miss-clicks outside multi-select checkboxes

acelaya:
> This is an alternative PR to https://github.com/hypothesis/frontend-shared/pull/1700

We have experienced (and more people has reported) that it is easy to miss-click outside a checkbox on a `MultiSelect` option. This is problematic, because clicking on a checkbox appends/removes the item to the selection, but clicking outside of it is captured as a click on the main option's body, causing the whole selection to be replaced with just that option.

https://github.com/user-attachments/assets/803908d5-56d5-4ec6-9d7b-ac4f24b3b15d

This PR changes that behavior, so that clicking on "the surroundings" of the checkbox behaves as clicking the checkbox itself.

"The surroundings" in this case means anywhere inside the right side of the option, matching the option's padding. That way the checkbox is always clicked if you click anywhere on the "right" of an option.

This shows how it behaves with these changes:

https://github.com/user-attachments/assets/9c085e7c-813c-431d-bd39-a99a1cf653fc

> [!NOTE]  
> This PR is easier to review [ignoring whitespaces](https://github.com/hypothesis/frontend-shared/pull/1700/files?w=1)

### DIFF
--- a/src/components/input/Checkbox.tsx
+++ b/src/components/input/Checkbox.tsx
@@ -1,4 +1,4 @@
-import type { JSX } from 'preact';
+import type { JSX, Ref } from 'preact';
 import { useState } from 'preact/hooks';
 
 import type { CompositeProps, IconComponent } from '../../types';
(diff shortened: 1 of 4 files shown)


acelaya:
@marcospri a similar PR, https://github.com/hypothesis/frontend-shared/pull/1700, was already reviewed by Rob.

This is almost the same, but with a slightly less hacky approach, so you can do a high level review. No need to go too deep in the details.

marcospri (approved):
This looks very sensible :+1:

PR #1746, opened 2024-10-21
Prevent Select listbox to grow out of viewport

acelaya:
> Part of https://github.com/hypothesis/product-backlog/issues/1561

This PR update the logic that calculates styles for the Select's listbox when it is displayed, so that it takes into consideration the body width when calculating its position.

When we detect the listbox would render outside of the viewport, we let it "grow" in the oposite direction instead, to prevent horizontal scrolling.

https://github.com/user-attachments/assets/c4be9221-1121-4449-8c2a-b7a105ab6b75

https://github.com/user-attachments/assets/cf2b80fb-3261-444a-b56d-d1cde5c98756

As a side effect, the listbox can now be smaller than its content in some circumstances. For that, we now provide a new prop to decide if the content should be truncated, or wrap multiple lines.

Truncate:

https://github.com/user-attachments/assets/10ac64d5-cd81-4cb2-a673-a9928c7587c2

Wrap: 

https://github.com/user-attachments/assets/164beaf7-3c92-4572-a772-a4893373860b

Additionally, when the options have a more complex content, consumers can decide how to deal with content overflow.

### TODO

- [x] Add tests.
- [x] Make sure `right` listboxes render properly.
- [ ] Dynamically add `title` to items that get cropped.
- [x] Consider using available space on the opposite side when possible.
- [ ] Document `listboxOverflow` prop.
- [ ] Document listbox behavior when content is big

EDIT The missing points there ☝🏼 will be addressed separately to easy reviews.

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -26,7 +26,7 @@ import {
 } from '../icons';
 import Checkbox from './Checkbox';
 import { inputGroupStyles } from './InputGroup';
-import type { SelectValueOptions } from './SelectContext';
+import type { ListboxOverflow, SelectValueOptions } from './SelectContext';
 import SelectContext from './SelectContext';
 
 export type SelectOptionStatus = {
@@ -190,10 +190,7 @@ function SelectOption<T>({
     >
       <div
         className={classnames(
-          // Make items stretch so that all have the same height. This is
-          // important for multi-selects, where the checkbox actionable surface
-          // should span to the very edges of the option containing it.
-          'flex justify-between items-stretch',
+          'flex justify-between items-center',
           'w-full rounded',
           {
             'hover:bg-grey-1 group-focus-visible:ring': !disabled,
@@ -201,11 +198,16 @@ function SelectOption<T>({
           },
         )}
       >
-        <div className="flex items-center py-2 pl-3">
+        <div
+          className={classnames('py-2 pl-3', {
+            truncate: selectContext.listboxOverflow === 'truncate',
+            'whitespace-normal': selectContext.listboxOverflow === 'wrap',
+          })}
+        >
           {optionChildren(children, { selected, disabled })}
         </div>
         {!multiple && (
-          <div className="flex items-center py-2 px-3">
+          <div className="px-3">
             <CheckIcon
               className={classnames('text-grey-6 scale-125', {
                 // Make the icon visible/invisible, instead of conditionally
@@ -219,7 +221,9 @@ function SelectOption<T>({
         {multiple && (
           <Checkbox
             containerClasses={classnames(
-              'flex items-center py-2 px-3',
+              // Make the checkbox stretch, so that its actionable surface spans
+              // to the very edges of the option containing it.
+              'self-stretch px-3',
               // The checkbox is sized based on the container's font size. Make
               // it a bit larger.
               'text-lg',
@@ -250,6 +254,12 @@ SelectOption.displayName = 'Select.Option';
 
 /** Small space to apply between the toggle button and the listbox */
 const LISTBOX_TOGGLE_GAP = '.25rem';
+
+/**
+ * Space in pixels to apply between the listbox and the viewport sides to
+ * prevent the listbox from growing to the very edges.
+ */
+export const LISTBOX_VIEWPORT_HORIZONTAL_GAP = 8;
 
 type ListboxCSSProps =
   | 'top'
@@ -311,31 +321,52 @@ function useListboxPositioning(
       buttonDistanceToBottom < listboxHeight &&
       buttonDistanceToTop > buttonDistanceToBottom;
 
-    if (asPopover) {
-      const { top: bodyTop } = document.body.getBoundingClientRect();
-      const absBodyTop = Math.abs(bodyTop);
+    if (!asPopover) {
+      // Set styles for non-popover mode
+      if (shouldListboxDropUp) {
+        return setListboxCSSProps({
+          bottom: '100%',
+          marginBottom: LISTBOX_TOGGLE_GAP,
+        });
+      }
 
-      return setListboxCSSProps({
-        minWidth: `${buttonWidth}px`,
-        top: shouldListboxDropUp
-          ? `calc(${absBodyTop + buttonDistanceToTop - listboxHeight}px - ${LISTBOX_TOGGLE_GAP})`
-          : `calc(${absBodyTop + buttonDistanceToTop + buttonHeight}px + ${LISTBOX_TOGGLE_GAP})`,
-        left:
-          right && listboxWidth > buttonWidth
-            ? `${buttonLeft - (listboxWidth - buttonWidth)}px`
-            : `${buttonLeft}px`,
-      });
+      return setListboxCSSProps({ top: '100%', marginTop: LISTBOX_TOGGLE_GAP });
     }
 
-    // Set styles for non-popover mode
-    if (shouldListboxDropUp) {
-      return setListboxCSSProps({
-        bottom: '100%',
-        marginBottom: LISTBOX_TOGGLE_GAP,
-      });
+    const { top: bodyTop, width: bodyWidth } =
+      document.body.getBoundingClientRect();
+    const absBodyTop = Math.abs(bodyTop);
+
+    // The available space is:
+    // - left-aligned Selects: distance from left side of toggle button to right
+    //   side of viewport
+    // - right-aligned Selects: distance from right side of toggle button to
+    //   left side of viewport
+    const availableSpace =
+      (right ? buttonLeft + buttonWidth : bodyWidth - buttonLeft) -
+      LISTBOX_VIEWPORT_HORIZONTAL_GAP;
+
+    let left = buttonLeft;
+    if (listboxWidth > availableSpace) {
+      // If the listbox is not going to fit the available space, let it "grow"
+      // in the opposite direction
+      left = right
+        ? LISTBOX_VIEWPORT_HORIZONTAL_GAP
+        : left - (listboxWidth - availableSpace);
+    } else if (right && listboxWidth > buttonWidth) {
+      // If a right-aligned listbox fits the available space, but it's bigger
+      // than the button, move it to the left so that it is aligned with the
+      // right side of the button
+      left -= listboxWidth - buttonWidth;
     }
 
-    return setListboxCSSProps({ top: '100%', marginTop: LISTBOX_TOGGLE_GAP });
+    return setListboxCSSProps({
+      minWidth: `${buttonWidth}px`,
+      top: shouldListboxDropUp
+        ? `calc(${absBodyTop + buttonDistanceToTop - listboxHeight}px - ${LISTBOX_TOGGLE_GAP})`
+        : `calc(${absBodyTop + buttonDistanceToTop + buttonHeight}px + ${LISTBOX_TOGGLE_GAP})`,
+      left: `${Math.max(LISTBOX_VIEWPORT_HORIZONTAL_GAP, left)}px`,
+    });
   }, [asPopover, buttonRef, listboxOpen, listboxRef, right]);
 
   useLayoutEffect(() => {
@@ -405,6 +436,19 @@ type BaseSelectProps = CompositeProps & {
 
   /** A callback passed to the listbox onScroll */
   onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
+
+  /**
+   * Indicates how overflowing content should be handled in the listbox.
+   *
+   * - `truncate`: Truncate the options via `text-overflow: ellipsis`, so that
+   *               they all fit in one line. This is the default value.
+   * - `wrap`: Let options content wrap onto multiple lines via
+   *           `white-space: normal`
+   *
+   * Complex content may still need to provide its own styling to handle content
+   * overflow.
+   */
+  listboxOverflow?: ListboxOverflow;
 };
 
 export type SelectProps<T> = BaseSelectProps & SingleValueProps<T>;
@@ -435,6 +479,7 @@ function SelectMain<T>({
   onListboxScroll,
   right = false,
   multiple,
+  listboxOverflow = 'truncate',
   'aria-label': ariaLabel,
   'aria-labelledby': ariaLabelledBy,
   /* eslint-disable-next-line no-prototype-builtins */
@@ -550,11 +595,15 @@ function SelectMain<T>({
           value: value as typeof multiple extends false ? T : T[],
           selectValue,
           multiple,
+          listboxOverflow,
         }}
       >
         <ul
           className={classnames(
-            'absolute z-5 max-h-80 overflow-y-auto',
+            'absolute z-5 max-h-80 overflow-y-auto overflow-x-hidden',
+            // We don't want the listbox to ever render outside the viewport,
+            // and we give it a 16px gap
+            'max-w-[calc(100%-16px)]',
             'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
             !listboxAsPopover && {
               // Hiding instead of unmounting to
--- a/src/components/input/test/Select-test.js
+++ b/src/components/input/test/Select-test.js
@@ -1,11 +1,15 @@
 import { checkAccessibility, waitFor } from '@hypothesis/frontend-testing';
 import { mount } from 'enzyme';
 
-import { MultiSelect, Select } from '../Select';
+import {
+  LISTBOX_VIEWPORT_HORIZONTAL_GAP,
+  MultiSelect,
+  Select,
+} from '../Select';
 
 describe('Select', () => {
   let wrappers;
-  const items = [
+  const defaultItems = [
     { id: '1', name: 'All students' },
     { id: '2', name: 'Albert Banana' },
     { id: '3', name: 'Bernard California' },
@@ -18,20 +22,27 @@ describe('Select', () => {
    * @param {number} [options.paddingTop] - Extra padding top for the container.
    *                                        Defaults to 0.
    * @param {boolean} [options.optionsChildrenAsCallback] -
-   *        Whether to render Select.Option children with callback notation.
+   *        Whether to render `Select.Option` children with callback notation.
    *        Used primarily to test and cover both branches.
    *        Defaults to true.
    * @param {MultiSelect | Select} [options.Component] -
    *        The actual "select" component to use. Defaults to `Select`.
+   * @param {Array<{ id: string; name: string }>} [options.items] -
+   *        Alternative list of items to render.
    */
   const createComponent = (props = {}, options = {}) => {
     const {
       paddingTop = 0,
       optionsChildrenAsCallback = true,
       Component = Select,
+      items = defaultItems,
     } = options;
     const container = document.createElement('div');
     container.style.paddingTop = `${paddingTop}px`;
+    // Add horizontal paddings to the container, so that there's room for the
+    // listbox to grow if needed
+    container.style.paddingLeft = '200px';
+    container.style.paddingRight = '200px';
     document.body.append(container);
 
     const wrapper = mount(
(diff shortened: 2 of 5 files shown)


src/components/input/Select.tsx
```diff
@@ -190,22 +190,19 @@ function SelectOption<T>({
     >
       <div
         className={classnames(
-          // Make items stretch so that all have the same height. This is
-          // important for multi-selects, where the checkbox actionable surface
-          // should span to the very edges of the option containing it.
-          'flex justify-between items-stretch',
```
acelaya, line 196: Truncating text did not properly work when nesting flex containers, so I simplified this to be a single flex container again, and making the checkbox for multi-selects be `self-stretch`.

This not only solves the problem, but also simplifies the overall layout making it more clear.

src/components/input/Select.tsx
```diff
@@ -311,31 +320,51 @@ function useListboxPositioning(
...
+    if (!asPopover) {
+      // Set styles for non-popover mode
+      if (shouldListboxDropUp) {
+        return setListboxCSSProps({
+          bottom: '100%',
+          marginBottom: LISTBOX_TOGGLE_GAP,
+        });
+      }
+
+      return setListboxCSSProps({ top: '100%', marginTop: LISTBOX_TOGGLE_GAP });
```
acelaya, line 332: The logic for popover listboxes is now more complex, so I decided to flip this condition and move the rest of the logic down.

acelaya:
The `right` prop was originally introduce to let the listbox align with the right side of the toggle button, for when the listbox is located close to the right of the viewport.

image

I'm just now realizing that with the logic introduce here, we could get rid of it entirely, as a listbox would grow to the left "naturally" if it doesn't fit the listbox.

acelaya:
Example of how these changes look when used in the dashboard filters:

https://github.com/user-attachments/assets/bd6ab243-1391-4b23-a065-8520e73b6f6b

src/components/input/Select.tsx
```diff
@@ -250,6 +254,11 @@ SelectOption.displayName = 'Select.Option';
 
 /** Small space to apply between the toggle button and the listbox */
 const LISTBOX_TOGGLE_GAP = '.25rem';
+/**
+ * Space in pixels to apply between the listbox and the viewport sides when the
+ * listbox grows to the very edges
+ */
+const LISTBOX_VIEWPORT_GAP = 8;
```
robertknight, line 261: Is this the horizontal gap, vertical gap, or both?

src/components/input/Select.tsx
```diff
@@ -311,31 +320,50 @@ function useListboxPositioning(
...
-      });
+    const { top: bodyTop, width: bodyWidth } =
+      document.body.getBoundingClientRect();
+    const absBodyTop = Math.abs(bodyTop);
+
+    // Calculate space available for the listbox in regular conditions, to make
+    // sure it never increases the body size, which could cause horizontal
+    // scrollbars to appear
+    const regularAvailableSpace =
+      (right ? buttonLeft + buttonWidth : bodyWidth - buttonLeft) -
```
robertknight, line 343: The meaning of the existing `right` function parameter is as clear as it could be. This ultimately comes from `right` in `BaseSelectProps`. That has some documentation but it could be improved. The summary reads "Align the listbox to the right." To the right of _what_?

src/components/input/Select.tsx
```diff
@@ -311,31 +320,50 @@ function useListboxPositioning(
...
-    if (shouldListboxDropUp) {
-      return setListboxCSSProps({
-        bottom: '100%',
-        marginBottom: LISTBOX_TOGGLE_GAP,
-      });
+    const { top: bodyTop, width: bodyWidth } =
+      document.body.getBoundingClientRect();
+    const absBodyTop = Math.abs(bodyTop);
+
+    // Calculate space available for the listbox in regular conditions, to make
```
robertknight, line 339: What is meant by "regular" here? That word seems unnecessary in the comment and variable name.

src/components/input/Select.tsx
```diff
@@ -311,31 +320,50 @@ function useListboxPositioning(
...
+      (right ? buttonLeft + buttonWidth : bodyWidth - buttonLeft) -
+      LISTBOX_VIEWPORT_GAP;
+
+    let left = buttonLeft;
+    if (listboxWidth > regularAvailableSpace) {
+      // If the listbox is not going to fit the available space, let it "grow"
+      // in the opposite direction
+      left = right
+        ? LISTBOX_VIEWPORT_GAP
+        : left - (listboxWidth - regularAvailableSpace);
```
robertknight, line 352:
```suggestion
        : left - listboxWidth - regularAvailableSpace;
```

src/components/input/Select.tsx
```diff
@@ -405,6 +433,19 @@ type BaseSelectProps = CompositeProps & {
 
   /** A callback passed to the listbox onScroll */
   onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
+
+  /**
+   * Indicates how overflowing content should be handled in the listbox.
+   * * `truncate` -> Truncate the options so that they all fit in one line. An
```
robertknight, line 439:
```suggestion
   * Indicates how overflowing content should be handled in the listbox.
   *
   * * `truncate` -> Truncate the options so that they all fit in one line. An
```

There should be a blank line between the one-line summary and detailed description.

src/components/input/Select.tsx
```diff
@@ -405,6 +433,19 @@ type BaseSelectProps = CompositeProps & {
...
+   *                 ellipsis will be rendered where needed.
+   *                 This is the default value.
+   * * `wrap` -> Let options content wrap multiple lines so that all content is
+   *             visible.
+   *
+   * This behavior can be also overwritten by providing more complex options
+   * content.
```
robertknight, line 446: I don't find the meaning of this very clear. I presume what you mean to say is that when the rendered content meets certain criteria, the `wrap` and `truncate` styles applied to the listbox won't have an effect and the caller will have to provide their own styles to control this.

One way to resolve this would be to document the behavior explicitly as setting the `text-overflow` and `white-space` CSS properties and noting as a caveat that some content may need to provide its own styling to handle cases where the content is not wide enough.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +432,83 @@ describe('Select', () => {
     });
   });
 
+  context('when listbox does not fit in available space', () => {
+    async function getOpenListbox(wrapper) {
+      toggleListbox(wrapper);
+      // Wait for listbox to be open
```
robertknight, line 438: This comment feels redundant to me.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +432,83 @@ describe('Select', () => {
...
+        const listboxStyleLeft = listbox.getDOMNode().style.left;
+        const listboxLeft = Number(listboxStyleLeft.replace('px', ''));
+        const { left: buttonLeft } = getToggleButton(wrapper)
+          .getDOMNode()
+          .getBoundingClientRect();
+
+        assert.equal(listboxLeft < buttonLeft, shouldGrow);
+      });
+    });
+    [
```
robertknight, line 483: New line needed between tests here.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +432,83 @@ describe('Select', () => {
...
+      { name: 'long name'.repeat(500), shouldGrow: true },
+      { name: 'short name', shouldGrow: false },
+    ].forEach(({ name, shouldGrow }) => {
+      it('makes right Select listbox grow to the right if there is space', async () => {
```
robertknight, line 487:
```suggestion
      it('makes right-aligned Select listbox grow to the right if there is space', async () => {
```

src/components/input/test/Select-test.js
```diff
@@ -425,12 +432,83 @@ describe('Select', () => {
...
+          {
+            items: ['1', '2', '3'].map(id => ({ id, name })),
+          },
+        );
+
+        const listbox = await getOpenListbox(wrapper);
+        const listboxStyleLeft = listbox.getDOMNode().style.left;
+        const listboxLeft = Number(listboxStyleLeft.replace('px', ''));
+
+        // 8 is the minimum `left` value that acts as a gap with the edge of the
```
robertknight, line 499: This value could be exported  from `Select.tsx` and then imported here.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +432,83 @@ describe('Select', () => {
...
+        // viewport
+        assert.equal(listboxLeft === 8, shouldGrow);
```
robertknight, line 501: This might be better written as two separate asserts, called according to `shouldGrow`. One branch would test `assert.equal(listboxLeft, 8)` the other a not-equal version of that. This will result in a more helpful error if the test fails.

src/pattern-library/examples/select-right.tsx
```diff
@@ -3,11 +3,11 @@ import { useId, useState } from 'preact/hooks';
 import { Select } from '../..';
 
 const items = [
-  { id: '1', name: 'All students' },
-  { id: '2', name: 'Albert Banana' },
-  { id: '3', name: 'Bernard California' },
-  { id: '4', name: 'Cecelia Davenport' },
-  { id: '5', name: 'Doris Evanescence' },
+  { id: '1', name: 'All students (this item has a very long text)' },
```
robertknight, line 6:
```suggestion
  { id: '1', name: 'All students (this item has very long text)' },
```

In this context _text_ is an uncountable noun, so no "a".

robertknight:
> I'm just now realizing that with the logic introduce here, we could get rid of it entirely, as a listbox would grow to the left "naturally" if it doesn't fit the viewport.

I think I would be inclined to keep it. The UI looks tidier if dropdowns on the right edge of the screen have their right edges exactly aligned with the toggle button, when there is space.

src/components/input/Select.tsx
```diff
@@ -219,7 +221,9 @@ function SelectOption<T>({
         {multiple && (
           <Checkbox
             containerClasses={classnames(
-              'flex items-center py-2 px-3',
+              // Make the checkbox stretch, so that its actionable surface spans
+              // to the very edges of the option containing it.
+              'self-stretch px-3',
```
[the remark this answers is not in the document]
acelaya: Yes.

src/components/input/Select.tsx
acelaya: horizontal only. The listbox have a max vertical height after which it has it's own scroll.

In the rare occasions that doesn't fit the viewport height, the main scroll would allow you to see all items, which is not perfect, but has worked well so far.

src/components/input/Select.tsx
acelaya: I'm not sure if I follow. Are you suggesting to update the comment somewhere?

src/components/input/Select.tsx
acelaya: With "regular condition" I mean the space assuming the listbox fits the viewport.

* For left-aligned Selects, that's the space from the left side of the button to the right side of the viewport.
* For right-aligned Selects, the space from the left side of the viewport, to the right side of the button.

I'll try to reword this.

src/components/input/Select.tsx
acelaya: Actually, the parentheses are needed here, because they change the priority:

* `100 - (30 - 10)` -> `100 - 20` -> `80`
* `100 - 30 - 10` -> `70 - 10` -> `60`

image

src/components/input/Select.tsx
acelaya: > I don't find the meaning of this very clear. I presume what you mean to say is that when the rendered content meets certain criteria, the `wrap` and `truncate` styles applied to the listbox won't have an effect and the caller will have to provide their own styles to control this.

Yes, that's correct.

> One way to resolve this would be to document the behavior explicitly as setting the `text-overflow` and `white-space` CSS properties and noting as a caveat that some content may need to provide its own styling to handle cases where the content is not wide enough.

Ok!

src/components/input/Select.tsx
robertknight: 🤦 indeed.

acelaya:
> > I'm just now realizing that with the logic introduce here, we could get rid of it entirely, as a listbox would grow to the left "naturally" if it doesn't fit the viewport.
> 
> I think I would be inclined to keep it. The UI looks tidier if dropdowns on the right edge of the screen have their right edges exactly aligned with the toggle button, when there is space.

Yeah, I tend to agree

src/components/input/Select.tsx
acelaya: I have renamed this to `availableSpace` and changed the comment to simply explain what is the available space.

Later on, when this value is used, there are other comments which explain what happens when the listbox does not fit the `availableSpace`.

src/components/input/Select.tsx
acelaya: I have renamed this to `LISTBOX_VIEWPORT_HORIZONTAL_GAP`.

src/components/input/test/Select-test.js
acelaya: I refactored this test to test for the exact left and right coordinates of the listbox in three different cases:

* Small content: the listbox matches the button size due to the `min-width` CSS prop.
* Medium content: the listbox is a bit bigger than the button and grows to the right.
* Big content: the listbox grows to the very right side (considering the gap) and then grows a bit further to the right than the toggle button.

src/components/input/Select.tsx
```diff
@@ -251,6 +255,12 @@ SelectOption.displayName = 'Select.Option';
 /** Small space to apply between the toggle button and the listbox */
 const LISTBOX_TOGGLE_GAP = '.25rem';
 
+/**
+ * Space in pixels to apply between the listbox and the viewport sides to
+ * prevent the listbox to grow to the very edges.
```
robertknight, line 260:
```suggestion
 * prevent the listbox from growing to the very edges.
```

src/components/input/Select.tsx
```diff
@@ -311,31 +321,52 @@ function useListboxPositioning(
...
+
+    let left = buttonLeft;
+    if (listboxWidth > availableSpace) {
+      // If the listbox is not going to fit the available space, let it "grow"
+      // in the opposite direction
+      left = right
+        ? LISTBOX_VIEWPORT_HORIZONTAL_GAP
+        : left - (listboxWidth - availableSpace);
+    } else if (right && listboxWidth > buttonWidth) {
+      // If a `right` listbox fits the available space, but it's bigger than the
```
robertknight, line 357:
```suggestion
      // If a right-aligned listbox fits the available space, but it's bigger than the
```

src/components/input/Select.tsx
```diff
@@ -405,6 +436,19 @@ type BaseSelectProps = CompositeProps & {
 
   /** A callback passed to the listbox onScroll */
   onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
+
+  /**
+   * Indicates how overflowing content should be handled in the listbox.
+   *
+   * - `truncate`: Truncate the options via `text-overflow: ellipsis`, so that
+   *               they all fit in one line.
+   *               This is the default value.
```
robertknight, line 445: This sentence can be moved to the previous line.

src/components/input/Select.tsx
```diff
@@ -405,6 +436,19 @@ type BaseSelectProps = CompositeProps & {
...
+   * - `wrap`: Let options content wrap multiple lines via `white-space: normal`
```
robertknight, line 446:
```suggestion
   * - `wrap`: Let options content wrap onto multiple lines via `white-space: normal`
```

src/components/input/test/Select-test.js
```diff
@@ -425,12 +436,137 @@ describe('Select', () => {
...
+        const wrapper = createComponent(
+          { buttonContent: 'Select a value' },
+          {
+            items: ['1', '2', '3'].map(id => ({ id, name })),
+          },
+        );
+
+        const listbox = await getOpenListbox(wrapper);
+        const listboxStyleLeft = listbox.getDOMNode().style.left;
+        const listboxLeft = Number(listboxStyleLeft.replace('px', ''));
```
robertknight, line 478: This is fine. You can use `parseInt` on a string like "42px" or `parseFloat` if the value may be a float, and get the expected result. This is because these functions stop parsing once they encounter a character that is not a valid digit.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +436,137 @@ describe('Select', () => {
...
+      const listbox = await getOpenListbox(wrapper);
+      const { width: listboxWidth } = listbox
+        .getDOMNode()
+        .getBoundingClientRect();
+
+      assert.isTrue(listboxWidth < window.window.innerWidth);
+    });
+
+    [
+      { name: 'long name'.repeat(50), shouldGrow: true },
```
robertknight, line 465: This string is about 2300px wide on my system (using `CanvasRenderingContext2D.measureText`). If the viewport happens to be really large in the test browser, it is possible the listbox might not need to grow.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +436,137 @@ describe('Select', () => {
...
+        );
+
+        const listbox = await getOpenListbox(wrapper);
+        const listboxStyleLeft = listbox.getDOMNode().style.left;
+        const listboxLeft = Number(listboxStyleLeft.replace('px', ''));
+        const { left: buttonLeft } = getToggleButton(wrapper)
+          .getDOMNode()
+          .getBoundingClientRect();
+
+        assert.equal(listboxLeft < buttonLeft, shouldGrow);
```
robertknight, line 483: I would suggest adding the exact values of these into the assertion error. One way is to call either `assert.isBelow` if `shouldGrow` or `assert.isAtLeast` otherwise. The values can also be added in an assertion message.

src/components/input/test/Select-test.js
```diff
@@ -425,12 +436,137 @@ describe('Select', () => {
...
+        name: 'very long name'.repeat(6),
+        getExpectedCoordinates: (wrapper, listboxDOMNode) => {
+          const listboxRect = listboxDOMNode.getBoundingClientRect();
+          return {
+            left: LISTBOX_VIEWPORT_HORIZONTAL_GAP,
+            right: listboxRect.width + LISTBOX_VIEWPORT_HORIZONTAL_GAP,
+          };
+        },
+      },
+    ].forEach(({ name, getExpectedCoordinates }) => {
```
robertknight, line 531: It seems odd to me that there are an asymmetric set of tests for left and right-aligned listboxes. I would expect the behavior to be the same, but mirrored.

robertknight (commented):
I had a query about the asymmetry in tests for left and right-aligned list boxes.

src/components/input/test/Select-test.js
acelaya: Yeah, good point. They kind-of were before the last change I pushed.

I'll apply the same checks to the left-aligned Select tests, as they cover more cases.

src/components/input/test/Select-test.js
acelaya: I have consolidated this test with the one testing all `right: true` cases, so now they all check for exact values.

acelaya:
> I had a query about the asymmetry in tests for left and right-aligned list boxes.

I have consolidated the tests for left-aligned and right-aligned listboxes in a single test case with a dataset that covers 6 scenarios:

1. Left-aligned listbox with small content: the listbox has the same size as the button.
2. Right-aligned listbox with small content: the listbox has the same size as the button.
3. Left-aligned listbox with slightly bigger content: The listbox matches the toggle button's left side and grows further to the right.
4. Right-aligned listbox with slightly bigger content: The listbox matches the toggle button's right side and grows further to the left.
5. Left-aligned listbox with much bigger content: The listbox matches the body's right side (minus the gap) and grows to the left further than the button's side.
6. Right-aligned listbox with much bigger content: The listbox matches the body's left side (minus the gap) and grows to the right further than the button's side.

robertknight (approved):

PR #1754, opened 2024-10-28
Document the `Select` component `listboxOverflow`

acelaya:
The `listboxOverflow` prop was added in https://github.com/hypothesis/frontend-shared/pull/1746. This PR documents it and adds examples.

https://github.com/user-attachments/assets/a47e450b-3f37-46a8-afd6-f911919bf830

### DIFF
--- a/src/pattern-library/components/patterns/prototype/SelectPage.tsx
+++ b/src/pattern-library/components/patterns/prototype/SelectPage.tsx
@@ -387,24 +387,6 @@ export default function SelectPage() {
               </div>
             </Library.Demo>
           </Library.Example>
-          <Library.Example title="listboxClasses">
-            <Library.Info>
-              <Library.InfoItem label="description">
-                Additional classes to pass to listbox.
-              </Library.InfoItem>
-              <Library.InfoItem label="type">
-                <code>string | string[]</code>
-              </Library.InfoItem>
-              <Library.InfoItem label="default">
-                <code>undefined</code>
-              </Library.InfoItem>
-            </Library.Info>
-            <Library.Demo title="Custom listbox">
-              <div className="w-96 mx-auto">
-                <SelectExample listboxClasses="border-4 border-yellow-notice" />
-              </div>
-            </Library.Demo>
-          </Library.Example>
           <Library.Example title="listboxAsPopover">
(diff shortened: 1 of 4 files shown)


src/pattern-library/components/patterns/prototype/SelectPage.tsx
```diff
@@ -447,6 +447,40 @@ export default function SelectPage() {
               </Library.InfoItem>
             </Library.Info>
           </Library.Example>
+          <Library.Example title="listboxOverflow">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Determines the behavior of the listbox options when their
```
robertknight, line 453:
```suggestion
                Determines the behavior of the listbox options when their
```

"options" or "items"? The component is called `Select.Option` but code examples tend to use the term "items" for the data arrays from which the options are generated.

robertknight (approved):

src/pattern-library/components/patterns/prototype/SelectPage.tsx
```diff
@@ -447,6 +447,40 @@ export default function SelectPage() {
               </Library.InfoItem>
             </Library.Info>
           </Library.Example>
+          <Library.Example title="listboxOverflow">
```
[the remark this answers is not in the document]
acelaya: > The most straightforward thing to do would be to order all props alphabetically

Right! This is how we have usually agreed to do it, but I tend to forget it 🤦🏼

src/pattern-library/components/patterns/prototype/SelectPage.tsx
acelaya: I think I would use "items" when conceptually talking about the entries of the list, and "options" when talking about the markup elements explicitly.

In this case, since we are talking about how they render, I think I would use "options".

But this is not a strong opinion, and I see a lot of grey areas in that reasoning, so I'm ok using a different wording.

robertknight: > I think I would use "items" when conceptually talking about the entries of the list, and "options" when talking about the markup elements explicitly.

That makes sense.

PR #1756, opened 2024-10-29
Add new `alignListbox` prop to `Select` to replace `right` one

acelaya:
This PR deprecates the `right: true | false` prop in Selects, and replaces it with `alignListbox: 'left' | 'right'`.

The behavior does not change, but the new one is easier to understand and reason about.

The `right` is not removed for backwards compatibility. If it is provided and `alignListbox` is not, the behavior will be the same as providing `alignListbox: 'right'`.

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -280,7 +280,7 @@ function useListboxPositioning(
   listboxRef: RefObject<HTMLElement | null>,
   listboxOpen: boolean,
   asPopover: boolean,
-  right: boolean,
+  alignListboxToRight: boolean,
 ) {
   const adjustListboxPositioning = useCallback(() => {
     const listboxEl = listboxRef.current;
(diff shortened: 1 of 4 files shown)


src/components/input/Select.tsx
```diff
@@ -478,6 +484,7 @@ function SelectMain<T>({
   containerClasses,
   onListboxScroll,
   right = false,
+  alignListbox = right ? 'right' : 'left',
```
acelaya, line 487: For now we define the default value of `alignListbox` based on `right`, but as soon as we remove `right` in the next major version, we can change this to `alignListbox = 'left'`.

src/components/input/Select.tsx
```diff
@@ -417,13 +418,18 @@ type BaseSelectProps = CompositeProps & {
...
+
   /**
-   * Align the listbox to the right.
-   * Useful when the listbox is bigger than the toggle button and this component
-   * is rendered next to the right side of the page/container.
-   * Defaults to false.
+   * How to align the listbox relative to the toggle button.
+   *
+   * Useful when rendering a Select with a big listbox, close to the right side
+   * of the viewport. In that case it looks nicer if the listbox matches the
```
robertknight, line 428:
```suggestion
   * of the viewport. In that case it looks nicer if the listbox aligns with the
```

src/components/input/Select.tsx
```diff
@@ -417,13 +418,18 @@ type BaseSelectProps = CompositeProps & {
...
+   * right side of the toggle button and spans to the left.
```
robertknight, line 429:
```suggestion
   * right side of the toggle button and extends to the left.
```

robertknight (approved):
LGTM. I note that capitalization of "listbox" varies across UI libraries. The choice here aligns with eg. WAI documentation (https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)

PR #1759, opened 2024-10-31
Export useSelectContext for downstream projects to use

acelaya:
This PR adds a new `useSelectContext()` hook, which is an alias for `useContext(SelectContext)`.

This hook is exposed from this library's entry point, and can be used by consumers which define their own custom `Select.Option` wrappers and want to dynamically react to the select status.

Additionally, the context now exposes a `listboxOpen` flag, which is `true` while the listbox is opened. This can be used to add behavior that is relevant only while the listbox is open, like event listeners or observers.

image

A good use case for this hook is the logic added in https://github.com/hypothesis/lms/pull/6812, where the options sizes are computed to dynamically add a `title` attribute if they are truncated.

### TODO

- [x] Reference hook in documentation

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -2,7 +2,6 @@ import classnames from 'classnames';
 import type { ComponentChildren, JSX, RefObject, Ref } from 'preact';
 import {
   useCallback,
-  useContext,
   useId,
   useLayoutEffect,
   useMemo,
@@ -27,6 +26,7 @@ import {
 import Checkbox from './Checkbox';
 import { inputGroupStyles } from './InputGroup';
 import type { ListboxOverflow, SelectValueOptions } from './SelectContext';
+import { useSelectContext } from './SelectContext';
 import SelectContext from './SelectContext';
 
 export type SelectOptionStatus = {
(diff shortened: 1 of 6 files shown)


robertknight:
What is the status of this following the revised approach taken in https://github.com/hypothesis/lms/pull/6819?

robertknight (commented):
I had a query about whether this is still needed given the downstream approaches in the LMS. If it is still useful in future, we might want to think about whether we expose the full context type that is used internally, which is what currently happens, or just a restricted subset.

acelaya:
> I had a query about whether this is still needed given the downstream approaches in the LMS. If it is still useful in future, we might want to think about whether we expose the full context type that is used internally, which is what currently happens, or just a restricted subset which is easier to ensure future support for.

I guess this would still be useful, considering the approach that was finally taken, as something like this would allow to attach the observers only while the listbox is open.

However, we discussed a different approach last week for the listbox as a whole, in which we only mount the listbox children while the listbox is open. I did a quick test and other than a few failing tests, it seemed to work.

That would solve the same problem implicitly, so I think I'm going to close this and explore that other solution instead.

PR #1762, opened 2024-11-04
Do not render Select listbox children while it's closed

acelaya:
This PR improves a little bit the Select's resources usage, by making sure nothing is rendered inside the listbox while it is closed.

The listbox itself needs to be kept in the DOM, for proper size calculations, and because the use of the Popover API requires a reference to be available.

However, all of its children can be safely unmounted and mounted again when opened, allowing them to have their own hooks and even handlers that are released as long as the listbox is closed.

The changes in the `Select` itself are minimal, but a few tests needed to be updated, as they were interacting with options while the listbox was closed, which doesn't work anymore.

### Testing steps

Just a bit of sanity check in http://localhost:4001/input-select to verify everything keeps working as before.

### Considerations

This change makes it a bit harder to debug issues with the Select, as the children are removed as soon as you try to inspect the DOM.

This was somewhat a problem already, as the listbox is closed when the focus is removed from it, so debugging was already a multi-step process.

I may add later something in the lines of `debug={true}` that disables a few behaviors in the Select and allows for an easier debugging in downstream projects if needed.

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -545,14 +545,6 @@ function SelectMain<T>({
     selector: '[role="option"]:not([aria-disabled="true"])',
   });
 
-  useLayoutEffect(() => {
-    // Focus toggle button after closing listbox, only if previously focused
-    // element was inside the listbox itself
-    if (!listboxOpen && listboxRef.current!.contains(document.activeElement)) {
-      buttonRef.current!.focus();
-    }
-  }, [buttonRef, listboxOpen]);
-
   return (
     <div
       className={classnames(
@@ -640,7 +632,7 @@ function SelectMain<T>({
           popover={listboxAsPopover ? 'auto' : undefined}
           onScroll={onListboxScroll}
         >
-          {children}
+          {listboxOpen && children}
         </ul>
       </SelectContext.Provider>
     </div>
--- a/src/components/input/test/Select-test.js
+++ b/src/components/input/test/Select-test.js
@@ -101,6 +101,12 @@ describe('Select', () => {
   const isListboxClosed = wrapper =>
     getListbox(wrapper).prop('data-listbox-open') === false;
 
+  const openListbox = wrapper => {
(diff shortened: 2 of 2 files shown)


src/components/input/Select.tsx
```diff
@@ -545,14 +545,6 @@ function SelectMain<T>({
     selector: '[role="option"]:not([aria-disabled="true"])',
   });
 
-  useLayoutEffect(() => {
-    // Focus toggle button after closing listbox, only if previously focused
-    // element was inside the listbox itself
-    if (!listboxOpen && listboxRef.current!.contains(document.activeElement)) {
-      buttonRef.current!.focus();
-    }
-  }, [buttonRef, listboxOpen]);
```
acelaya, line 554: We no longer need to manually return the focus to the button once the listbox is closed, because since the listbox content is removed, the focus is automatically returned to the button, which was the last focused button.

robertknight:
> This change makes it a bit harder to debug issues with the Select, as the children are removed as soon as you try to inspect the DOM.

Chrome has an "Emulate focused page" feature in devtools specifically to help with this: https://developer.chrome.com/docs/devtools/rendering/apply-effects#:~:text=highlighted%20in%20red.-,Emulate%20a%20focused%20page,clear%20Emulate%20a%20focused%20page..

robertknight (approved):

acelaya:
> > This change makes it a bit harder to debug issues with the Select, as the children are removed as soon as you try to inspect the DOM.
> 
> Chrome has an "[Emulate focused page](https://developer.chrome.com/docs/devtools/rendering/apply-effects#emulate_a_focused_page)" feature in devtools specifically to help with this.

image

Just tested it in http://localhost:4001/input-select. It works great!

acelaya:
I have tested this change in client and LMS, including cases where dashboard filters loading fails and needs to be retried, and when multiple pages need to be loaded.

All works as expected, so I'll merge this.

PR #1767, opened 2024-11-05
Make Select button main content grow

acelaya:
This PR makes the content provided as `buttonContent` to a `Select` or `MultiSelect` component take all space available expect for the chevron.

This allows to pass a flex container to `buttonContent` and allow to place items at the end, right before the chevron icon.

Before:

https://github.com/user-attachments/assets/be399d66-bf06-4151-ad82-a2843903cece

After:

https://github.com/user-attachments/assets/6157129e-86f2-4749-935d-95dc9931ace2

The first use case for this is displaying a badge with the amount of items in the listbox, in the LMS dashboard filters.

Without this, we can only display the badge right after the text, and this would allow to "push" this badge to the end, right before the chevron.

It may be possible to work around this limitation applying other CSS properties in consuming code, but I haven't found one which works consistently and doesn't feel a bit hackier than this.


robertknight (approved):
Seems reasonable to me.

PR #1768, opened 2024-11-05
Make sure selected option is focused when listbox is opened

acelaya:
This PR fixes an unintended regression introduced by making [options render only when the listbox is open](https://github.com/hypothesis/frontend-shared/pull/1762).

We now make sure the `tabIndex` is `0` for the selected option, which will make [`useArrowKeyNavigation` focus it the first time](https://github.com/hypothesis/frontend-shared/blob/986c9d5a7fdc1a269e1a4bcf26ba283f7c2d2bb6/src/hooks/use-arrow-key-navigation.ts#L162).

This worked before because the options were kept in the DOM, preserving their previously set `tabIndex`, but now we render them "from scratch" every time it is opened.

Before:

https://github.com/user-attachments/assets/0dc715e7-1064-4fbd-8423-b1fa31da0ccd

After:

https://github.com/user-attachments/assets/ee45dfcc-19e9-460e-b3d6-63f78f1d7896

### Testing steps

1. Check out this branch.
2. Go to http://localhost:4001/input-select
3. Using the keyboard, focus the first Select, then click <kbd>ArrowDown</kbd> to open the listbox. The first option should be focused.
4. Using arrows, move the focus to a different option, and select it by pressing <kbd>Enter</kbd>. The focus will return to the button.
5. Press <kbd>ArrowDown</kbd> again to open the listbox. The focused option should be the selected one.

If you follow these steps in `main` branch, you'll notice the first option is always focused when the listbox is opened, regardless of which one is selected.

### DIFF
--- a/src/components/input/test/Select-test.js
+++ b/src/components/input/test/Select-test.js
@@ -118,18 +118,18 @@ describe('Select', () => {
     return listboxTop < buttonTop;
   };
 
+  const getOption = (wrapper, id) =>
+    wrapper.find(`[data-testid="option-${id}"]`).closest('[role="option"]');
+
   const clickOption = (wrapper, id) => {
     openListbox(wrapper);
-    wrapper.find(`[data-testid="option-${id}"]`).simulate('click');
+    getOption(wrapper, id).simulate('click');
   };
 
   function clickOptionCheckbox(wrapper, id) {
     openListbox(wrapper);
 
-    const checkbox = wrapper
-      .find(`[data-testid="option-${id}"]`)
-      .closest('[role="option"]')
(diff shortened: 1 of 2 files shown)


acelaya:
I forgot to add a test. Moving back to draft

src/components/input/test/Select-test.js
```diff
@@ -391,6 +382,19 @@ describe('Select', () => {
     });
   });
 
+  it('sets tabIndex=0 to selected option', () => {
```
robertknight, line 385:
```suggestion
  it('sets tabIndex=0 for the selected option', () => {
```

robertknight (approved):

PR #1775, opened 2024-11-11
Add new globe-with-lock icon

acelaya:
> Depends on https://github.com/hypothesis/frontend-shared/pull/1774

Add a new globe-with-lock to use with restricted groups.

image


robertknight:
There is an existing `GlobeAltIcon` which looks very similar. What is the status of that?

acelaya:
> There is an existing `GlobeAltIcon` which looks very similar. What is the status of that?

Oh right! I simply overlooked that. We can keep it, and I'll just name the lock-one GlobeAltLockIcon.

robertknight (approved):

PR #1777, opened 2024-11-13
Move Select's toggle button padding to its children elements

acelaya:
This PR changes how padding is handled in the `Select`'s toggle button, so that instead of having a surrounding padding in the button itself, we replace it with the corresponding padding to its two children.

This is important because the first children truncates its content via tailwind's `truncate`. This class, among other things, adds `overflow: hidden`, but we want to allow content to overflow as long as it doesn't grow outside of the button limits.

Before:

https://github.com/user-attachments/assets/3130234d-31fd-4052-a296-26613e620d15

After:

https://github.com/user-attachments/assets/ee57a952-be51-4e0f-aecd-86766761b2af

Visually, there should be no changes, but the element with the `truncate` class now spans to the button limits, making its behavior more intuitive.

> [!NOTE]
> The motivation for this change is described in this thread https://github.com/hypothesis/lms/pull/6820#discussion_r1839884418

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -560,7 +560,7 @@ function SelectMain<T>({
         id={buttonId ?? defaultButtonId}
         className={classnames(
           'focus-visible-ring transition-colors whitespace-nowrap',
-          'w-full flex items-center justify-between gap-x-2 p-2',
+          'w-full flex items-center justify-between gap-x-2',
           'bg-grey-0 disabled:bg-grey-1 disabled:text-grey-6',
           // Buttons are center-aligned by default. Overwrite it.
           'text-left',
@@ -589,8 +589,8 @@ function SelectMain<T>({
         }}
         data-testid="select-toggle-button"
       >
-        <div className="truncate grow">{buttonContent}</div>
-        <div className="text-grey-6">
+        <div className="pl-2 py-2 truncate grow">{buttonContent}</div>
+        <div className="pr-2 py-2 text-grey-6">
           {listboxOpen ? <MenuCollapseIcon /> : <MenuExpandIcon />}
         </div>
       </button>


robertknight:
> This is important because the first children truncates its content via tailwind's truncate. This class, among other things, adds overflow: hidden, but we want to allow content to overflow as long as it doesn't grow outside of the button limits.

I'm not clear on what you mean. The first child still has `truncate` applied to it, which means that overflow is still hidden. Do you mean that content which previously overflowed into the button's padding region is now instead part of the content's padding region, and thus not overflowing?

Can you clarify how this helps in the context of https://github.com/hypothesis/lms/pull/6820?

acelaya:
> > This is important because the first children truncates its content via tailwind's truncate. This class, among other things, adds overflow: hidden, but we want to allow content to overflow as long as it doesn't grow outside of the button limits.
> 
> I'm not clear on what you mean. The first child still has `truncate` applied to it, which means that overflow is still hidden. Do you mean that content which previously overflowed into the button's padding region is now instead part of the content's padding region, and thus not overflowing?
> 
> Can you clarify how this helps in the context of [hypothesis/lms#6820](https://github.com/hypothesis/lms/pull/6820)?

So, without applying any extra styles to the Select's button, this is how it looks when we add the count badge

image

It grows a bit more than when it only has "plain text", which a) makes it a little bit clunkier, and b) can make selects have inconsistent heights depending on the content of the button, as the badge is displayed only sometimes.

Additionally, if some selects display the badge and some not, some styles break (which we could address separately, but we still have point a).

image

So, in order to compensate for the extra height added by the badge and ensure Selects look closer to the designs, I tried to apply negative margins to the badge, but then this happens:

image

The badge gets cut, because the overflow is hidden at this element:

image

So, what this PR does is moving the padding to the same element with `truncate`, so that we still apply `overflow: hidden` and `text-overflow: ellipsis`, which is correct, but we allow overflowing vertically over the padded area, as long as it doesn't grow outside the button borders.

That's what I would intuitively expect.

image

Of course, we could argue if this is in fact the right problem to solve. Maybe we should not apply `truncate` at all, or allow it to be conditionally applied, but that has other side effects.

Or maybe the negative vertical margin on the badge is the hack that's driving everything in the wrong direction, but I also don't see a way to avoid that without altering the button's layout.

I have tried a couple different combinations, but with all the scenarios we support with `Select`, this was the less disruptive and less prone to introduce a regression somewhere else.

robertknight (approved):
OK, thanks for the explanation. That makes sense.

PR #1789, opened 2024-11-19
Split Select dual popover/listbox element in two

acelaya:
This PR is the first of a series aiming to simplify/extract capabilities from the `Select` component.

Here we are splitting the listbox-as-popover element into two independent elements:

1. A `div` which acts as popover, using the popover API when possible, and handling positioning and openness.
2. A `ul` with the listbox role, which handles accessibility aspects to handle a list of items, like aria attributes and keyboard navigation. It is the only children of the popover element.

For now, this split is done inside of the `Select` component, keeping it unchanged from a consumer or user point of view.

In follow-up PRs we'll we extracting some/all of those components.


src/components/input/Select.tsx
```diff
@@ -275,23 +275,27 @@ type ListboxCSSProps =
...
  * toggle button (below or over). This is mainly needed when the listbox is used
  * as a popover, as that makes it render in the top layer, making it impossible
  * to position it relative to the toggle button via regular CSS.
+ *
+ * @param asNativePopover - Native popover API is used to toggle the popover
+ * @param alignToRight - Whether the popover should be aligned to the right side
+ *                       of the button or not
  */
-function useListboxPositioning(
+function usePopoverPositioning(
```
acelaya, line 287: I removed all references to "listbox" from this hook, because it is likely going to be extracted and decoupled from that.

src/components/input/test/Select-test.js
```diff
@@ -461,22 +461,15 @@ describe('Select', () => {
         const { left: buttonLeft } = getToggleButton(wrapper)
           .getDOMNode()
           .getBoundingClientRect();
-        const listboxLeft = getListboxLeft(wrapper);
+        const popoverLeft = getPopoverLeft(wrapper);
 
-        assert.isTrue(listboxLeft < buttonLeft);
+        assert.isAtMost(popoverLeft, buttonLeft);
```
acelaya, line 466: This should be `assert.isBelow`, but in the `listboxAsPopover: false` case, it is not properly aligning the element with the right side of the button.

I'm investigating why, because on manual testing this works as expected.

src/components/input/Select.tsx
```diff
@@ -257,16 +257,16 @@ function SelectOption<T>({
 
 SelectOption.displayName = 'Select.Option';
 
-/** Small space to apply between the toggle button and the listbox */
```
acelaya, line 260: This PR is full of `listbox`/`popover` replacements.

src/components/input/Select.tsx
```diff
@@ -604,13 +607,17 @@ function SelectMain<T>({
...
             'absolute z-5 max-h-80 overflow-y-auto overflow-x-hidden',
-            // We don't want the listbox to ever render outside the viewport,
-            // and we give it a 16px gap
-            'max-w-[calc(100%-16px)]',
             'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
+            listboxAsPopover && [
+              // We don't want the listbox to ever render outside the viewport,
+              // and we give it a 16px gap
+              'max-w-[calc(100%-16px)]',
+              // Overwrite [popover] default styles
```
robertknight, line 618: What are the `popover` default styles?

src/components/input/Select.tsx
```diff
@@ -621,22 +628,25 @@ function SelectMain<T>({
...
-          id={listboxId}
-          aria-multiselectable={multiple}
-          aria-labelledby={buttonId ?? defaultButtonId}
-          aria-orientation="vertical"
-          data-testid="select-listbox"
-          data-listbox-open={listboxOpen}
+          ref={popoverRef}
           // nb. Use `undefined` rather than `false` because Preact doesn't
           // handle boolean values correctly for this attribute (it will set
           // `popover="false"` instead of removing the attribute).
```
robertknight, line 634: Incidentally I think this might be fixed now, but it needs testing.

robertknight (approved):

src/components/input/Select.tsx
acelaya: Yep. I considered changing it, but I think we can tackle it individually

src/components/input/Select.tsx
acelaya: > Surprisingly, it didn't have any effect when `[popover]` was used in a `ul` element. I suspect it was because of the style resets that tailwind sets.

Yep, exactly.

image

PR #1791, opened 2024-11-20
Create new Popover component

acelaya:
> Depends on https://github.com/hypothesis/frontend-shared/pull/1789

Extract the popover element and the logic to dynamically position it from `Select` and into its own `Popover` module.

This will allow us to use this component in other contexts.

### Considerations

The popover currently requires an anchor element to be provided, and calculates its position and size based on that.

For now, I decided to keep some decisions made, like the fact that the popover appears below the anchor element unless there's not enough space, in which case it appears above it.

It also matches the anchor element's size, unless it can't fit its content, in which case it grows.

It should be possible to customize both of those (define initial position, determine how to calculate size, etc.), but I'm not going to change that as part of this PR.

We may also need to support a different way to determine the anchor, like providing a rectangle rather than an anchor element.

### Out of scope

1. I have finally decided not to add a documentation page for `Popover` in this PR, as it has enough changes already.
    Separately, I will add a documentation page explaining the component, how it works and behaves by default, and including some examples.
    EDIT: Addressed in https://github.com/hypothesis/frontend-shared/pull/1793
2. I will also make the Popover render its children only when it's open, making the behavior transparent for the Select and consistent anywhere where the Popover is used.
    EDIT: This is addressed in https://github.com/hypothesis/frontend-shared/pull/1792

### Testing steps

Sanity check that all examples in http://localhost:4001/input-select keep working as expected

### TODO

- [x] Add tests for popover
- [x] Remove some tests from `Select`, if they overlap with Popover tests

### DIFF
--- a/src/components/feedback/Popover.tsx
+++ b/src/components/feedback/Popover.tsx
@@ -1,0 +1,261 @@
+import classnames from 'classnames';
+import type { ComponentChildren, RefObject } from 'preact';
+import { useCallback, useLayoutEffect, useRef } from 'preact/hooks';
+
+import { ListenerCollection } from '../../util/listener-collection';
+
+/** Small space to apply between the anchor element and the popover */
+const POPOVER_ANCHOR_EL_GAP = '.15rem';
+
+/**
+ * Space in pixels to apply between the popover and the viewport sides to
+ * prevent it from growing to the very edges.
+ */
+export const POPOVER_VIEWPORT_HORIZONTAL_GAP = 8;
+
+type PopoverCSSProps =
+  | 'top'
+  | 'left'
+  | 'minWidth'
+  | 'marginBottom'
+  | 'bottom'
+  | 'marginTop';
+
+/**
+ * Manages the popover position manually to make sure it renders "next" to the
+ * anchor element (above or below). This is mainly needed when using the
+ * popover API, as that makes it render in the top layer, making it impossible
+ * to position it relative to the anchor element via regular CSS.
+ *
+ * @param asNativePopover - Native popover API is used to toggle the popover
+ * @param alignToRight - Whether the popover should be aligned to the right side
+ *                       of the anchor element or not
+ */
+function usePopoverPositioning(
+  anchorElementRef: RefObject<HTMLElement | undefined>,
+  popoverRef: RefObject<HTMLElement | undefined>,
+  popoverOpen: boolean,
+  asNativePopover: boolean,
+  alignToRight: boolean,
+) {
+  const adjustPopoverPositioning = useCallback(() => {
+    const popoverEl = popoverRef.current!;
+    const anchorEl = anchorElementRef.current!;
+
+    /**
+     * Set the positioning styles synchronously (not via <div style={computedStyles} />),
+     * to make sure positioning happens before other side effects.
+     * @return - A callback that undoes the property assignments
+     */
+    const setPopoverCSSProps = (
+      props: Partial<Record<PopoverCSSProps, string>>,
+    ) => {
+      Object.assign(popoverEl.style, props);
+      const keys = Object.keys(props) as PopoverCSSProps[];
+      return () => keys.map(prop => (popoverEl.style[prop] = ''));
+    };
+
+    const viewportHeight = window.innerHeight;
+    const {
+      top: anchorElDistanceToTop,
+      bottom: anchorElBottom,
+      left: anchorElLeft,
+      height: anchorElHeight,
+      width: anchorElWidth,
+    } = anchorEl.getBoundingClientRect();
+    const anchorElDistanceToBottom = viewportHeight - anchorElBottom;
+    const { height: popoverHeight, width: popoverWidth } =
+      popoverEl.getBoundingClientRect();
+
+    // The popover should render above only if there's not enough space below to
+    // fit it and there's more absolute space above than below
+    const shouldBeAbove =
+      anchorElDistanceToBottom < popoverHeight &&
+      anchorElDistanceToTop > anchorElDistanceToBottom;
+
+    if (!asNativePopover) {
+      // Set styles for non-popover mode
+      if (shouldBeAbove) {
+        return setPopoverCSSProps({
+          bottom: '100%',
+          marginBottom: POPOVER_ANCHOR_EL_GAP,
+        });
+      }
+
+      return setPopoverCSSProps({
+        top: '100%',
+        marginTop: POPOVER_ANCHOR_EL_GAP,
+      });
+    }
+
+    const { top: bodyTop, width: bodyWidth } =
+      document.body.getBoundingClientRect();
+    const absBodyTop = Math.abs(bodyTop);
+
+    // The available space is:
+    // - left-aligned popovers: distance from left side of anchor element to
+    //   right side of viewport
+    // - right-aligned popovers: distance from right side of anchor element to
+    //   left side of viewport
+    const availableSpace =
+      (alignToRight ? anchorElLeft + anchorElWidth : bodyWidth - anchorElLeft) -
+      POPOVER_VIEWPORT_HORIZONTAL_GAP;
+
+    let left = anchorElLeft;
+    if (popoverWidth > availableSpace) {
+      // If the popover is not going to fit the available space, let it "grow"
+      // in the opposite direction
+      left = alignToRight
+        ? POPOVER_VIEWPORT_HORIZONTAL_GAP
+        : left - (popoverWidth - availableSpace);
+    } else if (alignToRight && popoverWidth > anchorElWidth) {
+      // If a right-aligned popover fits the available space, but it's bigger
+      // than the anchor element, move it to the left so that it is aligned with
+      // the right side of the element
+      left -= popoverWidth - anchorElWidth;
+    }
+
+    return setPopoverCSSProps({
+      minWidth: `${anchorElWidth}px`,
+      top: shouldBeAbove
+        ? `calc(${absBodyTop + anchorElDistanceToTop - popoverHeight}px - ${POPOVER_ANCHOR_EL_GAP})`
+        : `calc(${absBodyTop + anchorElDistanceToTop + anchorElHeight}px + ${POPOVER_ANCHOR_EL_GAP})`,
+      left: `${Math.max(POPOVER_VIEWPORT_HORIZONTAL_GAP, left)}px`,
+    });
+  }, [asNativePopover, anchorElementRef, popoverRef, alignToRight]);
+
+  useLayoutEffect(() => {
+    if (!popoverOpen) {
+      return () => {};
+    }
+
+    // First of all, open popover if it's using the native API, otherwise its
+    // size is 0x0 and positioning calculations won't work.
+    const popover = popoverRef.current;
+    if (asNativePopover) {
+      popover?.togglePopover(true);
+    }
+
+    const cleanup = adjustPopoverPositioning();
+
+    if (!asNativePopover) {
+      return cleanup;
+    }
+
+    // Readjust popover position when any element scrolls, just in case that
+    // affected the anchor element position.
+    const listeners = new ListenerCollection();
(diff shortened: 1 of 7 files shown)


src/components/input/Select.tsx
```diff
@@ -607,37 +454,18 @@ function SelectMain<T>({
...
-            listboxClasses,
-          )}
-          ref={popoverRef}
-          // nb. Use `undefined` rather than `false` because Preact doesn't
-          // handle boolean values correctly for this attribute (it will set
-          // `popover="false"` instead of removing the attribute).
-          popover={listboxAsPopover ? 'auto' : undefined}
-          data-testid="select-popover"
+        <Popover
+          relativeTo={wrapperRef}
```
acelaya, line 458: This change is not related with the extraction of `Popover` specifically, but I realized the border is set in the container, not the button, which makes popover positioning and sizing calculation fail if it's done relative to the button.

It is easier to see the issue with a thicker border.

Using the button as the reference:
Captura desde 2024-11-21 10-21-42

Using the wrapper as the reference:
Captura desde 2024-11-21 10-21-13

src/components/input/Select.tsx
```diff
@@ -607,37 +454,18 @@ function SelectMain<T>({
...
-            listboxClasses,
-          )}
-          ref={popoverRef}
-          // nb. Use `undefined` rather than `false` because Preact doesn't
-          // handle boolean values correctly for this attribute (it will set
-          // `popover="false"` instead of removing the attribute).
-          popover={listboxAsPopover ? 'auto' : undefined}
-          data-testid="select-popover"
+        <Popover
+          anchorElementRef={wrapperRef}
```
acelaya, line 458: This change is not related with the extraction of `Popover` specifically (this would have been `anchorElementRef={buttonRef}`, but I realized the border is set in the container, not the button, which makes popover positioning and sizing calculation fail if it's done relative to the button.

It is easier to see the issue with a thicker border.

Using the button as the reference:
Captura desde 2024-11-21 10-21-42

Using the wrapper as the reference:
Captura desde 2024-11-21 10-21-13

src/components/feedback/test/Popover-test.js
```diff
@@ -0,0 +1,316 @@
...
+            <>
+              Content of popover
+              <button
+                data-testid="inner-button"
+                onClick={() => setOpen(prev => !prev)}
+              >
+                Focusable element inside popover
+              </button>
+            </>
+          ))}
```
acelaya, line 32: The popover should not render its children when it's closed, but a bunch of `Select` tests break if that behavior is changed.

I will tackle that separately, checking what's the best way to fix the Select tests. Perhaps mocking the `Popover` component in that case is the right approach, now that the Popover has its own test.

For now, this condition allows to test it as if it was already behaving like that.

src/components/input/test/Select-test.js
acelaya, line 1: All test cases that have been removed from here are now part of `Popover-test.js` (with a few variables renamed and titles changed to reference "popover" instead of "listbox").

However, I left a couple of overlapping tests that have helped catch regressions in the past, like the test that checks focus returns to the toggle button when the listbox is closed.

src/components/feedback/Popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+      cleanup();
+      listeners.removeAll();
+    };
+  }, [adjustPopoverPositioning, asNativePopover, popoverOpen, popoverRef]);
+}
+
+export type PopoverProps = {
+  children?: ComponentChildren;
+  classes?: string | string[];
+  variant?: 'panel' | 'custom'; // TODO 'tooltip'
```
acelaya, line 165: "Panel" variant adds the style used by `Select`, but it's likely the one we'll use for at-mentions as well.

It has:
* White background
* Box shadow
* Border
* Rounded corners
* Padding

src/components/input/test/Select-test.js
acelaya, line 1: All test cases that have been removed from here are now part of Popover-test.js (with a few variables renamed and titles changed to reference "popover" instead of "listbox").

However, I left a couple of overlapping tests that have helped catch regressions in the past, like the test that checks focus returns to the toggle button when the listbox is closed.

src/components/feedback/test/Popover-test.js
acelaya: Turnso out fixing those tests was easy https://github.com/hypothesis/frontend-shared/pull/1792/files#r1853917255

src/components/input/test/Select-test.js
acelaya, line 1: All test cases that have been removed from here are now part of Popover-test.js (with a few variables renamed and titles changed to reference "popover" instead of "listbox").

However, I left a couple of overlapping tests that have helped catch regressions in the past, like the test that checks focus returns to the toggle button when the listbox is closed.

src/components/feedback/Popover.tsx
```diff
@@ -0,0 +1,260 @@
...
```
robertknight, line 165: I would remove the "TODO: tooltip" comment for now. TODOs are mostly useful if they highlight a pitfall of the implementation that the reader needs to be aware of, whereas this is just a possible future feature.

src/components/feedback/Popover.tsx
```diff
@@ -0,0 +1,260 @@
...
+  asNativePopover: boolean,
+  alignToRight: boolean,
+) {
+  const adjustPopoverPositioning = useCallback(() => {
+    const popoverEl = popoverRef.current!;
+    const anchorEl = anchorElementRef.current!;
+
+    /**
+     * Set the positioning styles synchronously (not via `style` prop and a
+     * piece of state), to make sure positioning happens before other side
```
robertknight, line 47: This comment is slightly confusing on first read because the function does set the `style` JS property of the element. I understand that it means "set styles synchronously as opposed to via the `style` prop on the next render" though.

robertknight (approved):
LGTM. Just a few minor questions to clarify comments.

src/components/feedback/Popover.tsx
acelaya: Oops! This is actually an overlook.

src/components/feedback/Popover.tsx
acelaya: True!

I did in fact mean to reference the `style` component prop (as in `Popover style={{ ... }} />`), as you mentioned, but this function does set a `style` property in an object, so it's misleading 😅 

Let me give it a thought.

src/components/feedback/Popover.tsx
```diff
@@ -0,0 +1,260 @@
...
+  /**
+   * Determines to what side of the anchor element should the popover be
+   * aligned.
+   * Defaults to `left`
+   */
+  align?: 'right' | 'left';
+
+  /**
+   * Determines if focus should be restored when the popover is closed.
+   * Defaults to false.
```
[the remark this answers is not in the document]
acelaya: > Can you remind me why this prop defaults to false? I think for most use cases we would probably want it to be true.

Yep, I think the same. I made it `false` by default mainly for two reasons:

a) Consistency with `Dialog` component, where it's false by default.
b) Allow passing `restoreFocusOnClose` rather than `restoreFocusOnClose={false}` when you want to override the default value.

None of them is particularly important, and we have some exceptions to b, because we have agreed not setting the default value for the sake of aesthetics when setting props.

So I would be ok setting it to `true`, or even applying the logic unconditionally and not adding a prop.

src/components/feedback/Popover.tsx
```diff
@@ -0,0 +1,260 @@
...
+        'absolute z-5',
+        variant === 'panel' && [
+          'max-h-80 overflow-y-auto overflow-x-hidden',
+          'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
+        ],
+        asNativePopover && [
+          // We don't want the popover to ever render outside the viewport,
+          // and we give it a 16px gap
+          'max-w-[calc(100%-16px)]',
+          // Overwrite [popover] default styles
```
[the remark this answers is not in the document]
acelaya: Yes, the default styles come from the browser. See https://github.com/hypothesis/frontend-shared/pull/1789#discussion_r1850278776

> do they vary by browser?

No, it seems to be consistent, at least between Chrome and Firefox on Linux.

src/components/feedback/Popover.tsx
```diff
@@ -0,0 +1,260 @@
...
+          'right-0': align === 'right',
+          'min-w-full': true,
+        },
+        classes,
+      )}
+      ref={popoverRef}
+      // nb. Use `undefined` rather than `false` because Preact doesn't
+      // handle boolean values correctly for this attribute (it will set
+      // `popover="false"` instead of removing the attribute).
+      popover={asNativePopover ? 'auto' : undefined}
```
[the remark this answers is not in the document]
acelaya: I'll address this separately.

src/components/feedback/test/Popover-test.js
```diff
@@ -0,0 +1,316 @@
...
+  let containers;
+
+  beforeEach(() => {
+    wrappers = [];
+    containers = [];
+  });
+
+  afterEach(() => {
+    wrappers.forEach(w => w.unmount());
+    containers.forEach(c => c.remove());
```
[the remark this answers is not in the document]
acelaya: I'll add it to my TODO list

src/components/feedback/test/Popover-test.js
```diff
@@ -0,0 +1,316 @@
...
+  });
+
+  function createComponent(props = {}, options = {}) {
+    const { paddingTop = 0 } = options;
+
+    const container = document.createElement('div');
+    container.style.paddingTop = `${paddingTop}px`;
+    // Add horizontal paddings to the container, so that there's room for the
+    // popover to grow if needed
```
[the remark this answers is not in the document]
acelaya: Yes, exactly. This padding allows to individually test cases where:

1. The popover is exactly the same size as the anchor element, because contents fit that space.
2. The popover is bigger than the anchor element, but smaller than the viewport.
3. The popover is as big as the viewport.

Without the horizontal padding, all scenarios would behave as 1 and 3 at the same time, and it would not be possible to test 2.

src/components/feedback/test/Popover-test.js
```diff
@@ -0,0 +1,316 @@
...
+          const buttonLeft = buttonDOMNode.getBoundingClientRect().left;
+
+          return {
+            left: buttonLeft,
+            right: popoverDOMNode.getBoundingClientRect().width + buttonLeft,
+          };
+        },
+      })),
+
+      // Content is slightly longer. The popover matches one of the toggle's
```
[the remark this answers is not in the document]
acelaya: Yes. I have adjusted these comments a bit to make it more clear what is being compared. I also replaced references of "toggle button" with "anchor element"

src/components/feedback/test/Popover-test.js
```diff
@@ -0,0 +1,316 @@
...
+      // sides but spans further to the opposite one
+      {
+        children: 'slightly longer text'.repeat(2),
+        align: 'left',
+        getExpectedCoordinates: (popoverDOMNode, wrapper) => {
+          const buttonDOMNode = getToggleButton(wrapper).getDOMNode();
+          const buttonRect = buttonDOMNode.getBoundingClientRect();
+          const popoverRect = popoverDOMNode.getBoundingClientRect();
+          const popoverRightIncrement = popoverRect.width - buttonRect.width;
```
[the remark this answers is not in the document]
acelaya: This word is new to me 😅

src/components/feedback/Popover.tsx
acelaya: I changed the comment to `Set the positioning styles synchronously (not via <div style={computedStyles} />), to make sure positioning happens before other side effects.`

src/components/feedback/Popover.tsx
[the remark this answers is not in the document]
acelaya: I finally kept the `restoreFocusOnClose` prop, but made it default to `true`.

PR #1792, opened 2024-11-22
Do not render Popover children while closed

acelaya:
Depends on https://github.com/hypothesis/frontend-shared/pull/1791

Make sure the `Popover` component renders its children only while it's open.

This also removes the requirement to control this in the `Select` component, while adding the same benefit for other places where we use `Popover`.

### DIFF
--- a/src/components/input/test/Select-test.js
+++ b/src/components/input/test/Select-test.js
@@ -79,7 +79,7 @@ describe('Select', () => {
   const toggleListbox = wrapper => getToggleButton(wrapper).simulate('click');
 
   const isListboxClosed = wrapper =>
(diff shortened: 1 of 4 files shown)


src/components/input/test/Select-test.js
```diff
@@ -79,7 +79,7 @@ describe('Select', () => {
   const toggleListbox = wrapper => getToggleButton(wrapper).simulate('click');
 
   const isListboxClosed = wrapper =>
-    wrapper.find('[role="listbox"]').prop('data-listbox-open') === false;
+    wrapper.find('Popover').prop('open') === false;
```
acelaya, line 82: This was the only change needed to fix a bunch of failing `Select` tests.

robertknight (approved):

PR #1793, opened 2024-11-22
Add documentation page for Popover component

acelaya:
Depends on https://github.com/hypothesis/frontend-shared/pull/1792

Document the new `Popover` component introduced in https://github.com/hypothesis/frontend-shared/pull/1791

image

### DIFF
--- a/src/pattern-library/components/patterns/feedback/PopoverPage.tsx
+++ b/src/pattern-library/components/patterns/feedback/PopoverPage.tsx
@@ -1,0 +1,176 @@
+import { Link } from '../../../../components/navigation';
+import Library from '../../Library';
+
+export default function PopoverPage() {
+  return (
+    <Library.Page
+      title="Popover"
+      intro={
+        <>
+          <p>
+            <code>Popover</code> is a floating element rendered above other
+            content and positioned next to an anchor element.
+          </p>
+        </>
+      }
+    >
+      <Library.Section
+        title="Popover"
+        intro={
+          <>
+            <p>
+              By default, <code>Popover</code> will be displayed below the
+              anchor element, unless there is not enough space below.
+            </p>
+            <p>
+              It will always be at least as wide as the anchor element, but it
+              can grow to fit its content if needed. However, it will never grow
+              beyond the viewport.
+            </p>
+            <p>
+              In browsers that support it, <code>Popover</code> uses the{' '}
+              <Link
+                target="_blank"
+                href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover"
+              >
+                <code>popover</code>
+              </Link>{' '}
+              attribute and gets toggled via{' '}
+              <Link
+                target="_blank"
+                href="https://developer.mozilla.org/en-US/docs/Web/API/Popover_API"
+              >
+                popover API
+              </Link>
+              . Otherwise, it is rendered as an absolute-positioned element, so
+              it is recommended to wrap it and its anchor element in a
+              relative-positioned container.
+            </p>
+          </>
+        }
+      >
+        <Library.Pattern>
+          <Library.Usage symbolName="Popover" />
+          <Library.Example>
+            <Library.Demo
+              title="Basic Popover"
+              exampleFile="popover-basic"
+              withSource
+            />
+          </Library.Example>
+        </Library.Pattern>
+
+        <Library.Pattern title="Popover component API">
+          <Library.Example title="align">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Determines to which side of the anchor element the popover
+                should be aligned.
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
+                <code>{"'left' | 'right'"}</code>
+              </Library.InfoItem>
+              <Library.InfoItem label="default">
+                <code>{"'left'"}</code>
+              </Library.InfoItem>
+            </Library.Info>
+            <Library.Demo
+              title="Right-aligned Popover"
+              exampleFile="popover-right"
+              withSource
+            />
+          </Library.Example>
+          <Library.Example title="anchorElementRef">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                A reference to the element to which the popover should anchor,
+                which will be used to calculate the popover size and
+                positioning.
+              </Library.InfoItem>
+              <Library.InfoItem label="type">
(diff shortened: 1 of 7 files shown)


src/pattern-library/examples/popover-basic.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { Popover } from '../../components/feedback';
+import { Button } from '../../components/input';
+import { useClickAway } from '../../hooks/use-click-away';
+
+export default function App() {
+  const [open, setOpen] = useState(false);
+  const buttonRef = useRef<HTMLButtonElement | null>(null);
+
+  useClickAway(buttonRef, () => setOpen(false));
```
acelaya, line 11: I added this because popovers handled via the popover API are automatically hidden when clicking away, which makes the `open` state no longer reflect reality.

I didn't notice this first because in the case of `Select` we have `useClickAway`, `useFocusAway`, etc.

I will address this separately, but we'll probably need to add an `onClose` callback to the `Popover` and handle a few native events, like `toggle` one.

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+export default function PopoverPage() {
+  return (
+    <Library.Page
+      title="Popover"
+      intro={
+        <>
+          <p>
+            <code>Popover</code> is a floating component that can be shown
+            next/relative to another anchor element.
```
robertknight, line 12:
```suggestion
            <code>Popover</code> is a floating element rendered above other content and positioned next to an anchor element.
```

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+          </p>
+        </>
+      }
+    >
+      <Library.Section
+        title="Popover"
+        intro={
+          <>
+            <p>
+              By default, <code>Popover</code> will be displayed under the
```
robertknight, line 22:
```suggestion
              By default, <code>Popover</code> will be displayed below the
```

"under" might be interpreted as referring to the Z axis instead of Y axis.

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+              anchor element, unless there is not enough space below.
+            </p>
+            <p>
+              It will always be at least as big as the anchor element, but it
```
robertknight, line 26:
```suggestion
              It will always be at least as wide as the anchor element, but it
```

Make it clear which dimension "big" refers to.

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+              can grow to fit its content if needed. However, it will never grow
+              or be displayed outside of the viewport.
```
robertknight, line 28:
```suggestion
              beyond the viewport.
```

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+              attribute and gets toggled via{' '}
+              <Link
+                target="_blank"
+                href="https://developer.mozilla.org/en-US/docs/Web/API/Popover_API"
+              >
+                popover API
+              </Link>
+              . Otherwise, it is rendered as an absolute-positioned element, so
+              it is recommended to wrap it and its anchor element in a
+              relative-positioned container.
```
robertknight, line 47: The examples in this page _don't_ do this, so if you modify them to set `asNativePopover={false}` they render incorrectly.

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+              <Library.InfoItem label="description">
+                Determines if the{' '}
+                <Link
+                  target="_blank"
+                  href="https://developer.mozilla.org/en-US/docs/Web/API/Popover_API"
+                >
+                  popover API
+                </Link>{' '}
+                should be used. It{"'"}s mainly used as a test seam, but can
+                help explicitly disabling this behavior if needed.
```
robertknight, line 106:
```suggestion
                be used to explicitly disable use of the native popover API.
```

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+              </Library.InfoItem>
+              <Library.InfoItem label="default">
+                <code>false</code>
+              </Library.InfoItem>
+            </Library.Info>
+          </Library.Example>
+          <Library.Example title="variant">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Set the <code>Popover</code> theming.
```
robertknight, line 157:
```suggestion
                Set the <code>Popover</code> style variant.
```

src/pattern-library/examples/popover-basic.tsx
robertknight: > I added this because popovers handled via the popover API are automatically hidden when clicking away, which makes the open state no longer reflect reality.

This seems like an obvious hazard that we should at least document.

> I will address this separately, but we'll probably need to add an onClose callback to the Popover and handle a few native events, like toggle one.

Yes, definitely. Is there an issue tracking this?

acelaya: > Is there an issue tracking this?

Done https://github.com/hypothesis/frontend-shared/issues/1799

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+            />
+          </Library.Example>
+        </Library.Pattern>
+
+        <Library.Pattern title="Popover component API">
+          <Library.Example title="align">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Determines to what side of the anchor element should the popover
+                be aligned.
```
robertknight, line 68:
```suggestion
                Determines to which side of the anchor element the popover should
                be aligned.
```

robertknight (approved):

PR #1803, opened 2024-11-29
fix: Types after preact 10.25

rschristian:
My attempt to help out with #1797

The new per-element type interfaces introduces in Preact 10.25 should provide better type assistance but does require a switch over. Additionally, whilst I'm not familiar with this code base I stumbled upon what looks like docs while grepping through; I took the opportunity to update those (seemingly) accordingly?

There have been a few attribute fixes upstream already since 10.25, but not yet released. This PR is therefore a result of manually copying over the type changes since to see if there are any issues we hadn't yet caught. As such, this is based on an in-progress 10.25.1, not really 10.25.0.

There's a few issues remaining:
 - ~~`src/components/icons/SpinnerCircle.tsx`: Property `type` does not exist...~~
   - ~~Upstream issue, was missing on the interface. Will get that corrected~~ **FIXED**, will go out in next version
 - `src/pattern-library/components/patterns/navigation/LinkButtonPage.tsx`: Property `href` does not exist...
   - This error seems correct? Very unfamiliar with the code, but it does appear like you're trying to pass `href` to a button
 - ~~`src/components/input/Checkbox.tsx`: Property `call` does not exist...~~
   - ~~This one's a bit weirder and I need some time with yet. I'm not quite sure what's going wrong, but I'd assume it's an upstream issue with our types.~~ **FIXED**, will also go out in the next version

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "mustache": "^4.2.0",
     "plop": "^4.0.0",
     "postcss": "^8.3.9",
-    "preact": "^10.5.6",
+    "preact": "^10.25.1",
     "prettier": "^3.0.0",
     "rollup": "^4.0.2",
     "rollup-plugin-string": "^3.0.0",
@@ -60,7 +60,7 @@
     "yalc": "^1.0.0-pre.50"
   },
   "peerDependencies": {
-    "preact": "^10.4.0"
+    "preact": "^10.25.1"
   },
   "scripts": {
     "build-lib": "babel src/ --out-dir lib/ --extensions '.js,.ts,.tsx' --source-maps --ignore '**/test' --ignore '**/karma.config.js'",
--- a/src/components/input/Button.tsx
+++ b/src/components/input/Button.tsx
@@ -41,7 +41,7 @@ type ComponentProps = {
 };
 
 export type HTMLButtonAttributes = Omit<
-  JSX.HTMLAttributes<HTMLButtonElement>,
+  JSX.ButtonHTMLAttributes<HTMLButtonElement>,
   'icon' | 'size'
 >;
 
--- a/src/pattern-library/components/patterns/navigation/LinkButtonPage.tsx
+++ b/src/pattern-library/components/patterns/navigation/LinkButtonPage.tsx
@@ -106,13 +106,9 @@ export default function LinkButtonPage() {
                   Underline: none (default)
                 </LinkButton>
 
-                <LinkButton href="https://www.example.com" underline="hover">
-                  Underline: hover
-                </LinkButton>
+                <LinkButton underline="hover">Underline: hover</LinkButton>
 
-                <LinkButton href="https://www.example.com" underline="always">
-                  Underline: always
-                </LinkButton>
+                <LinkButton underline="always">Underline: always</LinkButton>
                 <p>
                   LinkButtons should be{' '}
                   <LinkButton underline="always" inline>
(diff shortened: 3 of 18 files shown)


acelaya:
Thanks for taking a look @rschristian!

rschristian:
@acelaya Preact 10.25.1 is out will all the fixes, should be good to go-ish. You still have an issue with passing an `href` prop to an element that's annotated as taking `HTMLButtonAttributes`, and you seemingly don't try to use that `href` anywhere, so it might just be a mistake? Not quite sure.

Feel free to make additional edits, copy/paste from here, whatever. And of course ping if you run into any more trouble with the types. We've made a big change in recent versions but hope to get everyone switched over without too much difficulty.

src/components/input/Button.tsx
```diff
@@ -41,7 +41,7 @@ type ComponentProps = {
 };
 
 export type HTMLButtonAttributes = Omit<
-  JSX.HTMLAttributes<HTMLButtonElement>,
+  JSX.ButtonHTMLAttributes<HTMLButtonElement>,
```
acelaya, line 44: Just have a question. Are the new  `JSX.ButtonHTMLAttributes`, `JSX.InputHTMLAttributes`, etc., available in preact versions older than 10.25?

If not, we may need to bump the minimum required preact version of this library, since we ship types with it.

rschristian: Great question: they've were available from `preact/compat` since 10.23 but are only available from `preact` since 10.25.

I see only now there was a peer dependency on preact, sorry! In that case, indeed, you'll probably need a bump.

---

You might be able to work out a system if you need to keep compat with older versions, though it might require some tooling. Essentially, the `HTMLAttributes` used here was a mega interface of all possible attributes (across all elements) which has been renamed to `AllHTMLAttributes` in v10.25+. `HTMLAttributes` in v10.25+ therefore is the purely global attributes/props, things like `class`, `id`, etc. You could probably set up some sort of fallback situation there if you needed.

acelaya: Thanks! Bumping the minimum required version should be fine 🙂

acelaya:
Thanks @rschristian! This is much appreciated.

We just realized our workflows are not configured to run on pull requests, but on push only (we usually create PRs here directly), so continuous integration didn't trigger.

I'll clone this locally, fix the `href` issue, and verify everything works. Then I'll merge this.

package.json
```diff
@@ -60,7 +60,7 @@
     "yalc": "^1.0.0-pre.50"
   },
   "peerDependencies": {
-    "preact": "^10.4.0"
+    "preact": "^10.25.1"
```
acelaya, line 63: Setting minimum required version to 10.25.1 rather than 10.25.0, since it includes some important type fixes.

src/pattern-library/components/patterns/navigation/LinkButtonPage.tsx
```diff
@@ -106,13 +106,9 @@ export default function LinkButtonPage() {
...
 
-                <LinkButton href="https://www.example.com" underline="hover">
-                  Underline: hover
-                </LinkButton>
+                <LinkButton underline="hover">Underline: hover</LinkButton>
 
-                <LinkButton href="https://www.example.com" underline="always">
-                  Underline: always
-                </LinkButton>
+                <LinkButton underline="always">Underline: always</LinkButton>
```
acelaya, line 111: The `href` attribute was not used in `LinkButton`, so it's safe to remove.

These were probably some old examples from a time in which it may have supported it.

acelaya (approved):

PR #1818, opened 2024-12-03
Rename `Library.{Pattern, Example}` => `Library.{SectionL2, SectionL3}`

robertknight:
The hierarchical relationship between `Library.{Section, Pattern, Example}` was unclear and the `Pattern` and `Example` components had an unclear purpose since they were used as sections with different kinds of content. Rename these
components to make the structure clearer: `Page` -> `Section` -> `SectionL2` ->
`SectionL3` where `SectionL{N}` is effectively an alias for `<Section level={N}>`. In the process the meaning of the `level` prop has been adjusted to match the logical section level.

In the process also correct the usage of the `<main>` element to refer to just the main content of the page (ie. excluding navigation) instead of the whole page.

Discussed [here in Slack](https://hypothes-is.slack.com/archives/C1M8NH76X/p1733227809381629).


src/pattern-library/components/Library.tsx
```diff
@@ -633,12 +645,12 @@ export default {
...
   Demo,
-  Example,
+  SectionL3,
   Info,
   InfoItem,
   Link,
   Page,
-  Pattern,
+  SectionL2,
   Section,
```
acelaya, line 654: Maybe we should sort these.

acelaya (approved):
LGTM

PR #1828, opened 2024-12-09
Update Pagination styling

robertknight:
Update the styling of the Pagination control as follows:

 - Remove the background color from non-active items. This is a vestige from when this component was first created for use in the client's Notebook UI.
 - Use Unicode ellipsis for elided pages and page the elided-page markers approximately the same width as a (small) page number
 - Prevent text of pagination controls from being selected
 - Fix an issue where the Prev/Next button would have incorrect vertical alignment if the font size did not match that used in the client
 - Disable and hide prev/next controls rather than removing them when unavailable. This keeps the position of other controls consistent when moving from page to page, without needing to place these controls in a container with a hard-coded size

**Before:**

pagination-before

**After:**

pagination-after

### DIFF
--- a/src/components/navigation/Pagination.tsx
+++ b/src/components/navigation/Pagination.tsx
@@ -1,17 +1,25 @@
 import classnames from 'classnames';
-import type { JSX } from 'preact';
+import type { ComponentChildren } from 'preact';
 
-import type { PresentationalProps } from '../../types';
 import { pageNumberOptions } from '../../util/pagination';
 import { ArrowLeftIcon, ArrowRightIcon } from '../icons';
 import Button from '../input/Button';
-import type { ButtonProps } from '../input/Button';
 
-type NavigationButtonProps = PresentationalProps &
-  ButtonProps &
-  Omit<JSX.HTMLAttributes<HTMLButtonElement>, 'icon' | 'size'>;
+type NavigationButtonProps = {
+  children: ComponentChildren;
+  invisible?: boolean;
+  onClick: (e: MouseEvent) => void;
+  pressed?: boolean;
+  title?: string;
+};
 
-function NavigationButton({ ...buttonProps }: NavigationButtonProps) {
+function NavigationButton({
+  children,
+  invisible = false,
+  pressed = false,
+  onClick,
+  title,
+}: NavigationButtonProps) {
   return (
     <Button
       classes={classnames(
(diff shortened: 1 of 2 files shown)


src/components/navigation/Pagination.tsx
```diff
@@ -1,30 +1,49 @@
...
-type NavigationButtonProps = PresentationalProps &
-  ButtonProps &
-  Omit<JSX.HTMLAttributes<HTMLButtonElement>, 'icon' | 'size'>;
+type NavigationButtonProps = {
+  children: ComponentChildren;
+  invisible?: boolean;
+  onClick: (e: MouseEvent) => void;
+  pressed?: boolean;
+  title?: string;
+};
```
robertknight, line 14: I changed to explicitly listing all the used props here rather than using forwarding. This makes it easier to understand the variations in how this component is used. In general I am not a huge fan of using prop forwarding heavily except for components which are primarily intended to be styled wrappers around basic DOM components.

src/components/navigation/Pagination.tsx
```diff
@@ -1,30 +1,49 @@
...
         // These colors are the same as the "dark" variant of IconButton
-        'text-grey-7 bg-grey-2 enabled:hover:text-grey-9 enabled:hover:bg-grey-3',
+        'text-grey-7 enabled:hover:text-grey-9 enabled:hover:bg-grey-3',
         'disabled:text-grey-5 aria-pressed:bg-grey-3 aria-expanded:bg-grey-3',
+
+        // Disabled navigation buttons are rendered as invisible and disabled
+        // rather than removed so that the overall width of the component and
+        // positions of child controls doesn't change too much when navigating
+        // between pages.
+        invisible && 'invisible',
```
acelaya, line 36: I reckon this should be equivalent to `{ invisible }`

```suggestion
        { invisible },
```

src/components/navigation/Pagination.tsx
```diff
@@ -1,30 +1,49 @@
...
         'font-semibold rounded',
         // These colors are the same as the "dark" variant of IconButton
-        'text-grey-7 bg-grey-2 enabled:hover:text-grey-9 enabled:hover:bg-grey-3',
+        'text-grey-7 enabled:hover:text-grey-9 enabled:hover:bg-grey-3',
         'disabled:text-grey-5 aria-pressed:bg-grey-3 aria-expanded:bg-grey-3',
+
+        // Disabled navigation buttons are rendered as invisible and disabled
+        // rather than removed so that the overall width of the component and
+        // positions of child controls doesn't change too much when navigating
+        // between pages.
```
acelaya, line 35: I would even consider keeping them visible but disabled, but this should work for now without introducing design changes.

acelaya (approved):

PR #1829, opened 2024-12-09
Revise pagination item calculation

robertknight:
Revise how the pagination items (page numbers and elided page markers) are calculated, such that the number of _items_ rather than the number of _numbered pages_ remains consistent as the user navigates through the list. Provided each item is rendered with a roughly consistent width, this keeps the overall width of the control and the position of child components consistent when navigating through the list.

page-2

page-5

When this middle controls have a consistent width, the Pagination also works better in cases where the Prev/Next controls are adjacent to the middle items, rather than aligned with the edge of the content area above (the table here).

**Testing:**

Go to http://localhost:4001/navigation-pagination, navigate through the list. Observe that the number of items displayed is the same regardless of current page. Compare with https://patterns.hypothes.is/navigation-pagination where this is not the case.

### DIFF
--- a/src/util/test/pagination-test.js
+++ b/src/util/test/pagination-test.js
@@ -1,22 +1,129 @@
-import { pageNumberOptions } from '../pagination';
+import { paginationItems } from '../pagination';
 
-describe('sidebar/util/pagination', () => {
-  describe('pageNumberOptions', () => {
-    [
-      { args: [1, 10, 5], expected: [1, 2, 3, 4, null, 10] },
-      { args: [3, 10, 5], expected: [1, 2, 3, 4, null, 10] },
-      { args: [6, 10, 5], expected: [1, null, 5, 6, 7, null, 10] },
-      { args: [9, 10, 5], expected: [1, null, 7, 8, 9, 10] },
-      { args: [2, 3, 5], expected: [1, 2, 3] },
-      { args: [3, 10, 7], expected: [1, 2, 3, 4, 5, 6, null, 10] },
-      { args: [1, 1, 5], expected: [] },
-    ].forEach(testCase => {
-      it('should produce expected available page numbers', () => {
-        assert.deepEqual(
-          pageNumberOptions(...testCase.args),
-          testCase.expected,
-        );
+describe('paginationItems', () => {
+  [
+    // Only one page
+    {
+      current: 1,
+      total: 1,
+      expected: [],
+    },
+
+    // No pages elided.
+    {
+      current: 1,
+      total: 3,
+      expected: [1, 2, 3],
+    },
+
+    // Pages elided after current.
+    {
+      current: 1,
+      total: 4,
+      expected: [1, 2, null, 4],
+    },
+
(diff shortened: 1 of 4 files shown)


src/util/test/pagination-test.js
```diff
@@ -1,22 +1,115 @@
...
+    it('should produce expected items', () => {
+      const items = paginationItems(current, total, {
+        boundaryCount,
+        siblingCount,
       });
+      assert.deepEqual(items, expected);
     });
   });
+
+  function* configurations() {
```
robertknight, line 65: The test above checks a few specific examples against expected results. The test below is a property test that generates a range of configurations and for each verifies that certain properties are satisfied for every possible `current` page.

src/util/test/pagination-test.js
```diff
@@ -1,22 +1,129 @@
...
+        for (let total = 1; total <= 10; total++) {
+          yield { total, boundaryCount, siblingCount };
+        }
+      }
+    }
+  }
+
+  function isStrictlyIncreasing(list) {
+    return list.every((item, idx) => idx === 0 || item > list[idx - 1]);
+  }
```
robertknight, line 87: "strictly increasing" as in "a strictly increasing function"

src/util/test/pagination-test.js
```diff
@@ -1,22 +1,129 @@
...
+      total: 4,
+      expected: [1, 2, null, 4],
+    },
+
+    // Pages elided before current.
+    {
+      current: 4,
+      total: 6,
+      expected: [1, null, 3, 4, 5, 6],
+    },
```
acelaya, line 31: This test is great, because it helps document the functionality, which is pretty complex.

However, it made me realize that, considering the elision in a 6-page pagination (`[1, null, 3, 4, 5, 6]`), I would have expected no elided pages for a 4-page pagination (`[1, 2, 3, 4]`) regardless of current page, but it seems we do elide (`[1, 2, null, 4]`).

Can you explain the algorithm to me as if I were a five-year-old 😅?

acelaya (approved):
The algorithm is maybe not super easy to understand at a glance, but it does bring the benefits you describe here, and the new tests help a lot documenting it.

src/util/test/pagination-test.js
robertknight: In prose:

1. If there is one page, return an empty list
2. Calculate if there are elided pages before the current page
3. If no, create a list of all pages before the current page. If yes, create a list of the pages before the elided range, then `null`, then pages after the elided range.
4. Add the current page to the list
5. Calculate if there are elided pages after the current page
6. If no, append all pages after the current page. If yes, append a list of pages after the current page, before the elided range, then `null`, then pages after the elided range
7. Calculate the maximum number of items we might show, for all possible values of the `current` page
8. While the current number of items in the list is less than the maximum, expand the list by inserting an extra page number and "shrinking" the elided range. We preferentially take the page numbers from the elided range after the current page, but if that is empty, take them from the elided numbers before the current range.

PR #1849, opened 2025-01-15
Aligning icons to Hypothicons library

jaredpdesigns:
PR aligns icons from [Figma library](https://www.figma.com/design/pzAaIXzI6ro4WvwYv6i21m/Hypothicons?node-id=554-68&p=f&t=tE5uIrAsOtn5qRF6-0) to what's in the Pattern library. All base `.svg` files have been consistently formatted and have the same properties except those noted below. 

***

**Deprecated icons:**
- `AnnotateAlt`
- `CheckboxOutline`
- `GlobeAltLock`
- `LockAltFilled`
- `LockAlt`
- `ReplyAlt`
- `SortAlt`
- `TagAltFilled`
- `TagAlt`

***

**Breaking change:**
There was were two export name changes that would require a separate PR for the `client` repo. I created a branch, but didn't have permission to publish.

Export name changes:
- `GlobeAltIcon` -> `GlobeIcon`
- `GlobeAltLockIcon` -> `GlobeLockIcon`

Files affected:
- `src/sidebar/components/GroupList/GroupListItem.tsx`
- `src/sidebar/components/GroupList/test/GroupListItem-test.js`

***

**Outstanding questions/comments:**

- I didn't adjust `MenuCollapse` or `MenuExpand` as they appear to be slightly larger versions of `CaretDown`/`CaretUp`—I don't understand why these exist or in what context a larger variant is needed.
- I didn't adjust `SpinnerCircle` or `SpinnerSpokes`, these feel like they should be a separate component, not part of the icon library.


acelaya:
I would prefer the "globe-lock" icon to keep a more generic name. Using it for restricted groups is one use-case, but we could find other in future, and it would feel weird to use an icon named "group-restricted" anywhere else.

acelaya:
It would be great to test the icons that have been adjusted everywhere where we are using them, as there have been icon changes in the past that looked reasonable in isolation, but then they had to be adjusted when used in actual apps.

jaredpdesigns:
Thanks for your feedback @robertknight I will adjust the name back to `GlobeLock`. Do you have a suggestion for thoroughly testing these in different environments other than the Pattern library? For example, can we cut a beta release and test in multiple repos that consume the library?

acelaya:
> Do you have a suggestion for thoroughly testing these in different environments other than the Pattern library?

We usually use yalc to locally test changes in the frontend-shared library before releasing them. It is documented here https://github.com/hypothesis/frontend-shared/blob/02555449419a7520321bda7d0c5b9c0fe48b030c/docs/developing.md#testing-locally-with-other-projects

robertknight:
> We usually use yalc to locally test changes in the frontend-shared library before releasing them

This would require Jared to get a complete Hypothesis development environment working, which is not trivial, especially for the LMS. We'll have to help with this stage.

acelaya:
> > We usually use yalc to locally test changes in the frontend-shared library before releasing them
> 
> This would require Jared to get a complete Hypothesis development environment working, which is not trivial, especially for the LMS. We'll have to help with this stage.

Yeah, that's true. Probably releasing a beta would have the same problem, so the most pragmatic approach is that one of us tests the changes.

acelaya:
Ireckon we could close https://github.com/hypothesis/frontend-shared/issues/603 once this is merged.

src/components/input/Checkbox.tsx
```diff
@@ -39,7 +39,7 @@ export type CheckboxProps = CompositeProps &
 export default function Checkbox({
   checked,
   defaultChecked = false,
-  icon = CheckboxOutlineIcon,
+  icon = CheckboxIcon,
   checkedIcon = CheckboxCheckedIcon,
```
acelaya, line 43: I'd suggest we take the opportunity and set `CheckboxCheckedFilledIcon` as the default `checkedIcon`

acelaya (changes requested):
I tried to test all these changes in `client`, `h`, `via` and `lms`, and I didn't find anything too problematic.

I created a couple draft PRs in [`client`](https://github.com/hypothesis/client/pull/6781) and [`h`](https://github.com/hypothesis/h/pull/9280), which are the only projects which need changes.

There's still something to be resolved in `client`, where the `PointerDown` and `PointerUp` icons do not apply the `fill`when passed via CSS class. I need to investigate a bit further.

That said, I'd propose these changes:

> Deprecated icons

Considering the PR already introduces breaking changes, I'd suggest we get rid of these entirely, instead of moving them to a `_deprecated` subfolder (both SVG and TSX files).

I already checked that none of our projects are using them.

Then, once we merge this PR we can release a new major version of the library. I don't think anyone is using it other than us, but for us it shouldn't make a difference to update from 8.16 to 9 instead of 8.17 anyway.

And since we are publishing a new major version, maybe we can take the opportunity and change the default checked icon for checkboxes, to use the filled variant.

acelaya:
> **Outstanding questions/comments:**
> 
> * I didn't adjust `MenuCollapse` or `MenuExpand` as they appear to be slightly larger versions of `CaretDown`/`CaretUp`—I don't understand why these exist or in what context a larger variant is needed.

I can only speculate, but I know Lyza had plans to move some Menu-related components from client to this library, but that never happened.

Those components use the MenuCollapse and MenuExpand icons, but I don't know why new icons were created rather than using the caret ones.

> * I didn't adjust `SpinnerCircle` or `SpinnerSpokes`, these feel like they should be a separate component, not part of the icon library.

I suppose it was just convenient to manage them in the same way, since they were SVGs to begin with.

acelaya:
I continued investigating the initial problem I found with `PointerUpIcon` and `PointerDownIcon`, and it seems they were intentionally build in a very particular way to allow them to be used in popover-like components, to indicate which is (point to) their anchor element.

Here are two examples of their usage in client:

image

image

Those icons had a different height than the rest so that it was easier to align them with the very edge of the popover.

Also, it was possible to customize their "inner" color and "border" color by setting text color (for the border) and fill color (for the inner color), via `fill-white text-grey-3` tailwind classes.

With the new implementation both things are broken: They now show a gap between the icon and the popover, and they are rendered fully in grey color.

image

The gap can be adjusted via negative margins. Not great, but perhaps worth it to keep consistency.

But I reckon fixing the colors would require some changes in the paths.

Based on current SVG:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="9" data-component="PointerUpIcon">
    <path stroke="currentColor" d="m.5 9 7-8 8 8"></path>
</svg>
```

I asked ChatGPT about the meaning of `d="m.5 9 7-8 8 8"`, and this was the answer, in case it helps:

> The `d` attribute in the `<path>` element defines the path to be drawn. The value `"m.5 9 7-8 8 8"` is a series of commands and parameters that describe the shape of the path. Here's a breakdown of what each part means: 
>
> * `m.5 9`: This is a "move to" command (m) that moves the pen to the coordinates (0.5, 9). The lowercase `m` indicates that the coordinates are relative to the current position.
> * `7-8`: This is a "line to" command that draws a line from the current position to a point that is 7 units to the right and 8 units up from the current position.
> * `8 8`: This is another "line to" command that draws a line from the current position to a point that is 8 units to the right and 8 units down from the current position.
>
> So, the path described by `d="m.5 9 7-8 8 8"` starts at (0.5, 9), draws a line to (7.5, 1), and then draws another line to (15.5, 9). This forms a "V" shape.

src/components/input/Checkbox.tsx
jaredpdesigns: Updated

jaredpdesigns:
@acelaya @robertknight I made the adjustments requested.

I ended up reverting `PointerDown` and `PointerUp` back to their original values since they have a very specific use case, and probably aren't really being used as _regular_ icons.

Let me know if there is anything else that needs adjusting.

acelaya (approved):
Looks good!

I tested the changes again, and I don't see any noticeable issue.

eslint.config.js
```diff
@@ -14,7 +14,7 @@ export default [
 
   // Icons
   {
-    files: ['src/components/icons/*.tsx'],
+    files: ['src/components/icons/**/*.tsx'],
```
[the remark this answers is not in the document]
acelaya: I took the liberty to rever it

PR #1857, opened 2025-01-21
Make Popover a presentational component

acelaya:
This PR makes the `Popover` a presentational component, accepting any DIV HTML attribute which is passed down to the outermost div.

The motivation for this PR is that we needed to be able to pass an `id` to `Popover`. Since I imagine we may need to add more arbitrary attributes in future, I checked how we handle similar components, and [PresentationalComponents](https://patterns.hypothes.is/using-components#presentational-components-api) seemed like a reasonable approach.


acelaya:
I'm putting this back as draft, as it may not be needed in the end.

acelaya:
As described above, the motivation for this was supporting proper accessibility in at-mentions (see https://github.com/hypothesis/client/pull/6769), but for that the `id` attribute needs to be set in the listbox, not the popover, so closing this for now.

PR #1864, opened 2025-01-31
Ensure Select onScroll is bound to the Popover, not the listbox

acelaya:
This PR fixes a regression that was likely introduced when the `Popover` component was implemented.

That made the `Select`'s `onListboxScroll` to never be invoked again, as due to the new DOM structure, the element that ends up with scroll is not the listbox, but the popover itself.

This PR deprecates `onListboxScroll`, and adds a new `onPopoverScroll` prop to replace it (its default value is whatever was passed to `onListboxScroll`, for backwards compatibility), and also makes sure the `onPopoverScroll` is passed to the right DOM element.

> This PR is part of https://github.com/hypothesis/support/issues/178

### DIFF
--- a/src/components/input/Select.tsx
+++ b/src/components/input/Select.tsx
@@ -305,8 +305,11 @@ type BaseSelectProps = CompositeProps & {
    */
   listboxAsPopover?: boolean;
 
-  /** A callback passed to the listbox onScroll */
+  /** @deprecated. Use onPopoverScroll instead */
   onListboxScroll?: JSX.HTMLAttributes<HTMLUListElement>['onScroll'];
+
+  /** A callback passed to the popover onScroll */
+  onPopoverScroll?: JSX.HTMLAttributes<HTMLElement>['onScroll'];
 
   /**
    * Indicates how overflowing content should be handled in the listbox.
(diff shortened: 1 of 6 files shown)


src/components/input/Select.tsx
```diff
@@ -455,6 +459,7 @@ function SelectMain<T>({
           asNativePopover={listboxAsPopover}
           align={alignListbox}
           classes={popoverClasses}
+          onScroll={onPopoverScroll}
```
acelaya, line 462: The overflowing and height controls are on the Popover now, so that's the element where we should listen for `onScroll`.

src/pattern-library/components/patterns/feedback/PopoverPage.tsx
```diff
@@ -142,6 +142,17 @@ export default function PopoverPage() {
               </Library.InfoItem>
             </Library.Info>
           </Library.SectionL3>
+          <Library.SectionL3 title="onScroll">
+            <Library.Info>
+              <Library.InfoItem label="description">
+                Callback passed to the outermost element <code>onScroll</code>,
+                so that it gets invoked when a scroll event is triggered there.
```
robertknight, line 149:
```suggestion
                Handler for "scroll" events on the outermost element.
```

robertknight (approved):

PR #1879, opened 2025-02-14
Allow time to be excluded when calling formatDateTime

acelaya:
Add a new `includeTime?: boolean` option for the `formatDateTime` function, which is `true` by default, and allows the time to be disabled when set to `false`.

We'll use this to format users join date in `@mentions` profile preview popovers.


src/util/date-and-time.ts
```diff
@@ -261,18 +264,20 @@ export type FormatDateTimeOptions = {
  */
 export function formatDateTime(
   date: Date | string,
-  options?: FormatDateTimeOptions,
+  { includeWeekday = false, includeTime = true }: FormatDateTimeOptions = {},
+  /* istanbul ignore next - Test seam. JS `Intl` API implementation. */
+  Intl?: IntlType,
```
acelaya, line 269: I have extracted the `Intl` test seam from the options, so that it can be individually marked as ignored for code coverage, while we can have a dedicated test for when the options object is not provided at all.

I could have left it as it was, and to `options?.includeTime`, but whereas this is handy when the default value is `false, like with `options?.includeWeekday`, when the default value is true, it's more cumbersome to do `(options?.includeTime ?? true) ? ... : ...`.

robertknight (approved):

Issue #1906, opened 2025-03-05
Popover should not restore focus on close if user focused another element outside the popover


robertknight:
When a `Popover` is closed as a result of the user focusing a different element outside of the popover, the popover should not attempt to restore focus to whichever element had the focus when the popover was initially opened.

See also notes in https://github.com/chakra-ui/chakra-ui/issues/213, which is about the same issue in a different component library.

Downstream issue: https://github.com/hypothesis/client/issues/6858

PR #1907, opened 2025-03-06
Avoid restoring focus when clicking on focusable element outside Popover

robertknight:
Fix an issue where clicking on a focusable element outside of an open popover, other than the previously focused element, would cause focus to be reverted back to the previously focused element.

When native popovers are used, they handle focus restoration for us [^1], so we don't need to do anything when the popover is closed. Setting `restoreFocusOnClose` to `false` didn't actually work in this case. When non-native popovers are used, focus reverts to `document.body` when the popover is hidden, unless the user clicked on a different focusable element outside the popover. Only when it has reverted to `document.body` do we want to restore focus.

Fixes https://github.com/hypothesis/frontend-shared/issues/1906

[^1]: See `focusPreviousElement` references in https://html.spec.whatwg.org/multipage/popover.html#dom-hidepopover

### DIFF
--- a/src/components/feedback/Popover.tsx
+++ b/src/components/feedback/Popover.tsx
@@ -248,6 +248,9 @@ export type PopoverProps = {
   /**
    * Determines if focus should be restored when the popover is closed.
    * Defaults to true.
+   *
+   * @deprecated - Setting this prop to `false` does not work if native popovers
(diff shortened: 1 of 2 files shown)


acelaya (approved):
LGTM

PR #1922, opened 2025-03-24
Fix logo size

acelaya:
A few weeks ago we [updated and consolidated](https://github.com/hypothesis/frontend-shared/pull/1849/files) all our icons. Among other things, that accidentally reduced the default size of the logo icon.

This PR ensures the same size as before is used when rendered in the page header.

Before:
image

After:
image

> I checked the internet archive wayback machine to verify what was the size we used to be using. See https://web.archive.org/web/20250117183607/https://patterns.hypothes.is/


robertknight:
>  Among other things, that accidentally reduced the default size of the logo icon.

To clarify, it made the logo icon the same size as other icons:

logo-diff

robertknight (approved):

PR #1940, opened 2025-04-08
Fix typo in input placeholder color class

acelaya:
Part of https://github.com/hypothesis/client/issues/6866

This PR does two things:

1. Fix two typos in `Input` tailwind classes around placeholder text color.
2. Consolidate `Input` and `Textarea` ~text color and~ placeholder color, following the changes in https://github.com/hypothesis/client/pull/6966

~Inputs were previously inheriting the text color from the page, having inconsistent values. Now we settle to `grey-8`~

The previous placeholder color (even though it was not working due to the typo mentioned above) was too light in contrast with white background. Now we settle to `grey-6`, which has more than 4.5 contrast, as required by accessibility guidelines.

image
image
image

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -16,7 +16,7 @@ export function inputStyles({ classes, feedback }: InputStylesOptions) {
   return classnames(
     'focus-visible-ring ring-inset border rounded w-full p-2',
     'bg-grey-0 focus:bg-white disabled:bg-grey-1',
-    'placeholder:text-color-grey-5 disabled:placeholder:color-grey-6',
+    'placeholder:text-grey-6 disabled:placeholder:text-grey-7',
 
     // On iOS, the input font size must be at least 16px to prevent the browser
     // from zooming into it on touch.


robertknight:
> Inputs were previously inheriting the text color from the page, having inconsistent values. Now we settle to grey-8.

Are there any cases where we _do_ want to inherit the color that you can think of? For example input fields that set the text to red if there is an error or something. Do we have a mechanism to do that if needed?

acelaya:
> > Inputs were previously inheriting the text color from the page, having inconsistent values. Now we settle to grey-8.
> 
> Are there any cases where we _do_ want to inherit the color that you can think of? For example input fields that set the text to red if there is an error or something. Do we have a mechanism to do that if needed?

Our existing `feedback` capability only changes the border color ATM.

image

I can't think on any case where we currently need to change the text color, but I think it would be reasonable to just set the placeholder text and continue inheriting the main text color.

I would in that case do the same in https://github.com/hypothesis/client/pull/6966, for consistency.

robertknight:
> So we could just keep grey-9 for all, via inheriting the page color.

This makes sense as an incremental step. In future we should consider whether we want to harmonize the text color between the client, LMS and other apps.

src/components/input/Input.tsx
```diff
@@ -16,7 +16,7 @@ export function inputStyles({ classes, feedback }: InputStylesOptions) {
   return classnames(
     'focus-visible-ring ring-inset border rounded w-full p-2',
     'bg-grey-0 focus:bg-white disabled:bg-grey-1',
-    'placeholder:text-color-grey-5 disabled:placeholder:color-grey-6',
+    'placeholder:text-grey-6 disabled:placeholder:text-grey-7',
```
acelaya, line 19: Neither `text-color-grey-5` or `color-grey-6` were valid tailwind classes, but even if they were, `grey-5` is too light in contrast with white.

acelaya:
> > So we could just keep grey-9 for all, via inheriting the page color.
> 
> This makes sense as an incremental step. In future we should consider whether we want to harmonize the text color between the client, LMS and other apps.

@robertknight can you take another look? I have reverted the `text-grey-8` class, and just addressed the typos and too low contrast for the placeholder color.

src/components/input/Input.tsx
robertknight: We really ought to have a lint to check validity of classes. I suspect there was confusion here because we do have some custom [`text-color-*` properties](https://github.com/hypothesis/frontend-shared/blob/9f84354fdd7541e93664c41422b6b0989fca3aad/src/tailwind.preset.js#L67).

robertknight (approved):
This change seems like the most straightforward way of addressing the accessibility hazard with placeholder text. It does however mean that contrast is a weaker cue than before that text in an input field is a placeholder. This raises the importance of adding other cues such as adding an ellipsis at the end.

GitHub recently made a similar change across their input fields and it runs into the same challenges.

src/components/input/Input.tsx
acelaya: > I suspect there was confusion here because we do have some custom [text-color-* properties](https://github.com/hypothesis/frontend-shared/blob/9f84354fdd7541e93664c41422b6b0989fca3aad/src/tailwind.preset.js#L67).
yep, that was also my thinking.

PR #1943, opened 2025-04-14
Add dotted-circle icon

acelaya:
Add new icon that we'll use for the annotation premoderation capability, when annotations are pending moderation.

image

> I exported the icon from the designs in figma https://www.figma.com/design/oWeHoOkzViYXFaOdSzd9lH/BioRxiv?node-id=828-1396&t=pIr3Ks33QKlwOG5L-0


robertknight:
It wasn't at all obvious to me what this icon was going to be used for, until I logged at the Figma mocks. From the mocks I can see it is supposed to represent a _pending moderation_ status. Should we give this a more semantic name?

PR #1944, opened 2025-04-14
Add new shades of green, red and yellow

acelaya:
This PR brings new "ligth" and "dark" shades for the green, yellow and red tailwind preset colors.

These colors where originally defined for the auto-grading project in LMS, but we are going to need them soon for premoderation in h and client:

image

These new shades are defined together with the original "green-success", "yellow-notice" and "red-error", which still exist, and become aliases for the `DEFAULT` variants of their corresponding colors.

image

### DIFF
--- a/src/pattern-library/components/patterns/ColorsPage.tsx
+++ b/src/pattern-library/components/patterns/ColorsPage.tsx
@@ -55,9 +55,29 @@ const slateExamples = (
 
 const stateExamples = (
   <>
-    <ColorSwatch colorClass="bg-green-success" colorName="green-success" />
-    <ColorSwatch colorClass="bg-yellow-notice" colorName="yellow-notice" />
-    <ColorSwatch colorClass="bg-red-error" colorName="red-error" />
+    <ColorSwatch
+      colorClass="bg-green-success"
+      colorName="green-success (green alias)"
+    />
+    <ColorSwatch
+      colorClass="bg-yellow-notice"
+      colorName="yellow-notice (yellow alias)"
+    />
+    <ColorSwatch colorClass="bg-red-error" colorName="red-error (red alias)" />
+  </>
+);
+
+const highlightingExamples = (
+  <>
+    <ColorSwatch colorClass="bg-green-light" colorName="green-light" />
+    <ColorSwatch colorClass="bg-green" colorName="green" />
+    <ColorSwatch colorClass="bg-green-dark" colorName="green-dark" />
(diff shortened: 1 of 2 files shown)


src/pattern-library/components/patterns/ColorsPage.tsx
```diff
@@ -5,12 +5,22 @@ import Library from '../Library';
...
+}: ColorSwatchProps) {
   return (
     <div>
-      <div className={classnames(colorClass, 'w-64 h-8 mr-4')} />
+      <div
+        className={classnames(colorClass, 'h-8', {
+          'md:w-64': span === 'third',
+          'md:w-96': span === 'half',
+        })}
+      />
```
acelaya, line 23: There's room for improvement in this page layout, but I decided to keep things simple for now.

We can combe back end define a proper responsive grid later.

acelaya:
> It feels a little awkward to be mixing semantic and descriptive names for colors (`green-success` vs `green-light`). Let's roll with it for now.

Yeah, I feel the same. Do you have some proposal to avoid mixing approaches?

robertknight:
> Yeah, I feel the same. Do you have some proposal to avoid mixing approaches?

We could have a descriptive scale (`green-{light, medium, dark}` with eg. `green-success` as an alias for one of those? Perhaps worth looking at what popular patterns are in the community.

acelaya:
> We could have a descriptive scale (green-{light, medium, dark} with eg. green-success as an alias for one of those

I like this approach, because if you check the screenshot above, the cards where the new light and dark colors are going to be used are also using the other existing colors as border, so perhaps it makes sense they become the "middle" ones.

 > Perhaps worth looking at what popular patterns are in the community.
 
 I checked a bit and I can't see a very clear approach, even in the same library.

[Tailwind](https://tailwindcss.com/docs/colors) just defines a list of colors, and 10 levels of darkness for each one, represented by numbers. We do something similar for our shades of grey.

[Bootstrap](https://getbootstrap.com/docs/5.3/utilities/colors/#colors) defines a list of named colors, but all have an `-emphasis` variant which is darker. On top of that, they do define a few shades of grey in the form of white, black, white-50 and black-50

[MUI](https://mui.com/material-ui/customization/color/) has a color palette similar to Tailwind's, but recommend you to configure your theme with three shades per color: light, main and dark.

So I think I will go with `light`, `dark` and `main` per color, and alias existing "named" ones to the `main` one.

PR #1947, opened 2025-04-14
Set-up to run tests with vitest browser mode

acelaya:
> Depends on https://github.com/hypothesis/frontend-build/pull/722

This is a POC which aims to replace karma as the test runner with vitest, using browser mode + playwright.

Karma has been deprecated and is no longer actively maintained, so we need to eventually find a replacement that allows us to continue writing our tests in the same way, and running a pre-bundled tests bundle.

This PR:

- Adds vitest and playwright as dev dependencies.
- Defines a new gulp task that bundles our tests, but does not run karma afterwards.
- Adds Vitest configuration, enabling browser mode in headless chromium.
- Adds a new npm task that runs tests with vitest rather than karma.

### Todo

- [x] Fix couple of still failing tests. (https://github.com/hypothesis/frontend-shared/pull/1950)
- [x] Ensure code coverage reports are generated.
- [x] Fix code coverage error unless `yarn.lock` is deleted and dependencies are cleanly installed.
    - I did not push the changes, but in order to avoid a weird error, I had to delete `yarn.lock` and install dependencies again. Then coverage would be generated properly.
    ~I need to find what's the exact dependency conflict that causes this. My theory is that there are a bunch of instanbul instances that are conflicting, and may be fixed by simply uninstalling all karma stuff, including its istanbul plugin.~
    **EDIT**: I just verified running `yarn dedupe` also fixes this issue. I assume it's because there are multiple instances of istanbul in probably incompatible versions, because multiple dependencies depend on it themeselves.
- [x] Install playwright dependencies during CI.
- [x] Document how to get everything set-up and automate where possible.
    - Makefile has been updated, so that we run `yarn playwright install` every time we run `yarn install`. That should keep playwright binaries up to date.
    Running that command when no updates are needed has a very low impact compared to the overall execution time of `yarn install`.
- [x] Support running tests with `--grep <some_pattern>` (https://github.com/hypothesis/frontend-build/pull/722)
- [x] Support live mode when `--live` is passed. (https://github.com/hypothesis/frontend-build/pull/722)
- [x] **Remove karma and mocha** once build is green.


test/bootstrap.js
```diff
@@ -12,31 +12,10 @@ sinon.assert.expose(assert, { prefix: null });
 // karma-sinon plugins
 globalThis.assert = assert;
 globalThis.sinon = sinon;
+globalThis.context = globalThis.context ?? globalThis.describe;
```
acelaya, line 15: Vitest does not have a global `context`, so this simply creates an alias from `describe`, which is basically what with mocha does.

gulpfile.js
```diff
@@ -60,3 +60,17 @@ gulp.task(
...
+// Some (eg. a11y) tests rely on CSS bundles. We assume that JS will always take
+// longer to build than CSS, so build in parallel.
+gulp.task(
+  'test:vitest',
+  gulp.parallel('build-test-css', () =>
+    runTests({
+      bootstrapFile: 'test/bootstrap.js',
+      rollupConfig: 'rollup-tests.config.js',
+      testsPattern: 'src/**/*-test.js',
+      vitestConfig: './vitest.config.js',
```
acelaya, line 73: This requires changes from https://github.com/hypothesis/frontend-build/pull/722

yarn.lock
acelaya, line 1: Most removals here are due to the execution of `yarn dedupe`, which was needed to solve a weird error when trying to generate code coverage reports.

By googling the error I found many people recommending the removal of the `yarn.lock` file, which did work, but introduced many more changes.

Based on that I assumed the issue was in the fact that istanbul was a dependency of multiple other dependencies, which was causing different conflicting versions to be installed.

Makefile
```diff
@@ -65,4 +65,5 @@ build: node_modules/.uptodate
 
 node_modules/.uptodate: package.json yarn.lock
 	yarn install
+	yarn playwright install
```
acelaya, line 68: This has a low impact when the binaries are already up to date. In my machine it's around 0.6s, which is acceptable if we only run it when `yarn install` is also needed.

.github/workflows/continuous-integration.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         path: node_modules
         key: ${{ runner.os }}-node_modules-${{ hashFiles('yarn.lock') }}
     - name: Install
-      run: yarn install --immutable
+      run: yarn install --immutable && yarn playwright install
```
robertknight, line 19: We don't actually need Firefox or WebKit at present. We could use `yarn playwright install chromium` to install only the browser we use. This will save some time and disk space. If someone wants to install additional browsers for testing they can do so.

robertknight:
I see a message about `RadioGroup` appear in the output, although it doesn't cause test failures:

```
[I]  ~/h/frontend-shared > make test
yarn test
[12:09:19] Using gulpfile ~/hypothesis/frontend-shared/gulpfile.js
[12:09:19] Starting 'test'...
[12:09:19] Starting 'build-test-css'...
[12:09:19] Starting '<anonymous>'...
[12:09:19] Building test bundle... (63 files)
[12:09:19] Finished 'build-test-css' after 366 ms
[12:09:22] Starting vitest...

 RUN  v3.1.1 /Users/robert/hypothesis/frontend-shared
      Coverage enabled with istanbul

stderr | build/scripts/tests.bundle.js > RadioGroup > when RadioGroup.Radio is used outside of RadioGroup > throws an error
Add @babel/plugin-transform-react-jsx-source to get a more detailed component stack. Note that you should not add it to production builds of your App for bundle size reasons.
 ✓  chromium  build/scripts/tests.bundle.js (651 tests) 908ms
...
```

vitest.config.js
```diff
@@ -0,0 +1,38 @@
...
+    include: [
+      // Test bundle
+      './build/scripts/tests.bundle.js',
+    ],
+
+    coverage: {
+      enabled: true,
+      provider: 'istanbul',
+      reportsDirectory: './coverage',
+      reporter: ['json', 'clover', 'html'],
```
robertknight, line 26: What is the `clover` reporter? Do we need it?

robertknight (commented):
The playwright install command can be optimized by installing only the browsers we are going to use, especially in CI. Otherwise this looks good.

vitest.config.js
acelaya: Isn't clover the one used by codecov?

.github/workflows/continuous-integration.yml
acelaya: Oh, that can be done? I was googling to see if that was possible and didn't find anything. I'll change it.

vitest.config.js
acelaya: Oh, they support the json one as well. I'll remove clover then.

robertknight (approved):

PR #1962, opened 2025-04-28
Add RichCheckbox component

acelaya:
Add a component that looks visually like the existing `Radio` from `RadioGroup`, but displays a checkbox icon and works on its own.

> [!NOTE]
> This component does not follow a common pattern where you have a focusable/actionable individual checkbox input with a label, and optionally a helper text underneath. Instead, this component is focusable/actionable entirely, and has some opinionated styles which try to make it visually consistent with the radio group radios.
>
> This has the side effect that it cannot contain other actionable elements, like links. If we ever need that kind of pattern, it would need to be implemented separately, probably by composing the existing `Checkbox` component with some extra capabilities.

https://github.com/user-attachments/assets/7af3295f-9d21-4259-a99c-f07a9332fa03

image

> [!NOTE]
> There's some functionality we may need in future, but I decided to not implement it yet, and wait for an actual use case to appear.
> - Allow a `name` to be provided, so that a hidden input with that name is rendered, allowing this component to be used with "traditional" forms.
> - Support disabling the component, as we do with `RadioGroup` radios.

### DIFF
--- a/src/components/input/RichCheckbox.tsx
+++ b/src/components/input/RichCheckbox.tsx
@@ -1,0 +1,75 @@
+import classnames from 'classnames';
+import type { ComponentChildren } from 'preact';
+import { useCallback } from 'preact/hooks';
+
+import { CheckboxCheckedFilledIcon, CheckboxIcon } from '../icons';
+
+export type RichCheckboxProps = {
+  checked: boolean;
+  onChange: (checked: boolean) => void;
+
+  /**
+   * Content provided as children is vertically aligned with the checkbox icon
+   */
+  children: ComponentChildren;
+
+  /**
+   * Allows to provide extra content to be displayed under the children, in a
+   * smaller and more subtle font color.
+   */
+  subtitle?: ComponentChildren;
+};
+
+/**
+ * An opinionated `[role="checkbox"]` component which displays a checkbox icon
+ * next to provided content.
+ *
+ * If a `subtitle` is provided, it will be shown in a lighter color right under
+ * the main content, and aligned on the left with it.
+ */
+export default function RichCheckbox({
+  checked,
+  onChange,
+  children,
+  subtitle,
+}: RichCheckboxProps) {
+  const toggle = useCallback(() => onChange(!checked), [checked, onChange]);
+
+  return (
+    <div
+      className={classnames(
+        'group focus-visible-ring',
+        'grid gap-x-1.5 items-center grid-cols-[auto_1fr]',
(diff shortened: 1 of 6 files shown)


acelaya:
I choose `RichCheckbox` as opposed to `Checkbox`, but I'm open to suggestions on a better name.

src/components/input/RichCheckbox.tsx
```diff
@@ -0,0 +1,73 @@
...
+      {!checked && <CheckboxIcon className="mt-1" />}
+      {checked && <CheckboxCheckedFilledIcon className="mt-1" />}
+      <div>
+        <p className="text-grey-7 group-hover:text-grey-8 group-aria-checked:text-grey-8">
+          {children}
+        </p>
+        {subtitle && (
+          <p
+            data-testid="subtitle"
+            className="text-grey-6 group-hover:text-grey-7 group-aria-checked:text-grey-7"
```
acelaya, line 65: The text colors are consistent with the ones set in https://github.com/hypothesis/frontend-shared/pull/1955

acelaya:
> On macOS (Safari and Chrome) the title text and checkbox are slightly misaligned in `RichCheckbox`, compared to `Checkbox`.

Hmm, interesting. It might also be the case in other OSes, but I just didn't notice it.

I reckon I took the same styles we used for the RadioGroup's radios, but it could be that there's some slight difference in the icon and this needs adjustments. I'll take a more careful look.

acelaya:
> On macOS (Safari and Chrome) the title text and checkbox are slightly misaligned in `RichCheckbox`, compared to `Checkbox`.

Ok, this was a general issue. The alignment was a bit brittle, manually set via vertical margins, just closely depend on the size of the icon.

I have refactored the whole component to use a grid layout with 2 columns, where each row is vertically align.

robertknight:
It feels unfortunate to me to have two separate checkbox components. Any new developer using the library has to try and figure out which they should use in a particular context. Also we may get unintended divergence where an improvement is implemented in one but not the other.

We definitely have some use cases where we need a subtitle and some without, but do we have a clear reason to have two different hover styles? It seems there is a design decision here. If we need to do an incremental rollout of a change then we could have different variants.

acelaya:
As per my comment in slack https://hypothes-is.slack.com/archives/C4K6M7P5E/p1746005737334339, I'm closing this for now, as it is opening a whole new discussion that is blocking other tasks.

We can come back to this topic later, and decide how to better address it.

Issue #1921, opened 2025-03-24
Make sure `Popover` does not render outside the viewport's y axis


acelaya:
Some time ago we [implemented some logic](https://github.com/hypothesis/frontend-shared/pull/1746) to make sure the Popover never spanned outside of the x axis of the viewport, causing a lot of horizontal scroll.

The popover also has some logic to drop up of the anchor element instead of down, if there's not enough space below. However, this logic does not take into consideration that there could also not be enough space above either, so the popover cannot be fully displayed.

We should add some extra logic that ensures the popover is conditionally displayed above or below, but based on where there's more available space. Then we should set a max height that makes scroll to be rendered, preventing part of the popover content to be cropped.

PR #2000, opened 2025-06-13
Allow to customize initial Popover placement

acelaya:
Add a new `placement?: 'above' | 'below'` property to the `Popover`, so that it is possible to customize the default placement of the popover relative to the anchor.

Up until now, popovers would always render below the anchor when there's enough space.

The component can still decide to render in the opposite placement if selected one does not allow for it to render inside the viewport.

https://github.com/user-attachments/assets/4b9a8645-04eb-491c-9f0b-9a1dfe82f009

### Test steps

1. Go to http://localhost:4001/feedback-popover
2. Scroll down to the `placement` property doc, which has an example of `above` and `below` popovers.
3. While the buttons are more or less in the middle, popovers should show in the placement indicated by every button.
4. If the buttons are close to the top or bottom, their placements may change.
5. Resize the window so that the popovers do not fit neither above or below.
6. They should now show in the direction where there's more available space, even if they don't fit.

> [!WARNING] 
> Ideally, when the popover does not fit neither above or below, it should re-position and overlap the anchor element, to make sure it is kept in the viewport.
> That is not part of this PR, and will be addressed as part of https://github.com/hypothesis/frontend-shared/issues/1921

### TODO

- [x] Ensure placement falls back only if the popover does not fit in available space, not if there's more space in the opposite direction.
- [x] Add tests.
- [x] Document new prop.

### DIFF
--- a/src/components/feedback/Popover.tsx
+++ b/src/components/feedback/Popover.tsx
@@ -41,6 +41,7 @@ function usePopoverPositioning(
   popoverOpen: boolean,
   asNativePopover: boolean,
   alignToRight: boolean,
+  placement: 'above' | 'below',
 ) {
   const adjustPopoverPositioning = useCallback(() => {
     const popoverEl = popoverRef.current!;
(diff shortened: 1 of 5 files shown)


src/components/feedback/Popover.tsx
```diff
@@ -250,6 +255,17 @@ export type PopoverProps = {
...
+  /**
+   * Where to position the popover if there's available space: above the anchor
+   * or below it.
+   * Defaults to 'below'.
+   *
+   * Notice that if there's no space to display the popover in selected
+   * placement, it will be displayed in the other one in an attempt to keep it
+   * inside the viewport.
+   */
+  placement?: 'above' | 'below';
```
acelaya, line 267: I'm not super happy naming this prop `placement`. I'm open to suggestions.

PR #2010, opened 2025-06-17
Refactor usePopoverPositioning internal hook with options

acelaya:
The `Popover` component has an internal `usePopoverPositioning` hook which needs to know about a bunch of props and refs in order to determine the right location for the popover.

This PR refactors that hook so that it expects one object with DOM element refs, and another one with options, so that we can reduce the amount of arguments.

I choose this approach because refs are potentially modified by the hook, while options are only used to modify its behavior.

I'm planning to add one more ref and option, to dynamically add an arrow pointing to the anchor, which also needs to be dynamically positioned based on some heuristics.

### DIFF
--- a/src/components/feedback/Popover.tsx
+++ b/src/components/feedback/Popover.tsx
@@ -25,27 +25,34 @@ type PopoverCSSProps =
   | 'bottom'
   | 'marginTop';
 
+type PopoverPositioningOptions = {
+  open: boolean;
+  placement: 'above' | 'below';
+
+  /**
+   * Whether the popover should be aligned to the right side of the anchor
+   * element or not
+   */
+  alignToRight: boolean;
+
+  /** Native popover API is used to toggle the popover */
+  asNativePopover: boolean;
+};
+
 /**
  * Manages the popover position manually to make sure it renders "next" to the
  * anchor element (above or below). This is mainly needed when using the
  * popover API, as that makes it render in the top layer, making it impossible
  * to position it relative to the anchor element via regular CSS.
- *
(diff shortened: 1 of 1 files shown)


src/components/feedback/Popover.tsx
```diff
@@ -25,27 +25,43 @@ type PopoverCSSProps =
   | 'bottom'
   | 'marginTop';
 
+type PopoverPositioningRefs = {
+  anchorRef: RefObject<HTMLElement | undefined>;
+  popoverRef: RefObject<HTMLElement | undefined>;
+};
+
+type PopoverPositioningOptions = {
+  popoverOpen: boolean;
```
robertknight, line 34: Name could be shortened to `open`.

src/components/feedback/Popover.tsx
```diff
@@ -25,27 +25,43 @@ type PopoverCSSProps =
...
- *                       of the anchor element or not
  */
 function usePopoverPositioning(
-  anchorElementRef: RefObject<HTMLElement | undefined>,
-  popoverRef: RefObject<HTMLElement | undefined>,
-  popoverOpen: boolean,
-  asNativePopover: boolean,
-  alignToRight: boolean,
-  placement: 'above' | 'below',
+  { anchorRef, popoverRef }: PopoverPositioningRefs,
```
robertknight, line 54: Since there are only two refs and both are required, it might be slightly nicer to pass them as positional arguments, rather than two separate options-like objects.

robertknight (approved):

PR #2011, opened 2025-06-17
Allow an arrow to be added to popover component

acelaya:
Add new `arrowed?: boolean` prop to `Popover` that allows an arrow/triangle pointing to the anchor element to be rendered.

The arrow is aligned to the same side as the popover, and points up or down depending on the Popover placement.

https://github.com/user-attachments/assets/9575a16b-246a-4a44-b401-132ef2f11f08

The functionality is very simple at this point and is not very customizable, but let's wait to see what new requirements appear before making it more complex.

Our initial use case is using the Popover for the share annotation behavior in the sidebar and the moderation queue. This is how it looks with the code from this PR:

https://github.com/user-attachments/assets/a642dd20-41c5-487d-9ed8-c73351eb1859

> [!TIP]
> This PR is easier to review [ignoring whitespaces](https://github.com/hypothesis/frontend-shared/pull/2011/files?w=1)

### TODO

- [x] Document new prop
- [x] Add tests

### DIFF
--- a/src/components/feedback/Popover.tsx
+++ b/src/components/feedback/Popover.tsx
@@ -1,15 +1,21 @@
 import classnames from 'classnames';
 import type { ComponentChildren, JSX, RefObject } from 'preact';
-import { useCallback, useEffect, useLayoutEffect } from 'preact/hooks';
+import {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useState,
+} from 'preact/hooks';
 
 import { useClickAway } from '../../hooks/use-click-away';
 import { useKeyPress } from '../../hooks/use-key-press';
 import { useSyncedRef } from '../../hooks/use-synced-ref';
 import { ListenerCollection } from '../../util/listener-collection';
 import { downcastRef } from '../../util/typing';
+import { PointerDownIcon, PointerUpIcon } from '../icons';
 
-/** Small space to apply between the anchor element and the popover */
-const POPOVER_ANCHOR_EL_GAP = '.15rem';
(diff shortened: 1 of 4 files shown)


src/components/feedback/Popover.tsx
```diff
@@ -378,16 +424,12 @@ export default function Popover({
     <div
       className={classnames(
         'absolute z-5',
-        variant === 'panel' && [
-          'max-h-80 overflow-y-auto overflow-x-hidden',
-          'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
-        ],
```
acelaya, line 384: The outermost Popover element is now "unstyled", and there's a new inner element wrapping children which is the one receiving styling classes.

This allows the arrow to be displayed inside the "unstyled" container, and keep the other one looking as before.

There's one consideration. The `classes` prop, which intuitively one would expect to be added to the outermost element, is now added to the "styled" inner element, to ensure it keeps looking the same. Maybe we should document this in the prop and in future we can add a new `containerClasses` prop if we need to provide classes to the outermost element.

acelaya:
> Safari is showing a border between the arrow and the popover:

Thanks! Good catch.

robertknight (commented):
The main feedback I have is to avoid using the word "arrowed". Use the phrase "Popover with arrow" in the example and either `arrow` or `showArrow` or `withArrow` as the prop name. I also found a render issue in Safari that seems to occur when the popover's calculated `top` position is not a whole number of pixels. Changing the value so that it is a whole number of pixels resolves the issue.

PR #2014, opened 2025-06-19
Fix regressions introduced in Popover

acelaya:
This PR reverts a few of the changes introduced in https://github.com/hypothesis/frontend-shared/pull/2011, which have caused a few visual regressions (see https://github.com/hypothesis/client/pull/7159#issuecomment-2984363252 and https://github.com/hypothesis/frontend-shared/pull/2013).

The root cause was that the DOM structure was changed, adding one extra element between the outermost Popover element and the `children`, and moving some of the classes to this new element, including the propagation of the `classes` prop.

https://github.com/hypothesis/frontend-shared/pull/2013 was an attempt at working around one of those issues, but by checking how we use `Popover.classes` in different places, I realized it would not be possible to consistently split those into `classes` and `containerClasses` (or `contentClasses`), specially considering `Select` makes use of these with its own `popoverClasses`, which would require splitting in two as well.

Instead, this PR restructures the DOM elements again, so that it is exactly as before, when `arrow` is false, and when it is true, it only changes `overflow-y-auto overflow-x-hidden` with `overflow-visible` in the outermost element.

This overflow difference was the motivation of the original DOM structure, but it's now clear the implemented solution was overkill.

Having visible overflow might require tweaks in consuming code, but since that is only going to affect new implementations where `arrow` is `true`, we can address those on a case-by-case basis, but ensuring all other existing `Popover` usages are "safe" of regressions.

### DIFF
--- a/src/components/feedback/Popover.tsx
+++ b/src/components/feedback/Popover.tsx
@@ -415,12 +415,18 @@ export default function Popover({
     <div
       className={classnames(
         'absolute z-5',
+        variant === 'panel' && [
+          'max-h-80 ',
+          'rounded border bg-white shadow hover:shadow-md focus-within:shadow-md',
+          !arrow && 'overflow-y-auto overflow-x-hidden',
+        ],
+        arrow && 'overflow-visible',
         asNativePopover && [
           // We don't want the popover to ever render outside the viewport,
           // and we give it a 16px gap
           'max-w-[calc(100%-16px)]',
           // Overwrite [popover] default styles
-          'p-0 m-0 overflow-visible',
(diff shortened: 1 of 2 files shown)


robertknight (approved):
Looks good. We discussed in [a Slack thread](https://hypothes-is.slack.com/archives/C1M8NH76X/p1750321742341809) that it isn't ideal to have `arrow: true` have this side effect of changing overflow behavior. We discussed that potentially it can be eliminated by moving the overflow rules down into components that need it (`Select` was mentioned as the main reason), but this is best done as a separate PR.

PR #2035, opened 2025-07-03
Add `lazy` utility to create lazy-loading UI components

robertknight:
Add a utility for creating components which are loaded only when rendered. When first rendered, a lazy component displays a fallback while the real implementation is fetched. It then renders the real component after the fetch completes. If the component fails to load an an error fallback is rendered in its place.

Part of https://github.com/hypothesis/h/issues/9709.

### DIFF
--- a/src/util/lazy.tsx
+++ b/src/util/lazy.tsx
@@ -1,0 +1,69 @@
+import type { ComponentChildren, FunctionalComponent, JSX } from 'preact';
+import { useState } from 'preact/hooks';
+
+export type LazyOptions<P> = {
+  /** Returns the content to render if the component is not yet loaded. */
+  fallback: (props: P) => ComponentChildren;
+  /** Returns the content to render if the component failed to load. */
+  errorFallback?: (props: P, err: Error) => ComponentChildren;
+};
+
+/**
+ * Create a lazy-loading version of a component.
+ *
+ * This utility allows deferring loading the code for a component until it is
+ * rendered. The returned component loads in two phases. In the first phase a
+ * placeholder is rendered and the {@link onLoad} callback is invoked to load
+ * the component. Then when the returned promise resolves, the placeholder is
+ * replaced with the real compoonent.
+ *
+ * @param displayName - Display name for the lazy wrapper component
+ * @param onLoad - A function which loads the JS component. This will usually
+ *   be an async function which does `import('path/to/module')` and then returns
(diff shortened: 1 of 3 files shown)


src/pattern-library/examples/lazy-basic.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <div className="p-4 border border-green-500 rounded">
+      <h3 className="text-green-700 font-bold">{name}</h3>
+      <p>{description}</p>
+    </div>
+  );
+}
+
+let resolve: (x: typeof Profile) => void;
+const promise = new Promise<typeof Profile>(resolve_ => (resolve = resolve_));
```
acelaya, line 18: Considering the limited scope of the pattern library (mainly used by us during development), perhaps you can use `Promise.withResolvers` here?

src/util/lazy.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param displayName - Display name for the lazy wrapper component
+ * @param load - A function which loads the JS component. This will usually
+ *   be an async function which does `import('path/to/module')` and then returns
+ *   one of the loaded module's exports.
+ * @param options - Options that specify what to render while the component is
+ *   loading or if it fails to load.
+ */
+export function lazy<P>(
+  displayName: string,
```
acelaya, line 28: I see why we would want to be able to set a display name, but it's unfortunate it has to be manually set every time.

I wonder if there's some way we can infer it. Perhaps we could move `displayName` to the last options argument, mandatory for now, so that in a later iteration we can explore making it optional, and trying to fall back to `compnent.name ?? component.displayName` if not provided.

Let's land this without the falling back logic, as that requires some testing, and I reckon it would need some side effect that updates `Lazy.displayName` as the component is loaded, and I don't know how well is that going to work, and how incorrect it is.

acelaya (approved):
This looks good to me. Just provided some feedback about the requirement of setting a display name.

src/pattern-library/examples/lazy-basic.tsx
robertknight: The immediate reason I didn't is because the TypeScript library target doesn't make the method available (it requires es2024). In order for the pattern library to use newer APIs than are allowed in the library itself, we'll need to adjust the TS configuration. Currently the npm library and pattern library use the same TS config.

PR #2050, opened 2025-08-04
Support callback returning ES module in `lazy` utility

robertknight:
Make the `lazy` utility more convenient to use by allowing the `load` callback to return an ES module whose default export is a component.

With this change, this common usage:

```js
const Widget = lazy('Widget', () => import('./Widget').then(mod =>
mod.default), options)
```

Can be simplified to:

```js
const Widget = lazy('Widget', () => import('./Widget'), options);
```


acelaya (approved):
👍🏼

PR #2090, opened 2025-09-15
Add a CSS version of the Tailwind JS config

robertknight:
This will enable downstream projects to migrate from JS-based configuration for Tailwind to the preferred CSS-based theme config. See also https://tailwindcss.com/docs/upgrade-guide#using-a-javascript-config-file.

The JS configuration has been kept for compatibility under all downstream projects have been migrated.

**TODO:**
- [x] Convert pattern library site to use this instead of the JS config
- [x] Update documentation to reference this instead of the JS docs

### DIFF
--- a/src/tailwind-config.css
+++ b/src/tailwind-config.css
@@ -1,0 +1,228 @@
+/* Standard theme and additional utilities for Hypothesis projects. */
+
+@theme {
+  /* Border colors */
+  --color-border: #dbdbdb;
+
+  /* Custom color palette */
+  --color-grey-0: #fafafa;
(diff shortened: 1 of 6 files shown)


src/tailwind-config.css
```diff
@@ -0,0 +1,228 @@
+/* Standard theme and additional utilities for Hypothesis projects. */
```
robertknight, line 1: This is a fairly straightforward translation of `tailwind.config.js` and the presets and plugins it imports.

acelaya (approved):

PR #2094, opened 2025-09-18
Export CloseableContext from the root of the package

acelaya:
Export the `CloseableContext` symbol from the root of the package, since the `exports` definition no longer allows importing from a specific path deep inside the library.


acelaya:
Actually, I'll create a new branch, so that it also has a correct name.

PR #2097, opened 2025-09-18
Remove Tailwind JS config

robertknight:
Going forwards downstream projects must use CSS-based config instead, which is the preferred method in Tailwind v4.


robertknight:
All downstream projects have been converted to use CSS-based config for Tailwind, so we can ship this now.

